PR #8, opened 2025-04-10
Build development extension with Yocto

fvasquez:
```
root@avocado-qemux86-64:~# systemd-sysext
HIERARCHY EXTENSIONS   SINCE
/usr      debug-tweaks Thu 2025-04-17 04:57:17 UTC
root@avocado-qemux86-64:~# systemd-confext
HIERARCHY EXTENSIONS   SINCE
/etc      debug-tweaks Thu 2025-04-17 04:57:17 UTC
```


meta-avocado/recipes-avocado/extensions/avocado-extension.bb
```diff
@@ -0,0 +1,87 @@
+SUMMARY = "Avocado combination system/configuration extension"
```
mobileoverlord: I think we should make this file .inc or a class as to not make it seem like it can be built as a stand alone recipe.

mobileoverlord (approved):
I think this is the only feedback I've got. Otherwise, let it rip!

meta-avocado/recipes-avocado/extensions/avocado-extension.bb
fvasquez: I agree.

PR #9, opened 2025-04-10
Sdk Target Metadata package

mobileoverlord:
feat: Add SDK Target Metadata and Repository Mapping

This pull request introduces a system for generating and managing Avocado SDK and target package repository configurations.

**Key Changes:**

1.  **`avocado-sdk-metadata.bb` Recipe:**
    *   A new Yocto recipe has been added to generate SDK/target repository configuration files.
    *   It creates `avocado-sdk.repo` (packaged into the SDK/image) containing repository definitions for `yum`/`dnf`.
    *   It generates `avocado-repo.map` in the deploy directory (`DEPLOY_DIR_RPM`), mapping package architectures to their corresponding repository paths on the server.
    *   The paths are determined dynamically based on `DISTRO_CODENAME`, `MACHINE`, `PACKAGE_ARCHS`, `SDK_PACKAGE_ARCHS`, and the `AVOCADO_REPO_BASE` URL.

2.  **`setup-repo.sh` Script:**
    *   The script has been refactored to utilize the `avocado-repo.map` file.
    *   Instead of scanning all subdirectories in the deploy directory, it now reads the map file to determine the source directory (`key`) and the target repository path (`value`) for each architecture.
    *   It copies the RPMs and creates the repository metadata (`createrepo_c`) based on the structure defined in the map file.

3.  **Configuration Updates:**
    *   Added `DISTRO_CODENAME = "alpha"` to `avocado.conf` to support the new metadata recipe logic.
    *   Updated `packagegroup-avocado-sdk.bb` to include `avocado-sdk-metadata` as a runtime dependency, ensuring the `.repo` file is included in the SDK.

**Motivation:**

These changes provide a more structured and flexible way to define and manage package repositories for different target machines and SDK variations within the Avocado build system. Using a map file centralizes the repository structure definition and allows the `setup-repo.sh` script to correctly populate the repository server based on the build output.

### DIFF
--- a/meta-avocado/recipes-avocado/sdk/avocado-sdk-metadata.bb
+++ b/meta-avocado/recipes-avocado/sdk/avocado-sdk-metadata.bb
@@ -49,10 +49,79 @@ python do_install() {
     gpg_check = d.getVar('AVOCADO_REPO_GPGCHECK') or '0'
 
     # --- Precompute values ---
-    # Machine name without 'avocado-' prefix
     machine_short_name = machine.replace('avocado-', '')
-    # SDK suffix with underscores
     sdk_suffix_us = sdk_pkg_suffix.replace('-', '_')
+
+    # --- Helper Functions ---
+    def _determine_repo_paths(arch, arch_dir, distro_codename, machine_short_name, sdk_suffix_us, machine, bb):
+        """Determines repository paths and names based on architecture rules."""
+        map_value_path = None
+        repo_url_path = None
+        repo_name = None
+        repo_section_name = None
+
+        # --- Apply Rules ---
+        if arch == "all_avocadosdk":
+            # Rule: all_avocadosdk -> DISTRO/sdk/all
+            map_value_path = f"{distro_codename}/sdk/all"
+            bb.note(f"Mapping arch '{arch}' (dir: {arch_dir}) to map path '{map_value_path}' (no repo entry)")
+
+        elif sdk_suffix_us and arch_dir.endswith(sdk_suffix_us):
+            # Rule: other *avocadosdk -> DISTRO/sdk/MACHINE_SHORT
+            map_value_path = f"{distro_codename}/sdk/{machine_short_name}"
+            repo_url_path = map_value_path
+            repo_name = f"{machine_short_name}-sdk"
+            repo_section_name = repo_name # Use cleaned name for section too
+            bb.note(f"Mapping SDK arch '{arch}' (dir: {arch_dir}) to path '{map_value_path}', repo name '{repo_name}'")
+
+        elif arch_dir == machine.replace('-', '_'):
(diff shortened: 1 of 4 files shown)


meta-avocado/recipes-avocado/sdk/avocado-sdk-metadata.bb
```diff
@@ -0,0 +1,154 @@
...
+                    repo_f.write(f"gpgcheck={gpg_check}\n")
+                    repo_f.write(f"priority={priority}\n")
+                    repo_f.write("\n")
+                    priority += 1
+                elif arch != "all_avocadosdk": # Only log warning if it wasn't the explicitly excluded arch
+                    bb.warn(f"Repo entry details not fully determined for arch '{arch}'. Skipping repo entry.")
+
+            else:
+                bb.note(f"Skipping arch '{arch}' as directory '{check_dir}' does not exist")
+}
```
fvasquez, line 191: This `do_install` task is a bit of a beast. I bet you could break the body of the `for` loop down into three or so helper functions.

mobileoverlord: Thanks!, Updated

meta-avocado/recipes-avocado/sdk/avocado-sdk-metadata.bb
```diff
@@ -0,0 +1,194 @@
...
+                if _write_repo_entry(
+                    repo_f, base_repo_url,
+                    repo_details["repo_url_path"],
+                    repo_details["repo_name"],
+                    repo_details["repo_section_name"],
+                    gpg_check, priority, arch, bb
+                ):
+                    priority += 1 # Increment priority only if repo entry was written
+            else:
+                bb.note(f"Skipping arch '{arch}' as directory '{check_dir}' does not exist")
```
fvasquez: Extract the body of this `for` loop into a helper function that only takes `arch` as an argument. Python has lexical scope so you can capture variables defined in the outer scope (e.g. `bb`, `deploy_dir_rpm`, `base_repo_url`, `distro_codename`, `machine`, `sdk_suffix_us`, `gpg_check`, `machine_short_name`, `repo_file_path`, `priority`) automatically by name inside of an inner function. Lambda capture make sense here since these values are all constant. The same technique can be applied to the other inner helper functions like `determine_repo_paths` and `write_repo_entry`.

fvasquez (changes requested):
Suggested refactoring.

meta-avocado/recipes-avocado/sdk/avocado-sdk-metadata.bb
mobileoverlord: Updated the function and loop

fvasquez (approved):
Thank you!

PR #10, opened 2025-04-16
Add cryptoauthlib to extra packages

mobileoverlord:
Build cryptoauth lib from Microchip for package support with the ATECC* cryptoauth processors. 

Unfortunately, `meta-atmel` conflates the cryptoauthlib recipes with their BSP in a single layer. A lot of the appends can cause some mysterious trouble and so I BBMASK'ed them out. Alternatively, maybe we manage this ourselves in a separate layer, for now I still desire upstream support. 

This also updates the build container with required `git-lfs` support

### DIFF
--- a/kas/base.yml
+++ b/kas/base.yml
@@ -1,5 +1,7 @@
 header:
   version: 11
+  includes:
+    - kas/extra/atecc.yml
 
 build_system: oe
 
--- a/kas/extra/atecc.yml
+++ b/kas/extra/atecc.yml
@@ -1,0 +1,29 @@
+header:
+  version: 11
+
+repos:
+  meta-atmel:
+    url: https://github.com/linux4sam/meta-atmel
+    branch: scarthgap
+    layers:
+      .:
+
+local_conf_header:
+  atecc: |
+    BBMASK = "meta-atmel/recipes-atmel/"
+    BBMASK += "meta-atmel/recipes-bsp/"
+    BBMASK += "meta-atmel/recipes-connectivity/"
+    BBMASK += "meta-atmel/recipes-core/"
+    BBMASK += "meta-atmel/recipes-devtools/"
+    BBMASK += "meta-atmel/recipes-egt/"
+    BBMASK += "meta-atmel/recipes-graphics/"
+    BBMASK += "meta-atmel/recipes-httpd/"
+    BBMASK += "meta-atmel/recipes-kernel/"
+    BBMASK += "meta-atmel/recipes-multimedia/"
+    BBMASK += "meta-atmel/recipes-security/optee"
+    BBMASK += "meta-atmel/recipes-support/"
+    BBMASK += "meta-atmel/recipes-utils/"
+
+  git-lfs: |
+    HOSTTOOLS += "git-lfs"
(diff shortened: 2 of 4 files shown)


kas/base.yml
```diff
@@ -1,5 +1,7 @@
 header:
   version: 11
+  includes:
+    - kas/extra/atecc.yml
 
 build_system: oe
```
fvasquez, line 7: Does this include belong in `base.yml`? Most targets don't have an ATECC.

kas/extra/atecc.yml
```diff
@@ -0,0 +1,29 @@
...
+    BBMASK += "meta-atmel/recipes-httpd/"
+    BBMASK += "meta-atmel/recipes-kernel/"
+    BBMASK += "meta-atmel/recipes-multimedia/"
+    BBMASK += "meta-atmel/recipes-security/optee"
+    BBMASK += "meta-atmel/recipes-support/"
+    BBMASK += "meta-atmel/recipes-utils/"
+
+  git-lfs: |
+    HOSTTOOLS += "git-lfs"
+    FETCHCMD_git = "git -c core.fsyncobjectfiles=0"
```
fvasquez, line 29: Committing binaries to Git? How barbaric!

meta-avocado/recipes-avocado/packagegroups/packagegroup-avocado-extra.bb
```diff
@@ -2,7 +2,9 @@ DESCRIPTION = "Packagegroup for Avocado extra"
 LICENSE = "Apache-2.0"
 
 inherit packagegroup
+PACKAGE_ARCH = "${MACHINE_ARCH}"
```
fvasquez, line 5: I wonder what implications this has on all the packages being built.

Does this mean Yocto will compile with all the tunes for a given target `MACHINE_ARCH` whenever possible?

fvasquez (commented):
two questions

kas/base.yml
mobileoverlord: most targets do not have one on the board, but it is possible to connect one like the dev kit for example, in which case, building cryptoauthlib as an extra package will make it possible to use it on any platform.

kas/extra/atecc.yml
mobileoverlord: Its sad, reminds me of the linux-firmware blob days

meta-avocado/recipes-avocado/packagegroups/packagegroup-avocado-extra.bb
mobileoverlord: Unclear. This was added in response to the dreaded, "allarch dynamic renamed package error". I am open to other ways of solving this.

kas/base.yml
fvasquez: makes sense

meta-avocado/recipes-avocado/packagegroups/packagegroup-avocado-extra.bb
fvasquez: If this shuts that up, then great.

fvasquez (approved):
Resolve conflicts before merging.

PR #12, opened 2025-04-17
Add build support for Seeed Studio's reTerminal

mobileoverlord:
While this recipe will build all the required artifacts to get the reTerminal booting and working, it raises an interesting discussion point regarding how to handle bootfiles that are deployed but not packaged. The recipes for Raspberry Pi config and cmdline (txt) are prime examples of this. They both inherit nopackage and therefore only exist in the deployed image dir in Yocto. We are going to need to package these so we can get them out to the next phase of tooling so they can be composed into a runtime. For now, this recipe does what it's supposed to and we will need to follow up on this extra requirement. These fragments were copied in from https://github.com/Seeed-Studio/meta-seeed-cm4

### DIFF
--- a/kas/reterminal.yml
+++ b/kas/reterminal.yml
@@ -1,0 +1,19 @@
+
+header:
+  version: 11
+  includes:
+    - kas/base.yml
+    - kas/vendor/raspberrypi.yml
+
+repos:
+  meta-avocado:
+    path: meta-avocado
+    layers:
+      meta-avocado-raspberrypi:
+
+local_conf_header:
+  reterminal: |
+    ROOTFS_IMAGE_EXTRA_INSTALL:append = " packagegroup-avocado-reterminal"
+
+
+machine: avocado-raspberrypi4
--- a/meta-avocado-raspberrypi/recipes-bsp/bootfiles/rpi-config_git.bbappend
+++ b/meta-avocado-raspberrypi/recipes-bsp/bootfiles/rpi-config_git.bbappend
@@ -1,0 +1,30 @@
+do_deploy:append:seeed() {
+    CONFIG=${DEPLOYDIR}/${BOOTFILES_DIR_NAME}/config.txt
+    grep -q "^dtoverlay=vc4-kms-v3d-pi4$" $CONFIG || echo "dtoverlay=vc4-kms-v3d-pi4" >> $CONFIG
+    grep -q "^dtoverlay=dwc2,dr_mode=host$" $CONFIG || echo "dtoverlay=dwc2,dr_mode=host" >> $CONFIG
+    grep -q "^enable_uart=1$" $CONFIG || echo "enable_uart=1" >> $CONFIG
+    grep -q "^dtparam=spi=on$" $CONFIG || echo "dtparam=spi=on" >> $CONFIG
+
+    if ${@bb.utils.contains('MACHINEOVERRIDES', 'seeed-reterminal-dm', 'true', 'false', d)} \
+        || ${@bb.utils.contains('MACHINEOVERRIDES', 'seeed-reterminal-dm-mender', 'true', 'false', d)}; then
+        grep -q "^dtoverlay=reTerminal-DM$" $CONFIG || echo "dtoverlay=reTerminal-DM" >> $CONFIG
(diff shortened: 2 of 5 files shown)


kas/reterminal.yml
```diff
@@ -0,0 +1,19 @@
...
+
+repos:
+  meta-avocado:
+    path: meta-avocado
+    layers:
+      meta-avocado-raspberrypi:
+
+local_conf_header:
+  reterminal: |
+    ROOTFS_IMAGE_EXTRA_INSTALL:append = " packagegroup-avocado-reterminal"
```
fvasquez, line 16: I'm not convinced `ROOTFS_IMAGE_EXTRA_INSTALL:append` actually works yet.

kas/vendor/raspberrypi.yml
```diff
@@ -10,4 +10,4 @@ repos:
 
 local_conf_header:
   raspberrypi: |
-    IMAGE_INSTALL:append = " packagegroup-avocado-raspberrypi"
+    ROOTFS_IMAGE_EXTRA_INSTALL:append = " packagegroup-avocado-raspberrypi"
```
fvasquez: Ditto my preceding comment.

meta-avocado-raspberrypi/recipes-bsp/bootfiles/rpi-config_git.bbappend
```diff
@@ -0,0 +1,30 @@
...
+        grep -q "^dtparam=i2c_arm=on$" $CONFIG || echo "dtparam=i2c_arm=on" >> $CONFIG
+        grep -q "^dtoverlay=i2c1,pins_44_45$" $CONFIG || echo "dtoverlay=i2c1,pins_44_45" >> $CONFIG
+        grep -q "^dtoverlay=i2c3,pins_2_3$" $CONFIG || echo "dtoverlay=i2c3,pins_2_3" >> $CONFIG
+        grep -q "^dtoverlay=i2c6,pins_22_23$" $CONFIG || echo "dtoverlay=i2c6,pins_22_23" >> $CONFIG
+        grep -q "^dtoverlay=audremap,pins_18_19$" $CONFIG || echo "dtoverlay=audremap,pins_18_19" >> $CONFIG
+        grep -q "^dtoverlay=reComputer-R100x,uart2$" $CONFIG || echo "dtoverlay=reComputer-R100x,uart2" >> $CONFIG
+    else
+        bbdebug 1 "No target device tree specified, check your MACHINEOVERRIDES"
+    fi
+}
```
fvasquez, line 30: I am very confused by this bootfile `.bbappend`. I know you copied it from `meta-seeed-cm4` so I trust that it does something sensible but I can't tell you what that is. I _think_ it generates a `config.txt` file from DTS, but I don't see why we can't just keep a complete `config.txt` in Git and rev that whenever the DTS revs.

fvasquez (approved):
See my comments.

kas/reterminal.yml
mobileoverlord: If it makes you feel better, It built the reterminal dtbo

meta-avocado-raspberrypi/recipes-bsp/bootfiles/rpi-config_git.bbappend
mobileoverlord: We could go that far. This is certainly a copy paste from upstream. Its trying to be "nice" and merge with upstream defaults. Therefore if raspberry pi add something as a new default, they are only appending to it via a grep >> echo.

mobileoverlord: Unclear how often config.txt values are being added, but I imagine thats the motivation

fvasquez: I see that this bootfile `.bbappend` only revs once or twice a year so Seeed must be getting some mileage out of it.

kas/reterminal.yml
fvasquez: Yes, but did the dtbo make it onto the boot partition?

PR #13, opened 2025-04-17
update IMAGE_INSTALL -> ROOTFS_IMAGE_EXTRA_INSTALL

mobileoverlord:
Since we are building multiple images at once, fix up the image install variables to user the more explicit ones.

### DIFF
--- a/kas/vendor/nxp.yml
+++ b/kas/vendor/nxp.yml
@@ -10,4 +10,4 @@ repos:
 
 local_conf_header:
   imx: |
-    IMAGE_INSTALL:append = " packagegroup-avocado-imx"
+    ROOTFS_IMAGE_EXTRA_INSTALL:append = " packagegroup-avocado-imx"
--- a/kas/vendor/raspberrypi.yml
+++ b/kas/vendor/raspberrypi.yml
@@ -10,4 +10,4 @@ repos:
 
 local_conf_header:
   raspberrypi: |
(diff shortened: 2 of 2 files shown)


fvasquez (commented):
The only place I've seen `ROOTFS_IMAGE_EXTRA_INSTALL` used is in `avocado-image-rootfs.bb`:

```
IMAGE_INSTALL = "packagegroup-avocado-rootfs ${ROOTFS_IMAGE_EXTRA_INSTALL}"
```

As you can see, `ROOTFS_IMAGE_EXTRA_INSTALL` ends up in `IMAGE_INSTALL` anyway, so I'm not sure what the purpose of this more explicit variable is.

My experience so far has been that if I append a package to `ROOTFS_IMAGE_EXTRA_INSTALL` inside of a `local.conf`, then the package does _not_ get installed to the root filesystem. Don't know what to attribute that to (a typo?).

mobileoverlord:
My concern with using generically `IMAGE_INSTALL` from local.conf s that with a slip of a typo, that could mean `INITRAMFS` or `ROOTFS`. While it inevitably ends up in `IMAGE_INSTALL`, I want to be more explicit of which `IMAGE_INSTALL` it should end up in. If `ROOTFS_IMAGE_EXTRA_INSTALL` is not functioning properly, that should be resolved.

fvasquez (approved):
I am working with QEMU at the moment so any missing packages won't affect me.

PR #14, opened 2025-04-23
Install sysext/confexts as btrfs subvolumes in /var partition

fvasquez:
```
root@avocado-qemux86-64:~# btrfs subvol list /var
ID 256 gen 10 top level 5 path lib/extensions
ID 257 gen 10 top level 5 path lib/confexts
```


kas/qemux86-64-secureboot.yml
```diff
@@ -14,6 +14,6 @@ local_conf_header:
   avocado-qemux86-64: |
     ROOT_FSTYPES = "squashfs wic"
     IMAGE_INSTALL:append = " avocado-extension-debug-tweaks"
-    INITRAMFS_MAXSIZE = "200000"
+    INITRAMFS_MAXSIZE = "240000"
```
mobileoverlord: I have suspicions that this was increased as a result of the line above it. I am pretty confident that appending to `IMAGE_INSTALL` affects INITRAMFS image

meta-avocado/recipes-avocado/extensions/avocado-extension-debug-tweaks.bb
```diff
@@ -23,20 +23,43 @@ EXT_DEPENDS = " \
     ltrace \
     iproute2 \
     htop \
+    btrfs-tools \
```
mobileoverlord: We could probably include this in the rootfs since it BTRFS support would be required for var support

kas/qemux86-64-secureboot.yml
fvasquez: I will switch to `ROOTFS_IMAGE_EXTRA_INSTALL` and see what happens.

meta-avocado/recipes-avocado/extensions/avocado-extension-debug-tweaks.bb
fvasquez: I agree. We need `btrfs-tools` to create subvolumes and take snapshots.

mobileoverlord (approved):
Awesome!

PR #15, opened 2025-04-30
Add support to build raspberry pi USB boot util for nativesdk

mobileoverlord:
This package is used to USB boot compute modules into a composite USB mass storage mode so the associated eMMC on the module can be programmed over USB from a host machine. Adding this recipe to the nativesdk enables its use during final image composition tooling for raspberry pi devices that would need this.

### DIFF
--- a/kas/vendor/raspberrypi.yml
+++ b/kas/vendor/raspberrypi.yml
@@ -11,3 +11,4 @@ repos:
 local_conf_header:
   raspberrypi: |
     ROOTFS_IMAGE_EXTRA_INSTALL:append = " packagegroup-avocado-raspberrypi"
+    SDK_PKG_EXTRA_INSTALL:append = " packagegroup-avocado-raspberrypi-sdk"
--- a/meta-avocado/recipes-avocado/images/avocado-pkg-sdk.bb
+++ b/meta-avocado/recipes-avocado/images/avocado-pkg-sdk.bb
@@ -3,4 +3,5 @@ LICENSE = "Apache-2.0"
 
 inherit image-packages-only
 
-IMAGE_INSTALL = "packagegroup-avocado-sdk"
+SDK_PKG_EXTRA_INSTALL ??= ""
+IMAGE_INSTALL = "packagegroup-avocado-sdk ${SDK_PKG_EXTRA_INSTALL}"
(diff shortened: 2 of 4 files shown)


meta-avocado/recipes-avocado/images/avocado-pkg-sdk.bb
```diff
@@ -3,4 +3,5 @@ LICENSE = "Apache-2.0"
 
 inherit image-packages-only
 
-IMAGE_INSTALL = "packagegroup-avocado-sdk"
+SDK_PKG_EXTRA_INSTALL ??= ""
```
fvasquez, line 6: What does this line do? Empty out `SDK_PKG_EXTRA_INSTALL`? Why would you want to do that?

fvasquez: Ah, I remember now. Because `SDK_PKG_EXTRA_INSTALL` installs _all_ the packages built for Avocado OS.

meta-avocado/recipes-avocado/images/avocado-pkg-sdk.bb
```diff
@@ -3,4 +3,5 @@ LICENSE = "Apache-2.0"
...
+IMAGE_INSTALL = "packagegroup-avocado-sdk ${SDK_PKG_EXTRA_INSTALL}"
```
fvasquez, line 7: Does `SDK_PKG_EXTRA_INSTALL` get populated in another layer?

fvasquez (commented):
two questions

kas/vendor/raspberrypi.yml
```diff
@@ -11,3 +11,4 @@ repos:
 local_conf_header:
   raspberrypi: |
     ROOTFS_IMAGE_EXTRA_INSTALL:append = " packagegroup-avocado-raspberrypi"
+    SDK_PKG_EXTRA_INSTALL:append = " packagegroup-avocado-raspberrypi-sdk"
```
mobileoverlord, line 14: SDK_PKG_EXTRA_INSTALL gets populated here. The intention is that any target / platform specific tools that may need to be present can go in this.

meta-avocado/recipes-avocado/images/avocado-pkg-sdk.bb
mobileoverlord: Yocto gets grumpy when you try to use variables that haven't been declared prior. This is just declaring it and if its not been set just make it empty so the next call can be successful

meta-avocado/recipes-avocado/images/avocado-pkg-sdk.bb
mobileoverlord: See other comments.

fvasquez (approved):
good to go

PR #19, opened 2025-05-03
Update local conf kas labels to better tie to the source

mobileoverlord:
The header label used in the kas config will be left in the local conf as a commend. Updated the headers to labels that will be more helpful in understanding the source of the values from each section when inspecting the local conf directly.


fvasquez (approved):
`qemux86-64-secureboot.yml` still builds.

PR #20, opened 2025-05-03
btrfs-tools: bump to 6.14 for --subvol support

mobileoverlord:
Bump to latest version to use `--subvol` in `mkfs.btrfs`. Adds a patch to work around a problem where file permissions in the btrfs volume would always be host permissions and not fakeroot permissions. This was introduced in c6464d3f99ed1dabceff1168eabb207492c37624 when rewriting to use nfts. nfts's callback function stat doesn't use the intercepted calls for fakeroot to function properly. This is likely "inappropriate" and long term either btrfs or more likely psuedo/fakeroot will need to support how nfts is stat'ing files


fvasquez (approved):
```
$ kas shell ../kas/qemux86-64-secureboot.yml -c "bitbake -c build btrfs-tools"
```

PR #22, opened 2025-05-12
extra: set pkg arch before inheriting from pkg grp

mobileoverlord:
Setting the PACKAGE_ARCH before inheriting package group to fix issue with building rename packages for allarch IE cruptoauthlib


fvasquez (approved):
Good.

PR #25, opened 2025-05-12
SDK Test Container Updates

mobileoverlord:
Updates the sdk-test container and compose to automate more of the process. 
* Automatically install the toolchain and populate the sysroot for the toolchain
* Automatically source the toolchain environment script
* Fix the sdk-metadata recipe to always deploy the map file. Fixes do_install being skipped when package is being run from sstate.
* Add helper script for running Qemu using the nativesdk in the container.
* Add btrfs-tools and dnf to the toolchain. Packages added to the toolchain should be necessary for all sdk toolchain base functions. Include these here so they are always available.
* Update the export to the DEPLOY_DIR and mount the images output to the sdk-test. In the future anything the user wants that ends up in deploy images should be made available via rpm package installation.


fvasquez (approved):
Wonderful. Thank you.

PR #28, opened 2025-05-14
Add script to bump and pin kas deps to edge

mobileoverlord:
Using kas without commits enables simplified testing of edge for all dependencies, but does not allow external repos to easily lock meta-peridio and its dependency refs. This adds a script that will walk the kas directory and for each repo, check the remote for the latest ref of the specified branch and bump the commit value to the latest. This lays the groundwork for a nightly operation to automatically run and present a PR to be merged (like dependabot).

### DIFF
--- a/kas/base.yml
+++ b/kas/base.yml
@@ -1,7 +1,7 @@
 header:
   version: 11
   includes:
-    - kas/extra/atecc.yml
+  - kas/extra/atecc.yml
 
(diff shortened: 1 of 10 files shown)


fvasquez:
Doesn't issuing a `git pull` from each layer achieve the same result?

fvasquez (approved):
LGTM

PR #29, opened 2025-05-14
Add support for running labeled builds for PRs

mobileoverlord:
Adding a label build-<MACHINE> will trigger a build on the build server for that target. Only the act of adding the label with trigger the build. If you want to re run a build job, you need to remove the label and add it again. These precautions work well for public repositories while limiting the ability to add labels to only repo maintainers preventing un-reviewed code from running on navocado build runners

### DIFF
--- a/.github/workflows/pr-build-labeled.yml
+++ b/.github/workflows/pr-build-labeled.yml
@@ -1,0 +1,51 @@
+name: PR-Build-Labeled
+
+on:
+  pull_request:
+    types: [ labeled ]
+
+jobs:
+  set-matrix:
+    if: ${{ github.event.label.name == 'build' }}
+    runs-on: molcajete
+    outputs:
+      matrix: ${{ steps.get-machines.outputs.matrix }}
+    steps:
+      - name: Fetch and filter PR labels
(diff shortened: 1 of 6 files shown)


mobileoverlord:
Screenshot 2025-05-14 at 15 07 19
A disappointing aspect to this is that the name of the "check" cannot dynamically include the label that triggered it.

.github/workflows/pr-build-labeled.yml
```diff
@@ -0,0 +1,51 @@
...
+              }
+            );
+            // 2) Keep only labels like "build-foo", emit ["foo","bar",…]
+            const machines = labels
+              .map(l => l.name)
+              .filter(n => n.startsWith('build-'))
+              .map(n => n.replace(/^build-/, ''));
+            core.info(`💡 Machines: ${machines.join(', ')}`);
+            // 3) JSON-stringify them into an output
+            core.setOutput('matrix', JSON.stringify(machines));
```
fvasquez, line 36: I like this style of JavaScript.

fvasquez (approved):
good stuff!

PR #30, opened 2025-05-15
Use SDK to build var image with debug-tweaks extension

fvasquez:
Still need to `kas build ../kas/machine/qemux86-64-secureboot.yml` to regenerate wic file with new `avocado-image-var-avocado-qemux86-64.btrfs` image inside.

### DIFF
--- a/support/sdk-test/build_var.sh
+++ b/support/sdk-test/build_var.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+mkdir -p ~/avocado-qemux86-64/rootfs/var
+
+dnf -y install avocado-extension-debug-tweaks \
+    --setopt=tsflags=noscripts \
+    --installroot ~/avocado-qemux86-64/rootfs/
+
+rm -f /images/avocado-qemux86-64/avocado-image-var-avocado-qemux86-64.btrfs
+
+mkfs.btrfs -r ~/avocado-qemux86-64/rootfs/var \
+    --subvol rw:lib/extensions --subvol rw:lib/confexts \
+    -f /images/avocado-qemux86-64/avocado-image-var-avocado-qemux86-64.btrfs
+
+sync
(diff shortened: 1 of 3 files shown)


support/sdk-test/entrypoint.sh
```diff
@@ -14,6 +14,12 @@ dnf install -y --setopt=tsflags=noscripts avocado-sdk-toolchain
 mkdir -p /opt/avocado/sdk/avocado-qemux86-64/0.1.0/sysroots/core2-64-avocado-linux/var
 dnf -y --setopt=tsflags=noscripts --installroot /opt/avocado/sdk/avocado-qemux86-64/0.1.0/sysroots/core2-64-avocado-linux/ install packagegroup-core-standalone-sdk-target
```
mobileoverlord, line 16: Lets add a comment here that this can be removed once the toolchain can install these files.

support/sdk-test/build_var.sh
```diff
@@ -0,0 +1,17 @@
...
+
+mkdir -p ~/avocado-qemux86-64/rootfs/var
+
+dnf -y install avocado-extension-debug-tweaks \
+    --setopt=tsflags=noscripts \
+    --installroot ~/avocado-qemux86-64/rootfs/
+
+rm -f /images/avocado-qemux86-64/avocado-image-var-avocado-qemux86-64.btrfs
+
+mkfs.btrfs -r ~/avocado-qemux86-64/rootfs/var \
```
mobileoverlord, line 13: Oh this is interesting. very neat

fvasquez: Thank god this `--subvol` feature exists. And thank you for patching `btrfs-tools` so I can use it.

support/sdk-test/entrypoint.sh
fvasquez: done

mobileoverlord (approved):

PR #36, opened 2025-05-20
Default to building wic and wic.bmap for bmaptool

mobileoverlord:
bmaptool speeds up writes to storage media by omitting 0 blocks. Include this by default in the distro `ROOT_FSTYPES` and add it to the nativesdk


fvasquez (approved):
Tim Orling created bmaptool.

PR #39, opened 2025-05-21
nvidia: Update build filetypes

mobileoverlord:
* Fixes the nvidia jetson builds
* Removes the `INITRAMFS_MAXSIZE` checks by clearing out the value. This seems to behaving incorrectly when building multiple images


kas/machine/jetson-orin-nano-devkit-nvme.yml
```diff
@@ -18,7 +18,5 @@ repos:
 
 local_conf_header:
   machine-avocado-jetson-orin-nano-devkit-nvme: |
-    TEGRA_OOT_ALL_DRIVER_PACKAGES:append = " nv-kernel-module-rtl8852ce"
```
fvasquez, line 21: No longer needed? Or is this appended someplace else?

fvasquez (commented):
Just one comment.

kas/machine/jetson-orin-nano-devkit-nvme.yml
mobileoverlord: This was a bug and should have never been here. It was causing a duplicate package in the build

fvasquez (approved):
LGTM

PR #43, opened 2025-05-27
Update yocto-build container to direnv allow build directories.

mobileoverlord:
Removes the need to call `direnv allow` when entering into the build directory.


fvasquez (approved):
LGTM

PR #49, opened 2025-06-04
Update container build scripts

mobileoverlord:
Adds support for `avocado-build image` and `avocado-run-qemu`. See the `support/sdk-test/README` for more info


support/sdk-test/README.md
```diff
@@ -71,3 +71,50 @@ Example adding libcryptoauth3 to the sysext sysroot. The sysext sysroot has been
...
+```bash
+/opt/_avocado/extensions/sysext/peridiod.raw
+```
+
+## Building and booting the image in qemu
+
+Build the final image
+
+```bash
+avocado-build image
```
fvasquez, line 105: You forgot `avocado-images repo`. Otherwise this step fails.

fvasquez (changes requested):
`avocado-build image` failing

support/sdk-test/README.md
```diff
@@ -71,3 +71,50 @@ Example adding libcryptoauth3 to the sysext sysroot. The sysext sysroot has been
...
+
+This will place files into the folder
+
+```bash
+/opt/_avocado/output
+```
+
+## Booting the image in qemu
+
+Extens the toolchain with qemu-system-x86-64
```
fvasquez: "Extends"

support/sdk-test/README.md
```diff
@@ -71,3 +71,50 @@ Example adding libcryptoauth3 to the sysext sysroot. The sysext sysroot has been
...
+
+```bash
+avocado-repo sdk install nativesdk-qemu-system-x86-64
+```
+
+Run the VM
+
+```bash
+avocado-run-qemu
```
fvasquez, line 125: `/var/lib/extensions` and `/var/lib/confexts` are both empty even though I know I ran `avocado-build sysext peridiod` beforehand.

```
root@avocado-qemux86-64:~# btrfs subvolume list /var
ID 256 gen 10 top level 5 path lib/extensions
ID 257 gen 10 top level 5 path lib/confexts
root@avocado-qemux86-64:~# ls /var/lib/extensions
root@avocado-qemux86-64:~# ls /var/lib/confexts
root@avocado-qemux86-64:~#
```

I will write this off as a race condition for the time being, although you don't really have a demo until you come up with a reliable workaround or fix for this problem.

fvasquez (changes requested):
two comments to address before merging

support/sdk-test/README.md
mobileoverlord: I see whats wrong. When you call `avocado-build image` it will not rebuild the var partition if one has already been built. Therefore, if you built an image which created a var without any extensions it will not create a new one after you make changes to extensions. This is something that should be addressed in how these scripts get moved into the cli where we have a much more robust environment to track changes to source directories. In the meantime, I'll update the doc to mention that when you make changes to the extensions, you would have to call `avocado-build var` to force the var partition to be rebuilt.

fvasquez (approved):
go! go! go!

PR #50, opened 2025-06-11
nvidia|tegra: Update recipes to boot avocado

mobileoverlord:
In addition to updating the repository to add required components for tegra boot, this also adds paths for vendors / machines to add packages to all image/pkg repos. Updated the Initramfs recipe to allow machine specific inclusion of kernel-modules via `KERNEL_MODULE_AUTOLOAD`. These modules will be early loaded by systemd.


fvasquez (approved):
we are the champions ...

PR #51, opened 2025-06-11
yocto-build: Add kas aliases for kb and ks

mobileoverlord:
adds terminal aliases 
`kb` -> `kas build $KAS_YML`
`ks` -> `kas shell $KAS_YML`


fvasquez (approved):
keystrokes

PR #53, opened 2025-06-12
Add MACHINE_EXTRAs to extras packagegroup

mobileoverlord:
Our images do not base off `core` and therefore we need to handle adding `MACHINE_EXTRA_*` to the build. Add them to the extras repo.


fvasquez (approved):
LGTM

PR #60, opened 2025-06-16
Update CI builds for production releases

danielspofford:


.github/workflows/pr-build-labeled.yml
```diff
@@ -46,6 +46,15 @@ jobs:
...
         uses: actions/checkout@v4
       - name: build ${{ matrix.machine }}
-        uses: ./.github/actions/avocado-build
-        with:
-          machine: ${{ matrix.machine }}
+        env:
+          BB_GIT_SHALLOW: "1"
+          PARALLEL_MAKE: "-j16"
+          SDKMACHINE: "x86_64"
+          DISTRO_CODENAME: ""
```
mobileoverlord, line 53: I think its okay that we don't set `DISTRO_CODENAME` so it defaults to dev, but we should probably leave it "unset" instead of ""

.github/workflows/pr-build-labeled.yml
```diff
@@ -46,6 +46,15 @@ jobs:
...
+        run: |
+          echo "TARGET: avocado-distro"
+          echo "SDKMACHINE: x86_64"
+          echo "DISTRO_CODENAME: "
+
+          source ./scripts/init-build kas/machine/${{ matrix.machine }}.yml
+          PARALLEL_MAKE_ERLANG="-j16" kas build meta-avocado/kas/machine/${{ matrix.machine }}.yml --target avocado-distro
```
mobileoverlord, line 60: `PARALLEL_MAKE` is being set in the env to the same value. We don't need to also set `PARALLEL_MAKE_ERLANG`

mobileoverlord (approved):
Tested on aarch64 and x86_64 SDK containers.

PR #65, opened 2025-07-01
systemd: bump to 257 for extension --mutable support

mobileoverlord:
Backported from walnascar.
Also backported the updated meson class to support tag builds for systemctl-native

### DIFF
--- a/meta-avocado/classes/meson_tags.bbclass
+++ b/meta-avocado/classes/meson_tags.bbclass
@@ -1,0 +1,187 @@
+#
+# Copyright OpenEmbedded Contributors
+#
+# SPDX-License-Identifier: MIT
+#
+
+inherit python3native meson-routines qemu
+
+DEPENDS:append = " meson-native ninja-native"
+
+EXEWRAPPER_ENABLED:class-native = "False"
+EXEWRAPPER_ENABLED:class-nativesdk = "False"
+EXEWRAPPER_ENABLED ?= "${@bb.utils.contains('MACHINE_FEATURES', 'qemu-usermode', 'True', 'False', d)}"
+DEPENDS:append = "${@' qemu-native' if d.getVar('EXEWRAPPER_ENABLED') == 'True' else ''}"
+
+# As Meson enforces out-of-tree builds we can just use cleandirs
+B = "${WORKDIR}/build"
+do_configure[cleandirs] = "${B}"
+
+# Where the meson.build build configuration is
+MESON_SOURCEPATH = "${S}"
+
+# The target to build in do_compile. If unset the default targets are built.
+MESON_TARGET ?= ""
+
+# Since 0.60.0 you can specify custom tags to install
+MESON_INSTALL_TAGS ?= ""
+
+def noprefix(var, d):
+    return d.getVar(var).replace(d.getVar('prefix') + '/', '', 1)
+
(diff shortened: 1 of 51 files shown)


meta-avocado/recipes-core/systemd/dlopen-deps.inc
```diff
@@ -0,0 +1,81 @@
...
+
+            if ".so." in f or f.endswith(".so"):
+                try:
+                    elf = oe.qa.ELFFile(f)
+                    elf.open()
+                    for dep in parse(extract_segment(f, ".note.dlopen"), elf.isLittleEndian()):
+                        for soname in dep["soname"]:
+                            if soname in shlibs:
+                                # TODO assumes the first match is good
+                                package, version = list(shlibs[soname].values())[0]
```
fvasquez, line 72: Can we count on the first match always being good?

meta-avocado/classes/meson_tags.bbclass
```diff
@@ -0,0 +1,187 @@
...
+}
+
+meson_tags_do_install() {
+    if [ "x${MESON_INSTALL_TAGS}" != "x" ] ; then
+        meson_install_tags="--tags ${MESON_INSTALL_TAGS}"
+    fi
+    meson install --destdir ${D} --no-rebuild $meson_install_tags
+}
+
+EXPORT_FUNCTIONS do_configure do_compile do_install
```
fvasquez, line 187: I don't like how this recipe writes scripts out to `${WORKDIR}`, but this has already been upstreamed so I just have to live with it.

meta-avocado/recipes-core/systemd/systemd-compat-units.bb
```diff
@@ -0,0 +1,49 @@
...
+	else
+		OPTS=""
+	fi
+
+	for i in ${SYSTEMD_DISABLED_SYSV_SERVICES} ; do
+		if [ -e $i -o -e $i.sh ]  &&   ! [ -e $D${sysconfdir}/systemd/system/$i.service -o -e $D${systemd_system_unitdir}/$i.service ] ; then
+			echo -n "$i: "
+			systemctl $OPTS mask $i.service
+		fi
+	done
```
fvasquez, line 45: When does this execute? first time the target boots?

fvasquez (approved):
Thank you. I know what a pain it is to backport something of this magnitude. I used to backport new Erlang/OTP releases to a Buildroot LTS so Helium blockchain devs could take advantage of the latest language features.

I see that most of the patches were generated by Chen Qi and Khem Raj. We're lucky to have contributors like these in the Yocto community. Frank Hunleth used to be that person for me back in my Buildroot days.

meta-avocado/recipes-core/systemd/dlopen-deps.inc
mobileoverlord: Looks like there is opportunity to improve upstream. Good first issue 😊

meta-avocado/recipes-core/systemd/systemd-compat-units.bb
mobileoverlord: Looks like this recipe is used to bridge SysV init to systemd unit files. I've read that SysV compat is being removed in systemd 258 I believe.

mobileoverlord:
Tested build update locally, passes.

PR #66, opened 2025-07-01
Update distro features to be more inclusive

mobileoverlord:
Updates the default distro features to include the upstream defaults

### DIFF
--- a/meta-avocado/conf/distro/avocado.conf
+++ b/meta-avocado/conf/distro/avocado.conf
@@ -4,9 +4,9 @@ ROOT_FSTYPES ?= "squashfs"
 INITRAMFS_IMAGE = "avocado-image-initramfs"
 INITRAMFS_IMAGE_BUNDLE ?= "0"
 
-AVOCADO_DEFAULT_DISTRO_FEATURES = "ext2 ipv4 ipv6 nfs pci multiarch ptest systemd usrmerge vfat secureboot security pam"
+AVOCADO_DEFAULT_DISTRO_FEATURES = "multiarch ptest systemd usrmerge secureboot security pam ldconfig"
 DISTRO_FEATURES_EXTRA ?= ""
-DISTRO_FEATURES:append = "${AVOCADO_DEFAULT_DISTRO_FEATURES} ${DISTRO_FEATURES_EXTRA}"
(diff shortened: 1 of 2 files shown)


meta-avocado/conf/distro/avocado.conf
```diff
@@ -4,9 +4,9 @@ ROOT_FSTYPES ?= "squashfs"
 INITRAMFS_IMAGE = "avocado-image-initramfs"
 INITRAMFS_IMAGE_BUNDLE ?= "0"
 
-AVOCADO_DEFAULT_DISTRO_FEATURES = "ext2 ipv4 ipv6 nfs pci multiarch ptest systemd usrmerge vfat secureboot security pam"
+AVOCADO_DEFAULT_DISTRO_FEATURES = "multiarch ptest systemd usrmerge secureboot security pam ldconfig"
```
fvasquez, line 7: Why were `nfs` and `pci` removed? Are they appended somewhere else?

meta-avocado/conf/distro/avocado.inc
```diff
@@ -14,8 +14,6 @@ REQUIRED_AVOCADO_BBLAYERS_CONF_VERSION = "${DISTRO}-1"
 
 LOCALCONF_VERSION = "2"
 
-DISTRO_FEATURES:remove = "pulseaudio ldconfig gobject-introspection-data"
-
```
fvasquez, line 18: Do we really want to support `pulseaudio`?

fvasquez (commented):
two comments

meta-avocado/conf/distro/avocado.conf
mobileoverlord: A lot of these come from the Yocto DEFAULT_DISTRO_FEATURES which we include below this line. I was debating on using this variable or instead explicitly listing the contents of it. I think now I am leaning more towards explicitly listing the features instead of the content being unknown.

fvasquez: I am okay with relying with `DEFAULT_DISTRO_FEATURES` as long as there is no overlap between the two variables. If you choose to be explicit about `AVOCADO_DEFAULT_DISTRO_FEATURES`, then I wouldn't include `DEFAULT_DISTRO_FEATURES` at all.

mobileoverlord: Leaving it with `DEFAULT_DISTRO_FEATURES` will enable us to automatically support any "new features" from upstream changes. I like depending on upstream as much as we can.

fvasquez (approved):
Thank you!

PR #67, opened 2025-07-01
Ensure opt dir exists for sysext

mobileoverlord:
The opt dir needs to exist for sysexts that contain opt directories in their images otherwise sysexts containing an opt dir fail to mount.


fvasquez (approved):
LGTM

PR #68, opened 2025-07-01
Change PARALLEL_MAKE_ERLANG to PARALLEL_MAKE_HIGHMEM

mobileoverlord:
This changes the variable name so it can be applied to other things that take a lot of memory when running in parallel on build machines with high CPU count but not drastically high memory. This also fixes an issue that causes infinite recursion if undefined. Allows `PARALLEL_MAKE_HIGHMEM` to be set for anyone who `inherits` or `PARALLEL_MAKE_HIGHMEM-pn-{PN}` for tuning individual recipes.


fvasquez (approved):
LGTM

PR #69, opened 2025-07-02
imx8mp-evk: Add new target

mobileoverlord:
Adds support for building the imx8mp-evk. The `packagegroup-core-standalone-sdk-target` was complaining about generating a spdx. Its upstream package was generating this metapackage for `TUNE_PKGARCH`, when we really should be putting this under the `MACHINE_ARCH`. Fixes issues with spdx creation.


fvasquez (approved):
`kas build ../kas/machine/imx8mp-evk.yml` succeeded.

PR #87, opened 2025-07-18
Add helper scripts for building all machines and sdk containers

mobileoverlord:


fvasquez (commented):
How were we building all machines and SDK containers before these scripts?

Don't we want to delete those existing scripts or YAML as part of this pull request?

mobileoverlord:
These scripts are just development helpers

fvasquez (approved):
Go for it.

PR #91, opened 2025-07-22
add avocado CLI build and provision hooks

danielspofford:


meta-avocado/recipes-avocado/sdk/avocado-sdk-scripts/avocado-build-qemux86-64
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# $1 - Runtime name.
```
fvasquez, line 3: This is weird. The `$1` is inside a comment, so argument substitution will never happen.

fvasquez (commented):
You could source a common `avocado-build` or `avocado-provision` script containing function definitions at the top of each target-specific script so you don't end up repeating yourself.

meta-avocado/recipes-avocado/sdk/avocado-sdk-scripts/avocado-build-qemux86-64
danielspofford: Yeah it was just a quick note of what I wanted the param(s) to the script to be because there was some discussion around it - so never meant to be substituted yeah.

danielspofford:
> You could source a common `avocado-build` or `avocado-provision` script containing function definitions at the top of each target-specific script so you don't end up repeating yourself.

Yeah I discussed a few patterns for that with @mobileoverlord and we opted to hold off for a bit to give time for a pattern to emerge.

**Nevertheless this PR is re-opened here due to juggling some branches around: https://github.com/avocado-linux/meta-avocado/pull/96.**

PR #92, opened 2025-07-22
sdk: add mkfat for fatfs image generation from JSON

mobileoverlord:
Existing tools that make filesystem or partition images use configuration formats that do not have proper serializes (I'm lookin at you libconfuse). Programmatically creating configuration files for these tools can be error prone. We are going to need a mechanism for enabling users to add or modify files to the boot partitions within the SDK. We've created `mkfat` to simply create fat filesystem images from a JSON config file format.


fvasquez (approved):
This looks good. Do we need `mkfat` for any other targets besides `raspberrypi4` and `reterminal`? Since https://github.com/avocado-linux/mkfat is a public repo, do you anticipate any maintenance burden supporting users besides ourselves?

mobileoverlord:
This PR will build it into the nativesdk to be available for installation and use on any target. It will be helpful whenever we need to make a boot fat volume to hook into customization from the CLI. PRs welcome to the tool's repo.

PR #98, opened 2025-07-27
add stone to recipes devtools

danielspofford:


danielspofford:
I'd like to get stone to a real 1.0 before merging this.

Which right now is gated by a successful `stone validate` and `stone build` RPI4 test.

mobileoverlord (approved):

PR #109, opened 2025-08-03
fix sdk lifecycle script inclusion

danielspofford:


mobileoverlord:
Merged support in https://github.com/avocado-linux/meta-avocado/pull/102

PR #125, opened 2025-08-07
switch qemu* from genimage to stone

mobileoverlord:
Switch qemu targets from using genimage to using stone manifests.

In the current state, this PR has two side effects:
* run-qemu* yocto helper scripts are broken because yocto is not executing target specific stone build scripts to construct yocto built ready images for qemu to boot
* The target specific build scripts for qemu need to be rewritten to support the stone manifest items

There is a big consideration we need to make around handling the remainder of the values between the stone manifest and the underlying tools. Currently, block offsets and sizes for the rootdisk partition tables are being hardcoded. This needs to change before merge.


mobileoverlord:
This has been updated to calculate the partition contents from the stone manifest. Remaining work is around CLI / stone binary calls and bringing that back into yocto to fix the yocto qemu helpers. However, this is only for internal testing purposes.

danielspofford (approved):

PR #151, opened 2025-08-14
automatically generate sysuser.d conf files 

mobileoverlord:
systemd-sysusers offers a composable way for system users and groups to be managed by include files instead of packages trying to mutate single file databases like `/etc/passwd` with scriptlets. Yocto doesn't have built in support for generating these files from recipes, but they do have a mechanism for accumulating user and group account additions to munge the passwd/group files before imaging the system via USERADD/GROUPADD. Since these directives are already present in yocto, hooking in to this system give us the information we need. This adds a global class which hooks into this process and automatically generates sysuser.d conf files and adds them to their respective package. This is a big win because it means that packages can influence the user/groups at runtime without clobbering each other.


mobileoverlord:
depends on https://github.com/avocado-linux/avocado-control/pull/3 and rebasing the avocadoctl changes
depends on https://github.com/avocado-linux/avocado-cli/pull/23

lee-reinhardt (approved):

PR #197, opened 2025-12-09
zram integration

bsandu-peridio:
Add systemd-zram-generator and enable it for all targets.


kas/machine/rubikpi3.yml
```diff
@@ -17,6 +17,6 @@ repos:
 
 local_conf_header:
   machine-avocado-rubikpi3: |
-    AVOCADO_ALLOW_ROOT_LOGIN = "1"
+    #AVOCADO_ALLOW_ROOT_LOGIN = "1"
```
mobileoverlord: Lets remove this key from this config completely and depend on this being set in base.yaml. I'd also like to update base.yaml to allow this option to be set from an env var, allowing maintainers to configure certain build boxes to always produce builds with early login.

kas/machine/qemuarm64.yml
```diff
@@ -18,5 +18,6 @@ local_conf_header:
   machine-avocado-qemuarm64: |
     ## Debugging prior to extensions loading
     #AVOCADO_ALLOW_ROOT_LOGIN = "1"
+    ROOTFS_IMAGE_EXTRA_INSTALL:append = " systemd-zram-generator"
```
mobileoverlord: Let's go about this a different way. Instead of adding this to each individual config, lets add this as a defaut rootfs / initramfs images. We can make it depend on a DISTRO_FEATURE of `zram`. For systems we don't want to enable zram, that machine conf can remove from DISTRO_FEATURES.

mobileoverlord (changes requested):
I would also like to explicitly add the zram kernel options to the avocado-core fragments. In the future it would be nice to centralize this specific fragment and share it across systems.

kas/base.yml
```diff
@@ -10,6 +10,7 @@ env:
   DISTRO_VERSION: "0.1.0"
   PARALLEL_MAKE_HIGHMEM: ""
   SDKMACHINE: "x86_64"
+  BB_ENV_PASSTHROUGH_ADDITIONS: "ENABLE_ROOT_LOGIN $BB_ENV_PASSTHROUGH_ADDITIONS"
```
mobileoverlord: This doesn't work if its just:
`AVOCADO_ALLOW_ROOT_LOGIN: "0"`

PR #200, opened 2025-12-19
Update qemu targets

mobileoverlord:
* Add support to qemu targets to use the peridio provisioning profile in avocado.
* Bump avocadoctl to 0.2.2
* Bump stone to 1.8.2
* Update `avocado sdk run -iE vm` to align with stone rootdisk output names `avocado-os-<target>`


mobileoverlord:
merged

PR #208, opened 2026-04-05
fix: SSH and service usability on read-only rootfs

aaln:
## Summary

SSH and mDNS service discovery are broken out of the box on Avocado's read-only erofs rootfs. This PR fixes the core issues so that `openssh-sshd` and `avahi-daemon` work without any additional configuration.

### Problems fixed

1. **Missing sshd/avahi system users** — OpenSSH requires a `sshd` user for privilege separation, and avahi-daemon requires an `avahi` user. Without them, both services fail silently on startup.

2. **SSH host keys written to read-only /etc/ssh** — The `sshdgenkeys` service generates keys to `/etc/ssh/` which is on the immutable erofs rootfs. Keys appear to generate but are lost. This redirects key generation to `/var/lib/ssh/` (writable btrfs).

3. **Socket-activated sshd silently fails** — The `sshd.socket` + `sshd@.service` (inetd mode) pattern starts a per-connection sshd that immediately exits on the read-only rootfs. Replaced with a standalone `sshd.service` daemon that works reliably.

4. **Root home on read-only filesystem** — `/root` is on the erofs rootfs, so users can't create files, `.ssh/authorized_keys`, etc. Moved root's home to `/var/roothome` on the writable btrfs partition.

5. **`/var` btrfs not expanded to fill partition** — The var image is ~109MB but the data partition may be much larger (e.g., 118GB on NVMe). Added `var-resize.service` to run `btrfs filesystem resize max /var` on every boot.

6. **No root home initialization** — Added `var-roothome-init.service` to create `/var/roothome/.ssh`, `/var/roothome/.profile`, and `/var/log` on first boot.

### Changes

- `avocado-users`: add sshd:x:74, avahi:x:75 to passwd/group; move root home to /var/roothome; add var-resize.service and var-roothome-init.service
- `openssh`: bbappend with sshd_config pointing to /var/lib/ssh for host keys, custom key generation script, standalone sshd daemon service

### Testing

Tested on NVIDIA Jetson Orin Nano 8GB Developer Kit Super with NVMe SSD:
- Fresh flash with `initrd-flash --erase-nvme`
- SSH login via password works on first boot
- Host keys persist across reboots (stored on /var)
- avahi-daemon starts successfully, device discoverable via mDNS
- `/var` expanded from 109MB to full NVMe partition size on first boot
- Root home at `/var/roothome` is writable, persists across reboots

### DIFF
--- a/meta-avocado-distro/recipes-avocado/avocado-users/avocado-users_1.0.bb
+++ b/meta-avocado-distro/recipes-avocado/avocado-users/avocado-users_1.0.bb
@@ -13,7 +13,12 @@ SRC_URI = " \
   file://shadow \
   file://group \
   file://gshadow \
+  file://var-resize.service \
+  file://var-roothome-init.service \
 "
+
+inherit systemd
+SYSTEMD_SERVICE:${PN} = "var-resize.service var-roothome-init.service"
 
 do_install() {
     install -d ${D}${sysconfdir}
@@ -30,5 +35,19 @@ do_install() {
         # Use default shadow file (root login disabled)
         install -m 0644 ${WORKDIR}/shadow ${D}${sysconfdir}/shadow
     fi
+
+    # sshd privilege separation directory
+    install -d -m 0755 ${D}/var/empty/sshd
+
+    # Root home on writable /var partition (read-only rootfs)
+    install -d -m 0700 ${D}/var/roothome
+
+    # /var/log for lastlog and journal
+    install -d -m 0755 ${D}/var/log
+
+    # Systemd services for first-boot setup
+    install -d ${D}${systemd_system_unitdir}
+    install -m 0644 ${WORKDIR}/var-resize.service ${D}${systemd_system_unitdir}/
+    install -m 0644 ${WORKDIR}/var-roothome-init.service ${D}${systemd_system_unitdir}/
 }
 do_install[vardeps] += "AVOCADO_DEV_ROOT_LOGIN"
--- a/meta-avocado-distro/recipes-avocado/avocado-users/files/group
+++ b/meta-avocado-distro/recipes-avocado/avocado-users/files/group
@@ -5,3 +5,5 @@ systemd-resolve:x:995:
 systemd-journal:x:996:
 messagebus:x:999:
 avocado:x:997:
+sshd:x:74:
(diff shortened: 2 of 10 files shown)


mobileoverlord:
A lot of this is handled already, but at different layers of the system. The meta-avocado repo is used primarily as a package builder for the distro feeds to be consumed by the cli tools to compose runtimes. I'll explain, but addressing some of your concerns. 

1. Missing sshd/avahi system users

We removed as many system users as possible form the baseline users and groups to try to address issues with passwd / group not supporting a layered extensibility. System users / groups are managed on the system using single file databases. IE `/etc/{passwd|group}`. To fix this issue, meta-avocado implements a global user class which builds upon the yocto useradd* class which maps user / group additions to a package generated systemd-sysuser conf file which is included in each package in `/usr/lib/sysusers.d/<package>`. The use of sysusers addresses the single file database issue and allows multiple extensions / packages to contribute to adding users and groups without clobbering each other. 

2. SSH host keys written to read-only /etc/ssh

I'm interested in adjusgint eh default package config to set the storage paths to var. Today, we ship our own confi file in the avocado-ext-ssh* extensions, For example https://github.com/avocado-linux/avocado-os/blob/main/extensions/sshd-dev/overlay/etc/ssh/sshd_config#L20C1-L22C42 In service of keeping the meta layers for package building, we try to push as much to the extension layer as possible. 

3. Socket-activated sshd silently fails

I am not experiencing this when building through the CLI. I think this is a side effect of other ways you are testing. 

4. Root home on read-only filesystem

We should address this. I think we should rally around user homes being held in `/var/home`, we would need to update avocado CLI to create this directory for users. 

5. `/var` btrfs not expanded to fill partition

We handle this in the avocado-ext-dev extenion today, but this could be bubbled up to a different extension. https://github.com/avocado-linux/avocado-os/blob/main/extensions/dev/avocado.yaml#L17-L18 

6. No root home initialization

This is interesting, but we would also need to do this in the CLI layer since the var part is generated at CLI build and the dummy var image in meta-avocado is not used.  

Overall, you can see for a variety of reasons we try to keep the yocto layer free of opinionation. If we bake too much in to the package feeds, users won't be able to compose the system without certain opinions.

PR #209, opened 2026-04-08
Add synaptics astra 1680 support

eichenberger:
This PR will add support for the Grinn Astra 1680 SBC (https://grinn-global.com/products/grinn-astraonesbc-1680).

It is a draft because the naming is probably not finalised yet. What would be the preferred way? Currently, it is 'grinn-astra-1680-sbc', but I suppose it should also include 'synaptics'. Some proposals:
- synaptics-grinn-astra-1680-sbc
- astra-one-1680-sbc
- synaptics-1680-grinn-astra-one-sbc
I'm open to other ideas. 

I don't like my current hacks for changing the 'MACHINE' in recipes that require it to be 'grinn-astra-1680-sbc', such as 'meta-avocado-synaptics/recipes-kernel/linux/linux-syna_%.bbappend' or 'meta-avocado-synaptics/recipes-rescue/astra-rescue/synasdk-config-native_%.bbappend'. The problem lies in how Grinn uses the MACHINE to find the correct files, for example in https://github.com/grinn-global/meta-grinn-astra/blob/scarthgap_6.12_v2.2.0/meta-grinn-astra-bsp/recipes-kernel/linux/linux-syna_%.bbappend. Have you encountered similar issues before?

I used the following branch to test building my local extensions and to start the test repository. A pull request will follow as soon as the naming has been finalised:
https://github.com/eichenberger/avocado-os/tree/add-grinn-astra-1680-sbc.

I used the following Avocado YAML to test the SDK and image generation. Currently I only tested SYNAIMG which is the image type that can be used to install it on a plain image. Once everything works, I will also test the fwup image:
https://github.com/eichenberger/avocado-examples


mobileoverlord:
Great work on this. I added a bit of extras and merged manually, your commits have been preserved. 

On one of your questions, a lot of other vendors will trigger OVERRIDES on MACHINE name and since we want to own all the configs, we prefix with avocado- and typically add the upstream name back to the OVERRIDES so it gets the same treatment.

PR #211, opened 2026-04-29
feat: enable netfilter/iptables kernel config for NXP iMX targets

ndivecha:
Adds avocado-netfilter.cfg fragment and wires it into both linux-imx and linux-fslc-imx bbappends for all NXP i.MX boards (imx8mp-evk, imx91-frdm, imx93-evk, imx93-frdm).

The fragment enables:
- Core netfilter framework (CONFIG_NETFILTER, NF_CONNTRACK, NF_TABLES, NF_NAT) required for any iptables/nftables use
- IPv4 iptables: filter, NAT, MASQUERADE, MANGLE, REJECT
- IPv6 ip6tables: equivalent filter, NAT, MASQUERADE, MANGLE, REJECT
- nftables IPv4/IPv6 table support (NF_TABLES_IPV4)
- xt_* match/target extensions: connmark, connlimit, conntrack, devgroup, iprange, limit, mac, mark, state, string, tcpmss, hmark, led, log, netmap, nflog, redirect, tcpmss
- IPVS for load balancing and advanced NAT scenarios
- Multicast, advanced routing, and policy routing (IP_MULTIPLE_TABLES)
- conntrackd support via NETFILTER_NETLINK_ACCT and NF_CT_NETLINK

All driver/extension options are built as modules (=m) so they appear as individual kernel-module-* packages in the rolling feed and can be installed on demand. Framework and routing options that cannot be modules are set =y.


mobileoverlord (approved):
Thanks!

PR #219, opened 2026-05-20
meta-avocado-qemu: simplify swtpm setup and add qemu-provision overlay

jetm:
## Problem

`run-qemux86-64-swtpm` orchestrates swtpm and QEMU by creating a tmux
session with two panes - one running `start-swtpm`, the other waiting for
the socket then launching QEMU. This requires tmux to be installed and
forces the user into a managed terminal session they did not ask for. It
also splits swtpm lifecycle management across two scripts, making the
startup sequence fragile and harder to reason about.

The `do_stone_provision` task that produces the bootable disk image is
excluded from the default build (`STONE_BUILD_TASK = "do_stone_bundle"`)
to keep CI build times down. There is no documented way for a developer
to opt into the full disk image build for local QEMU use.

## Solution

Collapse swtpm management into a single self-contained script. swtpm is
now started with `--daemon`, which keeps it alive in the background without
a terminal multiplexer. The script kills any leftover swtpm instance from
a previous run at startup to avoid stale socket conflicts, then polls the
control socket for up to 10 seconds before starting QEMU.

Native binaries (qemu-system-x86_64, qemu-img, swtpm) are invoked through
the uninative glibc loader using the same mechanism bitbake uses internally,
rather than relying on `LD_LIBRARY_PATH`.

Add `kas/target/qemu-provision.yml` as an opt-in overlay that promotes
`STONE_BUILD_TASK` to `do_stone_provision`, producing the bootable `.img`
the run scripts expect. This is explicitly not intended for CI.

## Key changes

- `run-qemux86-64-swtpm`: replace tmux two-pane session with `swtpm --daemon`;
  poll socket up to 10 s; kill stale swtpm on startup; add `set -euo pipefail`;
  add EXIT/INT/TERM trap to kill swtpm when QEMU exits
- `run-qemux86-64` and `run-qemux86-64-swtpm`: invoke native binaries via
  uninative loader (`run_native()` helper); add BUILD_DIR sanity check
- `find-library-path`: deleted (replaced by inline uninative loader setup)
- `kas/target/qemu-provision.yml`: new opt-in overlay for local QEMU disk
  image builds; enables `AVOCADO_DEV_ROOT_LOGIN` for convenience

## Reviewer notes

Full local QEMU boot with software TPM 2.0:

```sh
# Build the bootable disk image (promotes do_stone_bundle -> do_stone_provision)
kas build kas/machine/qemux86-64.yml:kas/target/qemu-provision.yml

# Boot with software TPM - no tmux needed
./run-qemux86-64-swtpm
```

`qemu-provision.yml` adds ~2 min to the build for the 1 GB image
assembly. It is documented as local-only in the overlay header.


mobileoverlord (approved):
🎉 Thanks!

PR #220, opened 2026-05-22
feat(networkmanager): enable modemmanager support and package networkmanager-wwan

ndivecha:
Without PACKAGECONFIG += "modemmanager wwan", NM is built with -Dmodem_manager=false and libnm-device-plugin-wwan.so is not compiled. At runtime NM logs 'wwan plugin not available' and marks any wwan interface as a generic unmanaged device, preventing autoconnect of GSM/LTE profiles even when ModemManager is running.

Add a distro-level bbappend that unconditionally enables both flags:
- modemmanager: passes -Dmodem_manager=true to meson, builds the wwan plugin, adds build/runtime deps on modemmanager and mobile-broadband-provider-info.
- wwan: packaging flag that puts libnm-device-plugin-wwan.so and libnm-wwan.so into networkmanager-wwan and adds it to RRECOMMENDS of the networkmanager meta-package.

Also add networkmanager-wwan to packagegroup-avocado-extra so the plugin is explicitly included in the feed's package set.

### DIFF
--- a/meta-avocado-distro/recipes-connectivity/networkmanager/networkmanager_%.bbappend
+++ b/meta-avocado-distro/recipes-connectivity/networkmanager/networkmanager_%.bbappend
@@ -1,0 +1,15 @@
+# Enable ModemManager integration and package the wwan device plugin.
+#
+# Without modemmanager in PACKAGECONFIG, NM is built with -Dmodem_manager=false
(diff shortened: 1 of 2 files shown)


mobileoverlord (approved):
🎉

PR #222, opened 2026-06-10
Add kas opt-out to disable browsers/qt in dev builds

jetm:
## Problem

`chromium-ozone-wayland` and `qtwebengine` are the two heaviest recipes in the tree (both Chromium-sized), pulled into `packagegroup-avocado-extra` on every opengl+wayland target via `BROWSER_PACKAGES` / `QT5_PACKAGES`. Since `avocado-complete` builds the extras, every such target build and CI pays both compiles - even for dev images that never ship them. Clients do use these packages, so they must keep building by default; what's missing is a way to skip them for a minimal dev build.

## Solution

Keep the default build unchanged (chromium and Qt still build for clients) and add two opt-in kas overlays that strip those sets from `packagegroup-avocado-extra` for dev images, packaging the `RDEPENDS:...:remove` pattern contributors already hand-roll.

## Key changes

- `kas/feature/no-browsers.yml` - removes `chromium-ozone-wayland` + `wpewebkit`/`wpebackend-fdo`/`cog`.
- `kas/feature/no-qt.yml` - removes the Qt5 stack including `qtwebengine`.
- No change to `packagegroup-avocado-extra` or any default build - the overlays are opt-in only.

## Usage

```text
kas build kas/machine/<machine>.yml:kas/feature/no-browsers.yml:kas/feature/no-qt.yml
```

## Reviewer notes

Reworked from the original "drop Chromium by default" approach after @mobileoverlord's review (clients need it built by default) and @eichenberger's `RDEPENDS:...:remove` example. Default builds are unchanged; the overlays only take effect when explicitly included. `no-qt.yml` mirrors `QT5_PACKAGES` and carries a sync-reminder comment.

### DIFF
--- a/kas/feature/no-browsers.yml
+++ b/kas/feature/no-browsers.yml
@@ -1,0 +1,11 @@
+header:
+  version: 16
+
+# Opt-in overlay for minimal dev images: drop the browser engines from
+# packagegroup-avocado-extra so avocado-complete does not pay the Chromium/WPE
+# compile (the heaviest recipes in the tree). The default build keeps them -
+# clients use them - so include this overlay only for a dev build, e.g.:
+#   kas build kas/machine/<machine>.yml:kas/feature/no-browsers.yml
+local_conf_header:
+  no-browsers: |
+    RDEPENDS:packagegroup-avocado-extra:remove = " chromium-ozone-wayland wpewebkit wpebackend-fdo cog"
--- a/kas/feature/no-qt.yml
+++ b/kas/feature/no-qt.yml
@@ -1,0 +1,15 @@
+header:
+  version: 16
+
+# Opt-in overlay for minimal dev images: drop the Qt5 stack from
+# packagegroup-avocado-extra, including qtwebengine - a second Chromium-sized
+# build. The default build keeps Qt - clients use it - so include this overlay
+# only for a dev build, e.g.:
+#   kas build kas/machine/<machine>.yml:kas/feature/no-qt.yml
+#
+# The list mirrors QT5_PACKAGES in
+# meta-avocado/recipes-avocado/packagegroups/packagegroup-avocado-extra.bb;
+# keep it in sync if that recipe's list changes.
+local_conf_header:
+  no-qt: |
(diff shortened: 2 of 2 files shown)


mobileoverlord:
we need to still build this package, clients use it, but it would be nice if we had the ability to build it optionally for dev builds. We have a target for avocado-core which builds everything except the extras repos, but its not enough for us to do a minimal dev image. I'd be interested in a way we can disable browsers or disable qt etc from kas builds.

eichenberger:
I'm using a custom kas file to do my customization (e.g. removing chromium). For example for imx95 I just created a kas-imx95-frdm.yaml in the top dir and then I use that:

```yaml
header:
  version: 16
  includes:
    - repo: meta-avocado
      file: kas/machine/imx95-frdm.yml

repos:
  meta-avocado:
    path: meta-avocado

local_conf_header:
  machine-avocado-imx95-frdm: |
    RDEPENDS:packagegroup-avocado-extra:remove = " chromium-ozone-wayland wpewebkit wpewebkit-fdo nodejs cog"
    AVOCADO_RM_WORK = "1"
```

Maybe that helps, else just ignore it.

jetm:
Reworked per the feedback - thanks both.

@mobileoverlord you're right that chromium has to stay in the default build for clients, so the default is unchanged (chromium + Qt still build). Instead I added two opt-in kas overlays for dev images:

- `kas/feature/no-browsers.yml` - `chromium-ozone-wayland` + `wpewebkit`/`wpebackend-fdo`/`cog`
- `kas/feature/no-qt.yml` - the Qt5 stack, including `qtwebengine` (the other Chromium-sized build)

A minimal dev image is now:

```text
kas build kas/machine/<machine>.yml:kas/feature/no-browsers.yml:kas/feature/no-qt.yml
```

so you get the per-component disable (browsers / qt / etc) without changing the default.

@eichenberger that `RDEPENDS:packagegroup-avocado-extra:remove` you showed is exactly the mechanism - I packaged it as these reusable overlays so it's a one-line include instead of a hand-rolled kas file. Thanks for the pointer.

mobileoverlord:
@jetm and I chatted about this, we are going to invert the style and instead of having duplication in kas to what we want to opt out of, we are going to have kas declate categories of content that the user can compose together with kas configurations to opt in to building more of the complete distro.

jetm:
Superseded by #229, which inverts this to opt-in feature groups driven by AVOCADO_FEATURE_GROUPS rather than opt-out remove overlays. The opt-in model scales to every capability instead of just chromium/Qt, and clients keep building the full set by default via kas/feature/complete.yml. Closing in favor of that approach; tracked in ENG-2003.

PR #226, opened 2026-06-15
Add avocado-rzv2h-rdk (Renesas RZ/V2H RDK) board

jetm:
> Draft / WIP - builds are validated; the board has not been booted on hardware yet.

## Problem

Avocado OS has no machine for the Renesas RZ/V2H Robot Development Kit
(WS125-V2HRDKREFZ, SoC `r9a09g057h44`). The BSP carries RZ/V2H EVK support but
not this board.

## Solution

Add `avocado-rzv2h-rdk`, derived from `rzv2n-sr-som`, on linux-renesas 6.12 +
TF-A v2h + u-boot 2024.07. Boot/provisioning are SD-only (the RDK has no
USB-device port). A first-cut RDK device tree is layered onto the kernel; u-boot
runs on the interim EVK config, patched to boot Avocado's extlinux layout. Also
adds a board-agnostic `kas/target/bringup.yml` overlay for a minimal, debuggable
bring-up image.

## Key changes

- New machine `avocado-rzv2h-rdk.conf` + `rzv2h-family.inc` (linux-renesas 6.12,
  `TFA_PLATFORM=v2h`) and the kas machine yaml.
- First-cut `r9a09g057h44-rzv2h-rdk.dts` (SD0 pinmux TRM-confirmed) + dtb
  Makefile registration; interim EVK u-boot defconfig.
- extlinux + stone SD/serial provisioning; u-boot patched to `sysboot` the
  extlinux.conf instead of the vendor `ext4load`/`booti` flow.
- u-boot: drop inapplicable fastboot (no USB-device port) to fit
  `BOARD_SIZE_LIMIT`.
- Board-agnostic `kas/target/bringup.yml` (`avocado-image-rootfs` +
  `debug-tweaks` + `kernel-modules` + bring-up tools; image-scoped, no
  packagegroup changes). Build:
  `kas build kas/machine/rzv2h-rdk.yml:kas/target/bringup.yml`.

## Reviewer notes (WIP)

- Validated: full image build (5217 tasks, 0 errors), dts compiles, u-boot
  builds clean multicore, manifest complete (systemd/serial-getty/sshd, 264
  kernel-modules, bring-up tools), extlinux PARTUUID matches stone `rootfs-a`.
- Not yet verified: boot-to-login on hardware. The u-boot extlinux
  `CONFIG_BOOTCOMMAND` patch is build-verified only; `sysboot` path/prefix, mmc
  index and load addresses need confirmation at the u-boot prompt.
- Interim: kernel uses the RDK dtb but u-boot is on the EVK config
  (`rzv2h-evk-ver1`, `TFA_BOARD=evk_1`) until a dedicated RDK defconfig lands.
  DTB is first-cut (SD0/console only); CAN-FD/GPIO/PCIe/USB3/GbE not ported;
  DRP-AI3 not wired.
- `debug-tweaks` = passwordless root (bring-up convenience; drop before
  shipping).
- Branch is a few commits behind `scarthgap`; will rebase before marking ready.


meta-avocado-renesas/conf/machine/avocado-rzv2h-rdk.conf
```diff
@@ -0,0 +1,78 @@
...
+MACHINE_EXTRA_RRECOMMENDS:append = " kernel-modules \
+	ethtool \
+	memtester \
+	stress-ng \
+	iperf3 \
+	libgpiod \
+	libgpiod-tools \
+	pciutils \
+	pciutils-ids \
+	usbutils"
```
nicksinas, line 78: **1 · High — `MACHINE_EXTRA_RRECOMMENDS` is dead config in Avocado; these 10 packages reach neither the image nor the feed.**

`MACHINE_EXTRA_RRECOMMENDS` is only consumed by `packagegroup-base`, which Avocado doesn't use. `packagegroup-avocado-rootfs.bb:12-38` documents at length that it deliberately does *not* RDEPEND on `packagegroup-core-boot` and drops `MACHINE_ESSENTIAL_EXTRA_RDEPENDS`; `packagegroup-avocado-solidrun-extra.bb` says the same thing explicitly for the sibling board ("the Renesas-style `MACHINE_ESSENTIAL_EXTRA_RDEPENDS` ... is a no-op in avocado-os's image build (we don't use core-image.bbclass)").

So `kernel-modules`, `ethtool`, `memtester`, `stress-ng`, `iperf3`, `libgpiod{,-tools}`, `pciutils{,-ids}`, `usbutils` are silently absent from both the rootfs and the package feed.

Second-order: this is exactly the gap that `kas/target/bringup.yml` in this same PR then re-lists by hand (`ethtool`, `libgpiod`, `libgpiod-tools`, `usbutils`, `pciutils`, ...). A whole new kas overlay is being added to work around dead config introduced two commits earlier.

Fix: move these into `packagegroup-avocado-rzv2h-rdk-extra` (feed, already wired via `PKG_EXTRA_INSTALL`) and/or `ROOTFS_IMAGE_EXTRA_INSTALL` for the image, and delete this block.

meta-avocado-renesas/recipes-bsp/u-boot/u-boot/0001-rzv2h-dev-boot-avocado-extlinux.patch
```diff
@@ -0,0 +1,46 @@
...
+@@ -80,7 +80,11 @@
+ 	"bootcmd_check=if mmc dev 1; then run sd1load; else run sd0load; fi \0"
+ #endif
+ 
+-#define CONFIG_BOOTCOMMAND	"env default -a;run bootcmd_check;run bootimage"
++#define CONFIG_BOOTCOMMAND \
++	"setenv scriptaddr 0x50000000; setenv kernel_addr_r 0x48080000; " \
++	"setenv fdt_addr_r 0x48000000; setenv ramdisk_addr_r 0x4a000000; " \
++	"sysboot mmc 1:1 any ${scriptaddr} /extlinux/extlinux.conf; " \
++	"sysboot mmc 0:1 any ${scriptaddr} /extlinux/extlinux.conf"
```
nicksinas, line 43: **2 · High — this replaces the repo's distroboot policy with a hardcoded, A-slot-only bootcmd.**

`meta-avocado-renesas/recipes-bsp/u-boot/u-boot/env-nowhere.cfg:10` already ships a board-agnostic boot policy for every Renesas machine:

```
CONFIG_BOOTCOMMAND="setenv boot_targets mmc0 mmc1; run distro_bootcmd; fastboot usb 0"
```

and `docs/adding-a-machine-target.md` §6 states the design intent: "Boot partitions use type `EF00` (EFI System) **so the bootloader's distroboot scan finds them by GUID**". The only reason that fragment has no effect here is the vendor `#define CONFIG_BOOTCOMMAND` in `include/configs/rzv2h-dev.h` shadowing the Kconfig symbol.

Substituting a second hardcoded command has three consequences:

1. **A/B is defeated.** `sysboot mmc 1:1` / `mmc 0:1` pins partition 1 = `boot-a`. The manifest declares `update_strategy: gpt-ab` and allocates `boot-b` (`stone-rzv2h-rdk.json:91`), which can now never be booted or rolled back into. `distro_bootcmd` scans all bootable partitions and would find both.
2. **Board divergence.** This is the only Avocado board whose boot path is a source patch to a vendor board header rather than a Kconfig fragment, so it won't inherit future changes to the shared policy.
3. Load addresses become board-header state instead of env (see next comment).

Suggested fix: patch the vendor `#define` **away** (delete it, or `#undef CONFIG_BOOTCOMMAND`) so the existing `env-nowhere.cfg` value applies, instead of replacing it. If `distro_bootcmd` isn't reachable in this board's config, that's worth saying in the patch header — it's the reason the whole rzv2n path works.

nicksinas (commented):
Review of the RZ/V2H RDK port. Findings 1-15 of a longer pass are inline below; the remaining lower-severity items (docs table not updated, patch header hygiene, dead `KERNEL_DTC_FLAGS -@`/`Image.gz`, `rzv2h-family.inc` values clobbered by the machine conf, non-conventional board `compatible`, missing `bsp/rzv2h-rdk/avocado.yaml` companion change, `bringup.yml` tool naming and `INIT_MANAGER` override) are held back to keep this readable.

Acknowledging this is a draft with hardware boot still pending. Flagging what I'd treat as blocking before it goes ready for review:

- **#1** `MACHINE_EXTRA_RRECOMMENDS` is dead config in Avocado - the board's utility list reaches neither the image nor the feed, and `bringup.yml` in this same PR works around it.
- **#2** the u-boot bootcmd patch abandons the repo's `distro_bootcmd` policy and pins partition 1, so `update_strategy: gpt-ab` + `boot-b` can never be booted.
- **#3** `kernel_addr_r` -> `ramdisk_addr_r` leaves 31.5 MiB for an `Image` that is typically 30-40 MiB.
- **#4** DRAM nodes and `TFA_BOARD=evk_1` are both inherited from the EVK and this is the one gap not recorded in any TODO.
- **#12** five files rzv2n consumes are modified; please add the `build-rzv2n-sr-som` CI label.

Cheapest high-leverage cleanups: **#6** (a shared stone override dir, per the existing `meta-avocado-nxp/stone/avocado-imx/` pattern, deletes 395 duplicated lines), **#7** (one line replaces the `EXTLINUX_PROVIDER` table in two bbappends plus two near-identical recipes), **#8** (use the `AVOCADO_DEV_ROOT_LOGIN` knob that already exists and already warns).

Two things I checked that are *not* problems, for the record: the `;`-chained `sysboot` fallback does run after a failure (`cli_simple_run_command` continues past a failing `;` token), and the absolute paths `/Image` / `/r9a09g057h44-rzv2h-rdk.dtb` resolve correctly against the FAT root because `do_sysboot` passes `allow_abs_path=true`. The reused `var` PARTUUID is also correct - docs section 6 says it's canonical.

meta-avocado-renesas/recipes-bsp/u-boot/u-boot/0001-rzv2h-dev-boot-avocado-extlinux.patch
```diff
@@ -0,0 +1,46 @@
...
+--- a/include/configs/rzv2h-dev.h
++++ b/include/configs/rzv2h-dev.h
+@@ -80,7 +80,11 @@
+ 	"bootcmd_check=if mmc dev 1; then run sd1load; else run sd0load; fi \0"
+ #endif
+ 
+-#define CONFIG_BOOTCOMMAND	"env default -a;run bootcmd_check;run bootimage"
++#define CONFIG_BOOTCOMMAND \
++	"setenv scriptaddr 0x50000000; setenv kernel_addr_r 0x48080000; " \
++	"setenv fdt_addr_r 0x48000000; setenv ramdisk_addr_r 0x4a000000; " \
```
nicksinas, line 41: **3 · High — `kernel_addr_r` → `ramdisk_addr_r` leaves only 31.5 MiB for the kernel; a normal-sized `Image` overwrites the initrd.**

`kernel_addr_r=0x48080000`, `ramdisk_addr_r=0x4a000000` → a `0x1F80000` (31.5 MiB) window. `boot/pxe_utils.c` `label_boot()` loads kernel, then initrd, then fdt at these addresses with no overlap checking. An arm64 `Image` from a cip vendor config is routinely 30-40 MiB.

The "addresses match the vendor booti" justification doesn't transfer: the vendor flow (`ext4load` + `booti`) loaded **no ramdisk**, so this gap was never exercised. Avocado's layout always has one (`avocado-image-initramfs-rzv2h-rdk.cpio.zst`, `extlinux.conf:8`).

Failure mode is a corrupted kernel or initrd with no clear error — exactly the class of bug that eats a hardware bring-up day.

Check `ls -l tmp/deploy/images/avocado-rzv2h-rdk/Image` and raise `ramdisk_addr_r` (and `scriptaddr` above it) to give real headroom.

meta-avocado-renesas/recipes-kernel/linux/files/r9a09g057h44-rzv2h-rdk.dts
```diff
@@ -0,0 +1,142 @@
...
+	memory@48000000 {
+		device_type = "memory";
+		/* first 128MB is reserved for secure area. */
+		reg = <0x0 0x48000000 0x1 0xF8000000>;
+	};
+
+	memory@240000000 {
+		device_type = "memory";
+		reg = <0x2 0x40000000 0x2 0x00000000>;
+	};
```
nicksinas, line 54: **4 · High — DRAM topology is copied verbatim from the EVK and is not in the TODO list.**

These two nodes are byte-identical to mainline v6.12 `arch/arm64/boot/dts/renesas/r9a09g057h44-rzv2h-evk.dts`: `0x48000000 + 0x1F8000000` and `0x240000000 + 0x200000000`, i.e. ~16 GiB across two banks.

Combined with `avocado-rzv2h-rdk.conf:28` `TFA_BOARD ?= "evk_1"`, BL2 also runs the **EVK's DDR init**. If the RDK's LPDDR differs in size, rank layout or part, the outcomes are (a) BL2 fails to bring up DRAM → no boot at all, nothing on the console past TF-A, or (b) the kernel maps absent memory → random corruption under load.

This is the single highest-risk unverified item in the PR, and it's the one gap **not** recorded: the dts TODO at lines 18-20 covers CAN-FD, the 40-pin header, PCIe, USB3.2 and GbE, and the conf TODOs cover TF-A board variant and u-boot defconfig, but nothing mentions memory.

Please add it to the TODO block and confirm both nodes against the RDK BOM/hardware manual before first boot.

meta-avocado-renesas/conf/machine/avocado-rzv2h-rdk.conf
```diff
@@ -0,0 +1,78 @@
...
+KERNEL_DTC_FLAGS += "-@"
+
+# RDK device tree, forward-ported from Renesas-SST/linux-rz (branch
+# ubuntu/rz-v2h-rdk, rzv2h-rdk-ver1.dts) onto linux-renesas 6.12 via the
+# linux-renesas bbappend.
+KERNEL_DEVICETREE = "renesas/r9a09g057h44-rzv2h-rdk.dtb"
+
+# Machine configuration. The RDK base board exposes no onboard WiFi/BT and no USB
+# gadget, so drop the SolidRun usbgadget/wifi/bluetooth features.
+MACHINE_FEATURES = "opengles opencl mali-g31 rtc serial usbhost"
```
nicksinas, line 59: **5 · Medium — `MACHINE_FEATURES` advertises hardware the first-cut DT doesn't enable.**

`opengles opencl mali-g31` plus `rzv2h-family.inc:2` `SOC_FAMILY = "rzv2h-family:mali-family"` pull the Renesas/mali GPU userspace and advertise EGL/GLES/OpenCL to anything that queries `MACHINE_FEATURES`. But this PR restricts both GPU dts patches to `:rzv2n-sr-som` (`linux-renesas_%.bbappend:10-11`), and the RDK dts enables no GPU node — mainline `r9a09g057.dtsi` has no GPU node at all, so there is nothing for panfrost/mali to bind to.

Same for `rtc` (no RTC node in the dts; mainline's EVK doesn't enable one either) and `usbhost` (USB3.2 explicitly listed as not ported, dts:18-20).

Result: applications get a GPU/OpenCL stack that fails at runtime, and image size grows for capability the board doesn't have yet. Suggest trimming to what the DT actually enables (`serial`) and re-adding each feature in the commit that ports its peripheral — that also makes the DT gaps self-documenting.

meta-avocado-renesas/stone/rzv2h-rdk/build-disk-image.sh
```diff
@@ -0,0 +1,176 @@
+#!/usr/bin/env bash
+#
+# Shared helper for the rzv2n-sr-som emmc and sd provisioning profiles.
```
nicksinas, line 3: **6 · Medium — all three scripts in this directory are byte-identical copies of the rzv2n set, and still describe rzv2n.**

```
4b400cc450c0420e54bbff767efdeec0  stone/rzv2h-rdk/build-disk-image.sh
4b400cc450c0420e54bbff767efdeec0  stone/rzv2n-sr-som/build-disk-image.sh
3f9b842258d10f844e719316a59f0978  stone/rzv2h-rdk/stone-provision-sd.sh
3f9b842258d10f844e719316a59f0978  stone/rzv2n-sr-som/stone-provision-sd.sh
0935fc161bb4de304256462a8ca4833a  stone/rzv2h-rdk/stone-provision-serial.sh
0935fc161bb4de304256462a8ca4833a  stone/rzv2n-sr-som/stone-provision-serial.sh
```

395 duplicated lines covering the SD-write and SPI-NOR flash paths — the two places where a bug destroys a card or bricks a board. A fix applied to one board now silently misses the other. Header comments still say "for rzv2n-sr-som" / "rzv2n-sr-som emmc and sd provisioning profiles" (`build-disk-image.sh:3`, `stone-provision-sd.sh:3`, `stone-provision-serial.sh:3`), which will mislead the next reader about which board they're editing.

The repo already has the fix pattern. `meta-avocado-nxp/conf/machine/include/avocado-imx.inc:1` sets `MACHINEOVERRIDES =. "avocado-imx:"`, and a single `meta-avocado-nxp/stone/avocado-imx/` directory serves all four i.MX machines — `meta-avocado/classes/stone.bbclass` puts `<layer>/stone/<override>` on FILESPATH, so any shared override name works. `meta-avocado-synaptics/stone/avocado/` does the same.

Suggest a shared renesas override (e.g. `MACHINEOVERRIDES =. "avocado-rz:"` in a common `.inc`) plus `meta-avocado-renesas/stone/avocado-rz/`, and delete both per-machine copies.

(For contrast: the `avocado-build-*`/`avocado-provision-*` SDK hook copies in this PR are fine — docs §5 explicitly endorses that duplication, and 8 other layers do it, because the base recipe's SRC_URI hardcodes `${MACHINE_SHORT_NAME}` in the filename.)

meta-avocado-renesas/recipes-avocado/distro/avocado-stone.bbappend
```diff
@@ -3,7 +3,11 @@
 do_compile[depends] += "firmware-pack:do_deploy"
 do_compile[depends] += "flash-writer:do_deploy"
 do_compile[depends] += "u-boot:do_deploy"
-do_compile[depends] += "extlinux-rzv2n-sr-som:do_deploy"
+# Per-machine extlinux provider (board id != ${MACHINE}); gate on MACHINEOVERRIDES.
+EXTLINUX_PROVIDER ?= ""
+EXTLINUX_PROVIDER:rzv2n-sr-som = "extlinux-rzv2n-sr-som"
+EXTLINUX_PROVIDER:rzv2h-rdk = "extlinux-rzv2h-rdk"
+do_compile[depends] += "${EXTLINUX_PROVIDER}:do_deploy"
```
nicksinas, line 10: **7 · Medium — the `EXTLINUX_PROVIDER` table is unnecessary and fails open.**

The provider name is exactly `extlinux-${MACHINE_SHORT_NAME}` for both boards (`extlinux-rzv2n-sr-som`, `extlinux-rzv2h-rdk`), and `MACHINE_SHORT_NAME` is already `${MACHINE}` minus the `avocado-` prefix (`meta-avocado/conf/distro/avocado.inc:20`). So this whole block, duplicated in `avocado-sdk-target.bbappend:8-13`, collapses to one line that needs no edit per board:

```bitbake
do_compile[depends] += "extlinux-${MACHINE_SHORT_NAME}:do_deploy"
```

As written it fails open rather than loud: a third Renesas machine that misses either file expands to `":do_deploy"` — a malformed-dependency parse error that doesn't point at "add your board to the table in two bbappends".

Going one step further: `extlinux-rzv2h-rdk_1.0.bb` and `extlinux-rzv2n-sr-som_1.0.bb` are identical except `SUMMARY` and `COMPATIBLE_MACHINE`. A single `extlinux-avocado` recipe with `extlinux.conf` under a machine-override files dir removes both recipes *and* this variable, and the next Renesas board needs only a 9-line conf file.

kas/target/bringup.yml
```diff
@@ -0,0 +1,71 @@
...
+    # The avocado distro already sets INIT_MANAGER = "systemd"; assert it here
+    # so a bringup image is correct even on a board that overrides it.
+    INIT_MANAGER = "systemd"
+
+    # Passwordless serial/root login for a fresh board with no provisioned
+    # credentials. Bringup convenience - drop before ship. No ssh-server-*
+    # feature here: avocado-image-rootfs uses image.bbclass, not
+    # core-image.bbclass, so those features are undefined; the sshd ships via
+    # the openssh package in the install list below instead.
+    EXTRA_IMAGE_FEATURES:append = " debug-tweaks"
```
nicksinas: **8 · Medium — reinvents passwordless root instead of using the existing knob, and `EXTRA_IMAGE_FEATURES` is global.**

The repo already has a first-class mechanism: `AVOCADO_DEV_ROOT_LOGIN` (`meta-avocado-distro/recipes-avocado/avocado-users/avocado-users_1.0.bb:25-31`), used by the sibling overlay `kas/target/qemu-provision.yml:20`. It edits `/etc/shadow` at `do_install` and — importantly — emits `bbwarn "Root login with empty password is enabled. This is insecure and should only be used for development!"` on every build.

`debug-tweaks` is worse on three counts:

1. `EXTRA_IMAGE_FEATURES` is **global**, not scoped to `avocado-image-rootfs`. It also applies to `avocado-image-initramfs` and any other image in the build.
2. It expands to `allow-empty-password allow-root-login empty-root-password post-install-logging` — broader than the intended empty root password, and now two mechanisms write `/etc/shadow` (avocado-users at `do_install`, the `empty-root-password` postcommand on the assembled rootfs).
3. No warning, and nothing marks the artifact. A bringup rootfs is byte-indistinguishable from a release one, so "drop before ship" is enforced by a comment only.

Suggest `AVOCADO_DEV_ROOT_LOGIN = "1"` instead, which gets the warning for free. If you want belt-and-braces, stamp something into `/etc/os-release` or `IMAGE_NAME` so a bringup image can't be mistaken for a release build.

kas/target/bringup.yml
```diff
@@ -0,0 +1,71 @@
...
+#
+# Passing an explicit target (bakar build --target X, or bakar bitbake X)
+# overrides the `target:` below.
+
+header:
+  version: 16
+
+# Replace avocado-distro (extras + SDK) with the minimal bootable rootfs.
+target:
+  - avocado-image-rootfs
```
nicksinas: **9 · Medium — this target produces a rootfs, not something bootable or flashable; the header overstates it.**

The header says "minimal, fast, **bootable** image", "so a fresh board boots and is debuggable". But replacing `avocado-distro` with `avocado-image-rootfs` drops everything the board needs to actually boot:

- u-boot, TF-A/FIP — pulled by `avocado-stone`/`avocado-sdk-target`, not by the image
- `extlinux.conf`, the initramfs, the var btrfs, the stone bundle — `avocado-stone.bb:41-45` is what depends on `avocado-image-initramfs:do_image_complete`, `avocado-image-var:do_deploy`, `virtual/kernel:do_deploy`

Only `firmware-pack` and `flash-writer` survive, via `EXTRA_IMAGEDEPENDS` (image.bbclass wires those into `do_image_complete`).

So `kas build kas/machine/rzv2h-rdk.yml:kas/target/bringup.yml` yields an erofs rootfs with no way to provision it — the edit-build-flash loop this overlay targets still needs a full `avocado-distro` build. Either add the boot/stone deps to the bringup target (`avocado-stone` is probably the right single dep) or correct the header to say rootfs-only.

kas/target/bringup.yml
```diff
@@ -0,0 +1,71 @@
...
+    EXTRA_IMAGE_FEATURES:append = " debug-tweaks"
+
+    # All built kernel modules, so every on-board peripheral has its driver
+    # available during bringup. kernel-modules is the same meta-package the
+    # machine extras already use (packagegroup-avocado-rzv2h-rdk-extra,
+    # packagegroup-avocado-alif-extra). On a multi-kernel board (jetson,
+    # raspberrypi) prefer the version-qualified
+    # packagegroup-avocado-rootfs-modules-${KERNEL_VERSION} to avoid pulling
+    # off-kernel modules.
+    ROOTFS_IMAGE_EXTRA_INSTALL:append = " kernel-modules"
```
nicksinas: **10 · Medium — this comment describes a multi-kernel guard that the code doesn't implement.**

The comment says "On a multi-kernel board (jetson, raspberrypi) prefer the version-qualified `packagegroup-avocado-rootfs-modules-${KERNEL_VERSION}` to avoid pulling off-kernel modules" — but the overlay is board-agnostic and unconditionally appends the unqualified `kernel-modules`.

`packagegroup-avocado-rootfs.bb:40-48` spells out why that's harmful: several kernels each `Provide` the unqualified virtual, dnf NVR-tiebreaks across them, and it "non-deterministically pull[s] in a kernel generation other than the one targeted for this rootfs" — dragging the off-kernel `kernel-${other_kver}` base in with it.

So on jetson/rpi this overlay does the documented wrong thing while reading as if it doesn't, and the failure is non-deterministic. Either make it conditional on the board being single-kernel, or drop the claim and scope the overlay explicitly.

meta-avocado-renesas/recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -5,3 +5,19 @@ SRC_URI:append = " \
...
   file://fastboot.cfg \
 "
+
+# The RDK has no USB device/gadget port (USB3.2 are host-only, the Micro-B is a
+# UART), so u-boot fastboot cannot run here. Drop the shared fastboot fragment on
+# this machine: it is dead weight that also pushed u-boot-nodtb.bin past the
+# board's 0x100000 size limit. (env-nowhere.cfg keeps a harmless `fastboot`
+# fallback token in BOOTCOMMAND, but the board config header overrides
+# BOOTCOMMAND anyway, so nothing references the removed command at runtime.)
+SRC_URI:remove:rzv2h-rdk = "file://fastboot.cfg"
```
nicksinas, line 15: **11 · Medium — the fastboot removal and the bootcmd patch are silently coupled.**

`env-nowhere.cfg:10` still ends the shared Renesas `BOOTCOMMAND` with `fastboot usb 0`, and on this machine that command no longer exists. As the comment says, it's inert only because a `#define` in the board header shadows the Kconfig string — a load-bearing coupling asserted in a comment, enforced nowhere.

Concretely: if the bootcmd patch is reworked per finding #2 (letting the Kconfig `BOOTCOMMAND` apply, which is the cleaner fix), boot ends in `Unknown command 'fastboot' - try 'help'` after the mmc scan fails. Two changes that look independent are not.

Suggest gating the fastboot token in the fragment per machine (e.g. a `fastboot-fallback.cfg` alongside `fastboot.cfg`, both `rzv2n`-scoped) so the removal is self-consistent regardless of what happens to `CONFIG_BOOTCOMMAND`.

Also worth recording here: commit `69f2fbb` measured `u-boot-nodtb.bin` at `0x1008f0` against a `0x100000` limit — 2288 bytes over, resolved by deleting a feature. The board ships on the interim EVK defconfig at the edge of `BOARD_SIZE_LIMIT`, so the next fragment re-breaks the build, and that margin is invisible from the recipe. Worth a note next to the `UBOOT_CONFIG` TODO in the machine conf.

kas/vendor/renesas.yml
```diff
@@ -21,13 +21,11 @@ repos:
...
-    path: rzv2n_drp-ai_driver
-    branch: main
-    commit: bb3e6a9283e2dda2bb4a3c25986a0a309f1db905
-    layers:
-      meta-rz-features/meta-rz-drpai:
+# NOTE: SoC-specific DRP-AI driver layers (renesas-rz/rzv2n_drp-ai_driver,
+# rzv2h_drp-ai_driver) are intentionally NOT included here. They carry
+# machine-specific u-boot/kernel patches (e.g. the rzv2n ether-setting patch
+# touches include/configs/rzv2n-*.h) that break sibling Renesas machines if
+# applied globally. Wire each in its own machine kas instead.
```
nicksinas, line 28: **12 · Medium — five files that rzv2n consumes are changed; "unaffected" is asserted, not built.**

Changes in this PR that land on the existing rzv2n build:

1. this file — DRP-AI layer removed from the vendor kas
2. `kas/machine/rzv2n-sr-som.yml` — same layer re-added there, which changes its position in BBLAYERS
3. `linux-renesas_%.bbappend` — its two GPU dts patches move from an unconditional `SRC_URI:append` to `:rzv2n-sr-som`
4. `avocado-stone.bbappend` — extlinux dep now goes through `EXTLINUX_PROVIDER`
5. `avocado-sdk-target.bbappend` — same

The PR notes say "the rzv2n build is unaffected" and the validated list covers only the rzv2h image. Since CI here is label-driven (`build-<machine>`), nothing verifies this automatically — please add the `build-rzv2n-sr-som` label so both boards build before merge.

The refactor itself is right, and the NOTE block explaining *why* SoC-specific layers stay out of the vendor file is a good addition — it's just untested on the board it changes.

meta-avocado-renesas/recipes-kernel/linux/linux-renesas_%.bbappend
```diff
@@ -2,10 +2,42 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
...
+# RZ/V2H Robot RDK device tree (first cut; see TODO above). The .dts ships as a
+# plain readable file; the patch only registers its Makefile dtb entry. Copy the
+# dts into the kernel tree before configure so the dtb target resolves.
+SRC_URI:append:rzv2h-rdk = " file://r9a09g057h44-rzv2h-rdk.dts \
+        file://0001-arm64-dts-renesas-add-rzv2h-rdk-to-makefile.patch"
+
+do_configure:prepend:rzv2h-rdk() {
+    install -m 0644 ${WORKDIR}/r9a09g057h44-rzv2h-rdk.dts \
+        ${S}/arch/arm64/boot/dts/renesas/r9a09g057h44-rzv2h-rdk.dts
+}
```
nicksinas, line 39: **13 · Medium — new, undocumented dts-injection pattern that breaks confusingly once the vendor ships this dts.**

This is the only in-tree `.dts` in the repo — every other board's device tree comes from a vendor BSP layer — and the split mechanism (loose `.dts` copied into `${S}` at `do_configure`, plus a patch for the Makefile line) has no precedent here and no entry in `docs/adding-a-machine-target.md` §10. The next board port has nothing to copy.

Second-order failure: the `install` is unguarded. When the vendor kernel eventually ships its own `arch/arm64/boot/dts/renesas/r9a09g057h44-rzv2h-rdk.dts`, this silently overwrites it with the (by then stale) Avocado copy — hiding DT divergence — while `0001-arm64-dts-renesas-add-rzv2h-rdk-to-makefile.patch` fails to apply because the entry already exists. That surfaces as a `do_patch` error pointing at the Makefile, nowhere near the real cause.

Options: carry both the dts and the Makefile line in one patch (standard, fails loudly and in the right place), or keep the copy but add an explicit "vendor now ships this, drop the bbappend" guard. Either way, please document the choice in §10 so the pattern is deliberate.

Related, in the block above (lines 13-28): that 16-line `TODO(gap-1/2)` is stale planning text — it claims "THREE [includes] are SST-downstream-only and missing" and lays out "Path 1 / Path 2 (fallback)", all superseded by the 5 lines directly beneath it that ship a working dts with 3 includes, all present. It will read as current constraints. Same class elsewhere: `kas/machine/rzv2h-rdk.yml:17-24` commented-out repo stanza, `avocado-sdk-target.bbappend:4` ("the rzv2n provisioning scripts"), `avocado-stone.bbappend:14` ("used by both the sd and emmc profiles" — the RDK has no emmc profile).

meta-avocado-renesas/stone/stone-rzv2h-rdk.json
```diff
@@ -0,0 +1,98 @@
...
+  },
+  "storage_devices": {
+    "rootdisk": {
+      "out": "avocado-rzv2h-rdk-rootdisk.tar.gz",
+      "devpath": "/dev/mmcblk0",
+      "block_size": 512,
+      "images": {
+        "bl2_spi": "bl2_bp_spi-avocado-rzv2h-rdk.bin",
+        "fip": "fip-avocado-rzv2h-rdk.bin",
+        "flash_writer": "Flash_Writer_SCIF_RZV2H_DEV_INTERNAL_MEMORY.mot",
```
nicksinas, line 67: **14 · Medium — `flash_writer` artifact name is unverified and fails only on hardware.**

`Flash_Writer_SCIF_RZV2H_DEV_INTERNAL_MEMORY.mot` has to match exactly what the BSP flash-writer recipe deploys for `MACHINE=avocado-rzv2h-rdk`. That recipe lives in `meta-renesas`/`meta-rz-bsp` (external, pinned at `b396a94`), and the name is not among the PR's validated items — the rzv2n counterpart is `Flash_Writer_SCIF_RZV2N_SR_SOM_8GB_LPDDR4X.mot`, i.e. these names are board-string-specific, not derived.

A mismatch doesn't fail the build; it fails at `avocado provision` on the serial profile, i.e. on the bench with a board in xSPI download mode. Cheap to confirm now: `ls tmp/deploy/images/avocado-rzv2h-rdk/*.mot`.

meta-avocado-renesas/recipes-kernel/linux/files/r9a09g057h44-rzv2h-rdk.dts
```diff
@@ -0,0 +1,142 @@
...
+	vccq_sdhi0: regulator-vccq-sdhi0 {
+		compatible = "regulator-gpio";
+		regulator-name = "SDHI0 VccQ";
+		regulator-min-microvolt = <1800000>;
+		regulator-max-microvolt = <3300000>;
+		/* SD0IOVS = PA0 (vendor RZV2H_GPIO(A, 0)). */
+		gpios = <&pinctrl RZG2L_GPIO(10, 0) GPIO_ACTIVE_HIGH>;
+		gpios-states = <0>;
+		states = <3300000 0>, <1800000 1>;
+	};
```
nicksinas, line 75: **15 · Medium — SD rail sequencing/polarity diverges from the EVK reference and is unmeasured; also contradicts the commit message.**

Compared with mainline v6.12 `r9a09g057h44-rzv2h-evk.dts`, which drives SD power via a `gpio-hog` (`sd1-pwr-en-hog`) plus an always-on fixed 3.3 V rail:

- `vmmc_sdhi0` is a gpio-enabled `regulator-fixed` with no `startup-delay-us`, so the card can be probed before the rail settles.
- `vccq_sdhi0`'s `states = <3300000 0>, <1800000 1>` polarity is derived from a manual table, not measured. With `sd-uhs-sdr50`/`sdr104` enabled (line 133-134), inverted `SD0IOVS` drives the card at 3.3 V while the host signals 1.8 V — out of spec, and it presents as flaky/failed init rather than an obvious error.

Consider dropping the UHS modes for first boot and adding them once the rail is scoped — high-speed modes are worth nothing until the card enumerates at all.

Separately, a provenance issue worth fixing: the file header (lines 14-16) says the mapping is "Confirmed against the SoC manual R01UH1032EJ Table 1.2-3 ... Only actual card detection remains to verify", while commit `b924bc1`'s message says "The SD0 card-detect pinmux and vmmc/vccq GPIO ports are placeholders pending translation ... each is marked TODO". The git log and the file now disagree about what has been verified, and there are no TODO markers on those lines. Whichever is accurate, please make them match — the next person debugging SD on the bench will read one and not the other.

meta-avocado-renesas/recipes-kernel/linux/linux-renesas_%.bbappend
jetm: Confirmed on both counts. This is the only tracked `.dts` in the repo, and
`docs/adding-a-machine-target.md` never mentions one - the string does not appear
in the file.

The unguarded `install` is fixed. `do_configure:prepend` now `bbfatal`s if the
kernel already ships
`arch/arm64/boot/dts/renesas/r9a09g057h44-rzv2h-rdk.dts`, naming the fix. That
converts the silent overwrite into a loud failure at the right task, which was
the property the single-patch shape had over this one.

Keeping the loose dts rather than folding it in. While the board is moving, a
reviewer can diff the file against the vendor source; inside a patch it is
context. Once it stops moving, folding it in is the better shape and the guard
is what makes the handover visible. Documented both shapes in section 10 with
the tradeoff and the guard, so the next port has something to copy.

The stale `TODO(gap-1/2)` block is gone. You are right that it read as current
constraints - it claimed three SST-only includes were missing and laid out two
candidate paths, all settled by the five lines under it, and the shipped dts has
three includes that all resolve. Replaced with a short note on what was actually
done.

`avocado-sdk-target.bbappend:4` and `avocado-stone.bbappend:14` are both fixed:
the first said "the rzv2n provisioning scripts" for a comment now serving both
boards, the second implied an emmc profile the RDK does not declare.

Half a disagreement on `kas/machine/rzv2h-rdk.yml:17-24`. That one is
`TODO(gap-4)` and points forward - the DRP-AI3 layer is not wired, and the
commented stanza is the template for wiring it, mirroring what
`rzv2n-sr-som.yml` does. Different class from the block above, which described a
problem that no longer existed. Leaving it unless you would rather it were prose
without the stanza.

One observation while in the dts, not from your comment: it includes
`rzg2l-pinctrl.h` directly and uses `RZG2L_GPIO(10, n)`, where mainline for this
SoC includes `renesas,r9a09g057-pinctrl.h` and writes `RZV2H_GPIO(A, n)`. The
two are identical by construction - that header defines `RZV2H_PA` as 10 and
`RZV2H_GPIO(port, pin)` as `RZG2L_GPIO(RZV2H_P##port, pin)` - so this is
readability against the manual, not correctness. Worth a follow-up.

meta-avocado-renesas/stone/stone-rzv2h-rdk.json
jetm: Fair, and the naming is board-string-specific rather than derived - the rzv2n
counterpart is `Flash_Writer_SCIF_RZV2N_SR_SOM_8GB_LPDDR4X.mot`, which is not a
pattern anything generates.

I cannot confirm it from here. The flash-writer recipe is in `meta-rz-bsp`,
pinned at `b396a94`, and the deployed filename comes out of a build rather than
out of the recipe text. `ls tmp/deploy/images/avocado-rzv2h-rdk/*.mot` after
`kas/machine/rzv2h-rdk.yml` is the check, and it is on the list for the same
build that has to prove the rest of this port.

Leaving the value as-is rather than guessing at a substitute. Your failure
analysis is the reason it matters: a mismatch does not fail the build, it fails
at `avocado provision` with a board already in xSPI download mode, which is the
expensive place to find out.

meta-avocado-renesas/conf/machine/avocado-rzv2h-rdk.conf
jetm: Good catch, and the mechanism checks out: nothing in an avocado build reads
`MACHINE_EXTRA_RRECOMMENDS` unless a layer expands it, so those ten names
reached neither the rootfs nor the feed.

Fixed a bit differently than suggested, though. Rather than move the list into
the recipe, the packagegroup now expands the variable, which is what
`meta-avocado-qcom`, `-raspberrypi` and `-rockchip` already do for exactly this
problem. That keeps the machine conf as the one place a board's package set is
written, and the next Renesas board needs no new recipe text. `kernel-modules`
came out of the conf list since the packagegroup already RDEPENDs on it
directly.

Feed-only is deliberate. `packagegroup-avocado-rzv2h-rdk-extra` is wired through
`PKG_EXTRA_INSTALL` in `kas/machine/rzv2h-rdk.yml:28`, so these land in the feed
and the BSP extension picks what to install, matching how the other layers stage
board extras.

`avocado-rzv2n-sr-som.conf:76` has the same dead block. Leaving it alone here so
this PR does not touch the shipping board, but it wants the same one-line fix.

Not build-verified: `kas/machine/rzv2h-rdk.yml` is what would prove these ten
resolve and the packagegroup parses.

kas/target/bringup.yml
jetm: `kas/target/bringup.yml` is no longer part of this PR. It landed separately in
`01db8bd6` and is on `scarthgap` now, which is why this thread reads as
outdated.

Your knob is what it uses. The version on `scarthgap` sets
`AVOCADO_DEV_ROOT_LOGIN = "1"`, and it also keeps the four expanded features -
with a comment explaining why both are there, which is worth reading since it
answers your point 2 from an angle neither of us had: `debug-tweaks` itself is
defined in `core-image.bbclass`, so naming it makes bitbake skip
`avocado-image-rootfs` entirely and report "Nothing PROVIDES
'avocado-image-rootfs'". And `empty-root-password` only preserves an
already-empty password, never creates one from the `root:*:` that avocado-users
ships, so without your knob the image has no login at all.

Your points 1 and 3 are still open there: `EXTRA_IMAGE_FEATURES` is global, and
nothing marks a bringup artifact. Both belong against `scarthgap` rather than
here.

meta-avocado-renesas/recipes-bsp/u-boot/u-boot/0001-rzv2h-dev-boot-avocado-extlinux.patch
jetm: The A/B point lands. `sysboot mmc 1:1` pins partition 1, `boot-b` at
`stone-rzv2h-rdk.json:91` can never be reached, and the manifest's
`update_strategy: gpt-ab` says otherwise. That is now written into the patch
header rather than left implicit, so nobody reads the manifest and assumes
rollback works.

Not switching the mechanism in this pass, and I would rather decide it with you
than pick. Undoing the vendor `#define` so `env-nowhere.cfg:10` applies is the
cleaner shape and I agree with the reasoning, but it is three coupled changes on
a board nobody has booted. The fragment's `BOOTCOMMAND` ends in `fastboot usb 0`
and fastboot no longer builds on this machine, so the moment `distro_bootcmd`
falls through, boot ends in `Unknown command 'fastboot'` - your finding on the
u-boot bbappend, same defect from the other side. That means scoping the
fastboot token off this machine first. Then `distro_bootcmd` and `BOOTSTD_FULL`
have to actually be reachable on the `rzv2h-evk-ver1` defconfig, which I cannot
establish from this tree, and the defconfig is 2288 bytes from its
`BOARD_SIZE_LIMIT` so adding the boot-std machinery may not fit.

So it is a defconfig question, a fragment-scoping question and a size question,
none answerable without a build, and boot-b is empty at provision time anyway
until the updater writes it. Worth doing as its own change with the rzv2n build
green beside it.

The board-divergence bullet holds, and I checked it: `0001-qemu-arm-mmc-init.patch`
is the only other u-boot source patch in the tree and it touches mmc init, not
boot policy. This is the one board whose boot policy lives in a board header
rather than a Kconfig fragment, so it does not inherit changes to the shared
one. That is an argument for doing the rework, not for leaving it.

The load addresses moved for your next comment; see there.

kas/target/bringup.yml
jetm: Same as the sibling thread: `kas/target/bringup.yml` left this PR and landed in
`01db8bd6`, so it is on `scarthgap` rather than here.

The header claim is still wrong there. It says "minimal, fast, bootable image"
and "so a fresh board boots and is debuggable", and the dependency walk you did
holds - swapping `avocado-distro` for `avocado-image-rootfs` drops u-boot,
TF-A/FIP, `extlinux.conf`, the initramfs, the var btrfs and the stone bundle,
leaving `firmware-pack` and `flash-writer` via `EXTRA_IMAGEDEPENDS`. So the
overlay produces a rootfs with no way to provision it.

Worth filing against `scarthgap`. I would take the header correction over adding
`avocado-stone` as a dep, since the overlay's whole point is skipping the slow
half and pulling stone back in gives most of that time back.

meta-avocado-renesas/recipes-bsp/u-boot/u-boot/0001-rzv2h-dev-boot-avocado-extlinux.patch
jetm: Right, and the arithmetic is worse than it reads: 31.5 MiB against an `Image`
that runs 30 to 40 MiB is not tight, it is a coin flip. The vendor-parity
argument was the actual mistake. That flow loaded no ramdisk, so its spacing was
never tested against a layout that always has one, and inheriting a number is
not inheriting the guarantee.

`ramdisk_addr_r` moves to `0x4c000000` and `scriptaddr` to `0x57000000`, which
partitions the bootm window so no two loads can meet:

    fdt_addr_r     0x48000000 -> 0x48080000    512 KiB
    kernel_addr_r  0x48080000 -> 0x4c000000    63.5 MiB
    ramdisk_addr_r 0x4c000000 -> 0x57000000    176 MiB
    scriptaddr     0x57000000 -> 0x58000000    16 MiB

The table is in the patch header now, so whoever moves one address next sees the
other three.

I have not measured an `Image` for this board, so the 63.5 MiB is headroom
chosen against your 30 to 40 MiB figure rather than against a real artifact.
`kas/machine/rzv2h-rdk.yml` is what would give a number to check, and a boot is
what would prove the whole layout sits inside the window `rzv2h-dev.h` actually
configures - the 0x48000000 to 0x58000000 bound is the patch header's own claim,
not something I can confirm from here.

kas/target/bringup.yml
jetm: Also outdated for the same reason - the overlay landed in `01db8bd6` and lives on
`scarthgap` now.

The comment was rewritten there and it no longer reads as a guard. It says
"prefer the version-qualified packagegroup" for jetson and raspberrypi and then
"name the modules a board actually needs when that applies", so the claim and
the code agree: the overlay appends the unqualified set and tells you when not
to use it. It picked up a second case worth knowing about, from
`avocado-imx93-frdm`: `kernel-module-hci-uart` takes the console about 84
seconds into boot, right after `Bluetooth: hci0: Opcode 0x0c03 failed: -110`,
and the board keeps running with no console until a power cycle.

Your underlying point about the non-determinism stands and
`packagegroup-avocado-rootfs.bb:40-48` still spells it out. Nothing to change
here.

meta-avocado-renesas/recipes-kernel/linux/files/r9a09g057h44-rzv2h-rdk.dts
jetm: Confirmed against mainline, and the nodes are byte-identical: v7.1
`r9a09g057h44-rzv2h-evk.dts:61-70` carries the same `0x48000000 + 0x1f8000000`
and `0x240000000 + 0x200000000`, secure-area comment included. So this is the
EVK's memory map with an RDK model string on it, and you are right that it was
the one gap nothing recorded.

Recorded in three places now: a `TODO(confirm against the RDK BOM)` in the dts
header, a pointer on the memory nodes themselves so a reader at line 45 does not
have to scroll up, and a note on the `TFA_BOARD` TODO in the machine conf. That
last one also fixes its own bar - it said "confirm against a build", which is
the wrong test. A build proves `evk_1` compiles; it says nothing about whether
the EVK's DDR init matches the RDK's part.

Not fixing the values, because there is nothing here to fix them from. No RDK is
on the bench, and the measurement is a documentation read, not a build: bank
count, size per bank, rank layout and the LPDDR part number off the RDK BOM or
the hardware manual, then the matching `rz-atf` board definition. If the answer
is that the RDK differs, `TFA_BOARD` and both nodes change together and the
board needs a vendored board def, which is a different change from this one.

Flagging it as the blocking item for first boot rather than a nice-to-have. If
BL2 runs the wrong DDR init the console goes silent past TF-A and there is
nothing to debug with.

meta-avocado-renesas/conf/machine/avocado-rzv2h-rdk.conf
jetm: Pushing back on this one, on both halves.

`MACHINE_FEATURES` is defined as "the list of hardware features the `MACHINE` is
capable of supporting" (Yocto reference manual). It describes the board, not the
current state of the device tree. The RZ/V2H carries a Mali-G31 and the RDK has
host-only USB3.2 ports, both true whether or not a node exists yet, and the
vendor's own `meta-rz-bsp/conf/machine/include/rzv2h-family.inc:4` declares the
same GPU triple for every v2h machine. Trimming to `serial` would make the
variable describe the porting progress instead of the hardware, and the porting
progress already has a home in the dts TODO block.

The stated consequence does not reach this configuration either. No layer in
this machine's `BBLAYERS` keys on `opengles`, `opencl` or `mali-g31`: grep across
`meta-rz-bsp` and `meta-rz-distro` at the pinned `b396a94` turns up only the four
`MACHINE_FEATURES` assignments in the vendor's own family includes, there is no
mali recipe and no `virtual/libgles2` provider, and oe-core has no consumer of
`MACHINE_FEATURES opengles` at all. The mali userspace lives in a graphics layer
that is not wired here. GLES comes from mesa, gated on `DISTRO_FEATURES opengl`
which `kas/vendor/renesas.yml:32` sets and which this comment does not touch. So
nothing is pulled, the image does not grow, and there is no runtime GPU stack to
fail - which also means dropping the features would change nothing about the
built image, only about what the conf claims.

Where the comment was wrong is the line above the assignment. It said the
assignment drops the SolidRun `usbgadget`/`wifi`/`bluetooth` features, which it
cannot: this machine requires `rzv2h-family.inc` and never the rzv2n conf, so
there was nothing in scope to drop. That is rewritten to say what the line does
and to note that a feature without a ported node is a gap tracked in the dts.

`rtc` is the one I am least sure of, since whether the RDK populates an RTC is a
BOM question rather than a DT one. Happy to drop it specifically if you know the
board does not have one.

meta-avocado-renesas/recipes-bsp/u-boot/u-boot_%.bbappend
jetm: The coupling is real and the comment claiming it is inert is doing load-bearing
work it cannot do. `env-nowhere.cfg:10` still ends in `fastboot usb 0`, that
command no longer exists on this machine, and the only thing keeping it
unreached is a `#define` in a vendor header. Reworking the bootcmd per your
other comment - which I want to do - is exactly what breaks it.

Two things fixed now, one deferred.

Fixed: the provisioning script told the operator U-Boot would drop to fastboot
after Phase 1 and to run `avocado provision emmc`. Both false on this board.
That text was the coupling actually reaching a human, and it now says the boot
command finds no `extlinux.conf` and drops to the U-Boot prompt.

Fixed: the size margin is recorded next to the `UBOOT_CONFIG` TODO, with the
`0x1008f0` against `0x100000` measurement and the fact that the overflow was
cleared by deleting a feature rather than by finding room. You are right that it
was invisible from the recipe, and it is the reason the boot rework is not free -
`BOOTSTD_FULL` plus `CMD_PXE` may not fit in what is left.

Deferred: the `fastboot-fallback.cfg` split. Agreed on the shape, and it is the
first of the three steps the bootcmd rework needs, so I would rather land it in
that change than on its own here - splitting the shared fragment touches rzv2n's
boot policy, and nothing in this PR builds rzv2n. Not verifiable without a
build either way: whether the token is reachable, and whether the boot-std
machinery fits, both need `kas/machine/rzv2h-rdk.yml` and then a board.

meta-avocado-renesas/stone/rzv2h-rdk/build-disk-image.sh
jetm: The md5s reproduce and the headers are worse than misleading. It is not only
comments: the text an operator reads at the bench names the SoM, gives the SoM's
DIP S5 positions for both Serial Downloader and SPI boot, offers
`avocado provision emmc` on a board with no eMMC, and says U-Boot will drop to
fastboot after Phase 1 and that `distro_bootcmd` will probe mmc0. On this
machine fastboot is not built at all and the boot command is the patched
`sysboot`. Each of those puts someone in front of a board waiting for something
that cannot happen.

All three files now describe the RDK. The switch positions are the one place I
did not substitute a value: that is a different part on a different board, and
guessing costs a bricked bootloader region, so the text names the requirement
and points at R16UH0052EU0100 instead of restating the SoM's numbers. If you
have the RDK positions to hand I will put them in.

On the dedup: agreed on the direction, and the pattern you cite works - though
by a different route than stated. `stone.bbclass` puts `<layer>/stone` and
`<layer>/stone/${MACHINE_SHORT_NAME}` on `FILESEXTRAPATHS`; there is no
override-keyed entry. `meta-avocado-nxp/stone/avocado-imx/` resolves because
bitbake expands every `FILESPATH` entry by `FILESOVERRIDES`, so line 1 plus
`MACHINEOVERRIDES =. "avocado-imx:"` is what gets you there. Same end result,
and `avocado-rz` would work the same way.

Not doing it in this PR. Collapsing the two sets means a new shared `.inc`
setting `MACHINEOVERRIDES` on both machines, moving six files, and it changes
the SPI-NOR and SD-write paths on rzv2n, which is the shipping board and which
nothing in this PR currently builds. That is worth its own change with
`build-rzv2n-sr-som` green on it - see the vendor kas thread, where the same
missing coverage comes up. One consequence of fixing the RDK headers now is that
`build-disk-image.sh` is no longer byte-identical to its rzv2n twin; the shared
copy would get a board-agnostic header anyway, so this does not cost the
refactor anything.

Noted on the SDK hook copies, and agreed - docs section 5 is explicit that those
are boilerplate to copy.

kas/vendor/renesas.yml
jetm: All five check out. I walked the diff against the merge base and those are
exactly the files on the rzv2n path: the DRP-AI layer moves out of the vendor
kas into `kas/machine/rzv2n-sr-som.yml`, which does change its `BBLAYERS`
position; the two GPU dts patches go from unconditional to `:rzv2n-sr-som`; and
both bbappends changed - more than when you wrote this, since the
`EXTLINUX_PROVIDER` table is now a derived `extlinux-${MACHINE_SHORT_NAME}`,
which touches the rzv2n dependency too.

"Unaffected" was an assertion and it should not have been in the notes without a
build behind it. `pr-build-labeled.yml` is label-driven, so please add
`build-rzv2n-sr-som` - I cannot set labels from here. Both boards green before
this leaves draft.

That coverage also gates three other things I deferred on this review rather
than doing blind: the stone-script dedup behind a shared `avocado-rz` override,
merging the two extlinux recipes, and splitting the fastboot token out of the
shared u-boot fragment. Each is right and each edits the shipping board, so they
want the rzv2n label green first.

meta-avocado-renesas/recipes-avocado/distro/avocado-stone.bbappend
jetm: Ack, and `avocado.inc:20` is exactly it. Both bbappends now say

    do_compile[depends] += "extlinux-${MACHINE_SHORT_NAME}:do_deploy"

so the table and the per-board edit in two files are gone.

Small nit on the framing: as written it failed loud rather than open. An empty
`EXTLINUX_PROVIDER` expands to `":do_deploy"` and bitbake rejects the malformed
dependency, so a missed board did stop the build - it just stopped it with a
message that says nothing about a table in two bbappends. Deriving the name
makes the same mistake name the recipe that is actually missing, which is the
part that matters.

Not merging the two recipes here. `extlinux-rzv2h-rdk_1.0.bb` and
`extlinux-rzv2n-sr-som_1.0.bb` are identical but for `SUMMARY` and
`COMPATIBLE_MACHINE`, so an `extlinux-avocado` with the conf under a
machine-override files dir would drop both plus this variable, and I think that
is the right end state. It also renames the recipe rzv2n currently depends on,
on a board this PR does not build. Same bucket as the stone-script dedup: one
change, both boards' labels on it.

meta-avocado-renesas/recipes-kernel/linux/files/r9a09g057h44-rzv2h-rdk.dts
jetm: Took the UHS suggestion - `sd-uhs-sdr50` and `sd-uhs-sdr104` are gone, with a
line saying they come back once the rail is scoped. High-speed modes buy nothing
until the card enumerates, and the failure they can cause presents as flaky init
rather than as anything pointing at a rail.

The provenance contradiction is real and the file was the accurate side. Fixed
by saying which part is established and which is not, rather than by picking one
of the two. What is confirmed is the port and pin numbering, and it turns out
that is checkable rather than a manual read:
`include/dt-bindings/pinctrl/renesas,r9a09g057-pinctrl.h` defines `RZV2H_PA` as
10 and `RZV2H_GPIO(port, pin)` as `RZG2L_GPIO(RZV2H_P##port, pin)`, so the
`PA -> 10` translation the header claims is exact. What is not confirmed is the
pin assignment itself - that PA5 is card detect on this board, that PA1 drives
SD0PWEN, that PA0 is SD0IOVS - plus the vmmc settle time. The header now names
those three rather than only card detection.

One refinement on the polarity argument. The `states = <3300000 0>,
<1800000 1>` table is not hand-derived: mainline `r9a09g057h44-rzv2h-evk.dts`
carries the identical table with `gpios-states = <0>` for its own `vqmmc_sdhi1`.
So the risk is not that the table is wrong for this signal, it is that PA0 is
not SD0IOVS on the RDK - which lands in the same place, and is why the UHS
removal is right regardless.

The vmmc divergence is the sharper half and you called it correctly. The EVK
drives SD power from an always-on `reg_3p3v` plus an `sd1-pwr-en-hog` gpio-hog
at `RZV2H_GPIO(A, 3)`; this dts uses a gpio-enabled `regulator-fixed` with no
`startup-delay-us`, so the card can be probed before the rail settles. I have
not added a delay because any number I put there would be invented. Scoping
SD0PWEN gives the real one, and that is on the same bench visit as card
detection.

The dts compiles clean, no dtc warnings, against mainline v7.1 with these
changes. That proves syntax and label resolution only - it is a newer tree than
the `rz-6.12-cip7` the recipe builds, so it is not a substitute for
`kas/machine/rzv2h-rdk.yml`.

PR #227, opened 2026-06-16
recipes-ml: add ExecuTorch 1.3.1 (+ torchgen-native)

jetm:
**Depends on #223** (feed-validation infra + qemuarm64 default machine). Draft until #223 lands; I will rebase onto `scarthgap` and retarget then.

## What

Adds a scarthgap recipe for ExecuTorch 1.3.1 (PyTorch's portable inference runtime), building the minimal portable CPU runtime (static libs + headers). Wires `executorch-staticdev` into `packagegroup-avocado-extra` and adds an SDK-tier feed-validation case.

New recipes:
- `recipes-ml/executorch/executorch_1.3.1.bb` - portable C++ runtime; all backends (XNNPACK, CUDA, Vulkan, QNN, OpenVINO, Ethos-U), Python bindings, tests, and devtools off.
- `recipes-devtools/torchgen/torchgen_2.12.0.bb` - native-only, pure-Python torchgen for the build-time operator codegen (no libtorch build).

## Build notes (the non-obvious bits)

- **gitsm fetch, `branch=release/1.3`** - the v1.3.1 tag is the tip of the release branch, not main; the bundled submodules need the gitsm fetcher (`git://...;submodules=1` silently fetches nothing).
- **`destsuffix=executorch`** - upstream CMakeLists aborts unless the source tree is named exactly `executorch` (issue 6475); destsuffix also serves as the PYTHONPATH import root for codegen.
- **native flatc** (`0001-use-native-flatc-for-cross-compilation.patch`) - executorch builds flatc and runs it at build time; the patch repoints the imported `flatc` target at `flatbuffers-native` so it does not cross-build.
- **torchgen-native pinned to 2.12.0** - executorch 1.3.1 requires `torch>=2.12.0a0`; the codegen's ATen schema must match (the packaged 2.4.1 lacks ops like `mean.dtype_out`). Pure-Python torchgen extracted from the CPU wheel - no nightly libtorch build.
- Packaging: relocate two statically-installed libs from the build dir to `${libdir}`, route cpuinfo's cmake/header to `-dev`, and rewrite the exported cmake targets to be install-prefix-relative (no `buildpaths` QA warning, working `find_package`).

## Validation

`bakar bitbake executorch meta-avocado/kas/machine/qemuarm64.yml` builds **warning-free**. Feed-validation SDK tier passes on qemuarm64: `executorch-staticdev` installs into the app extension and `libexecutorch.a` lands in the ext sysroot. executorch ships only static libs, so the case is SDK-tier only (`boot=no`).

### DIFF
--- a/meta-avocado-distro/recipes-devtools/torchgen/torchgen_2.12.0.bb
+++ b/meta-avocado-distro/recipes-devtools/torchgen/torchgen_2.12.0.bb
@@ -32,5 +32,12 @@ do_install() {
     cp -r ${S}/torchgen ${D}${PYTHON_SITEPACKAGES_DIR}/
 }
 
-# Build-time codegen tool for ExecuTorch; only the native variant is consumed.
-BBCLASSEXTEND = "native"
+# Claim the installed site-packages so the package ships them. The native
+# variant skips the installed-vs-shipped QA, but the nativesdk variant produces
+# a real SDK package and fails do_package without an explicit FILES entry.
+FILES:${PN} += "${PYTHON_SITEPACKAGES_DIR}"
+
+# Build-time codegen tool for ExecuTorch. The native variant is consumed during
+# the image build; the nativesdk variant ships in the Avocado SDK so a developer
+# can run the executorch codegen when cross-compiling against the runtime.
+BBCLASSEXTEND = "native nativesdk"
--- a/meta-avocado-distro/recipes-devtools/torchgen/torchgen_2.12.0.bb
+++ b/meta-avocado-distro/recipes-devtools/torchgen/torchgen_2.12.0.bb
@@ -1,0 +1,36 @@
+SUMMARY = "PyTorch operator code generator (torchgen)"
+DESCRIPTION = "The pure-Python torchgen package extracted from the PyTorch \
+2.12.0 CPU wheel: the ATen schema (native_functions.yaml, tags.yaml) and the \
+codegen modules that ExecuTorch's build-time operator codegen (codegen.gen and \
+codegen.tools.gen_oplist) consumes. Native-only build tool - no libtorch is \
+built or installed. Pinned to 2.12.0 because ExecuTorch 1.3.1 requires \
+torch>=2.12.0a0 and its functions.yaml references ATen overloads (e.g. \
+mean.dtype_out) absent from older schemas."
+HOMEPAGE = "https://github.com/pytorch/pytorch"
+SECTION = "devel"
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=b114fbe63fdb5f7a91332a4aefb61ee5"
+
+PV = "2.12.0"
+
+# The wheel is a zip; downloadfilename=*.zip makes BitBake unpack it so
+# ${WORKDIR}/torchgen (with its packaged ATen yamls) is available. Only the
+# pure-Python torchgen package is installed; the bundled libtorch is ignored.
+SRC_URI = "\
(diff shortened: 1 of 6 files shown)


mobileoverlord:
Looking good, can you rebase this please

jetm:
Hi Justin, done - rebased onto scarthgap. The feed-validation, zeromq, and qemu guest-agent commits had already merged there, so they dropped out as duplicates; what's left is the four executorch-specific commits (torchgen, the recipe, the SDK-tier feed-validation case, and the SDK staging). Verified locally with an avocado-complete build and the feed-validation E2E, both green (3/3). Ready for another pass.

mobileoverlord:
Merged to `scarthgap` (rebased + reviewed), so closing this out. GitHub didn't auto-detect the merge because the commits were rebased onto current `scarthgap` (after #223 landed) and carry review fixes, so the SHAs differ from this branch's head.

Landed as `83821ed..fa29ec9` on `scarthgap`:

- `83821ed` torchgen: add native-only recipe pinned to torch 2.12.0
- `636287e` executorch: add v1.3.1 recipe and wire into packagegroup-extra
- `145946e` feed-validation: add executorch SDK-tier case
- `71f770a` recipes-avocado/sdk: stage executorch into the Avocado SDK
- `191de6b` torchgen: source the wheel and LICENSE from stable, immutable refs
- `fa29ec9` executorch: build flatc/flatcc for the host; fix DESTDIR leak; drop flatbuffers-native

### Review fixes applied during the merge

- **`executorch-staticdev` placement.** The rebase resolved a conflict against the scarthgap change that moved target-arch SDK-sysroot packages out of `packagegroup-avocado-sdk-extra` (a host/nativesdk packagegroup whose feed routing only uploads `*_avocadosdk`, so target-arch packages built there never reach a feed). `executorch-staticdev` is target-arch, so it now rides `packagegroup-avocado-extra` (the target feed) alongside `avocado-sdk-target-sysroot`, **not** `SDK_SYSROOT_DEPENDS`. This reverses part of the "stage executorch into the Avocado SDK" approach; the `nativesdk-executorch`/`nativesdk-torchgen` host-tool entries are unchanged.

- **flatc/flatcc cross-build reworked (the load-bearing fix).** The original patch repointed `flatc` at the ambient `flatbuffers-native`, whose version varies with the target's layer set — e.g. the i.MX `meta-ml` stack pins flatbuffers `23.5.26` while ExecuTorch bundles `24.3.25`. That tripped the generated-schema flatbuffers version `static_assert` on i.MX (`Non-compatible flatbuffers version included`) while passing on qemu (which gets `24.3.25`). The new patch (`0001-build-bundled-flatc-flatcc-for-the-build-host.patch`) builds `flatc` **and** `flatcc` from the bundled `third-party` sources for the build host, so the generator and the headers it's compiled against are the same flatbuffers version on every target. `flatbuffers-native` is consequently dropped from `DEPENDS`.

- **torchgen sources made durable.** Wheel now pulled from the stable `whl/cpu` channel and the LICENSE from the immutable `v2.12.0` tag (were the pre-release `test` channel and the mutable `release/2.12` branch). Both artifacts are byte-identical to the previous ones, so the pinned `sha256` sums are unchanged.

- **`do_install` DESTDIR leak fix.** The two statically-installed libs leak into DESTDIR at an absolute build path; the prior `rm -rf ${D}/work` only cleaned that when `TOPDIR` was `/work`. Replaced with a TOPDIR-independent removal of `${D}${B}` and its empty parent chain (the i.MX build's `TOPDIR` is `/avocado-build`, which the old form missed).

### Validation

`nativesdk-executorch` builds clean through packaging on **imx93-frdm** with the reworked patch (the path that originally failed). Follow-up worth doing: a target (aarch64) `executorch` build to exercise the `flatcc` host-build on a real cross target, and the broader at-risk matrix (other NXP i.MX, CompuLab i.MX8MP, RubikPi/Qualcomm).

PR #228, opened 2026-06-17
fix multikernel merge bbclass

bsandu-peridio:


mobileoverlord (approved):
Ahhhhh, yeah I would like to eventually get rid of publicizing the meta packages without the fully qualified NEVRA. To support meta packages we would need to significantly tighten the publication window of the feed. for it to be unambiguous. In this instance, its just getting in the way.

PR #229, opened 2026-06-18
Redesign packagegroup-avocado-extra into opt-in feature groups

jetm:
## Problem

`packagegroup-avocado-extra` is a single monolithic packagegroup that pulls every
optional capability (Qt, Chromium/WebKit, GStreamer, Python, ROS, Java, containers,
AWS, cameras, AI) into every `avocado-complete` build. The only way to slim an image
was to hand-roll `RDEPENDS:...:remove` overlays (the opt-out approach in #222). That is
backwards: a dev image pays for the heaviest recipes by default and has to subtract
them, and the pattern only scales to the few packages someone writes a remove for.

## Solution

Invert the model to opt-in. Split the monolith into one
`packagegroup-avocado-feature-<group>.bb` per RPM category, and rewrite
`packagegroup-avocado-extra` as a composing group that expands a kas-set
`AVOCADO_FEATURE_GROUPS` variable into the matching feature packagegroups. Each group
ships a self-contained `kas/feature/<group>.yml` fragment that pulls its vendor layer
and appends its token. Stack the fragments you want; a `complete.yml` umbrella
reproduces today's full content for existing clients.

## Key changes

- 12 `packagegroup-avocado-feature-*.bb` recipes (system-base, networking, multimedia,
  python, graphics, qt, browsers, cameras, ai, cloud-aws, java, containers), migrated
  from the monolith with every `DISTRO_FEATURES` / `MACHINE_FEATURES` / `TARGET_ARCH`
  guard preserved verbatim.
- `packagegroup-avocado-extra.bb` rewritten as always-on base content plus the token
  expansion.
- One `kas/feature/<group>.yml` per group, a `complete.yml` umbrella, and layer-only
  fragments for clang, python-ai and ros.
- Vendor layers the targets force-pulled now live in their feature fragments;
  `kas/target/{distro,sdk}.yml` no longer pull `kas/extra.yml`, and the dead
  `kas/extra.yml` aggregator and all its overlays are removed (the grinn-astra deepx
  board was the last `kas/extra/` consumer; it now uses `complete.yml` + `ai.yml`).
- `meta-avocado-distro` recipes and bbappends that extend optional vendor recipes
  (qtwebengine, wpewebkit, cxxopts, openjdk-17, aws-iot-device-client, hpp-fcl) move
  under `dynamic-layers/<collection>/` gated by `BBFILES_DYNAMIC`, so a composition
  without the layer parses cleanly instead of failing.
- Base vendor layers that back always-on content (meta-fwup, meta-erlang, meta-atmel,
  meta-swupdate, meta-openjdk-temurin) are provisioned unconditionally in
  `kas/base.yml`. openjdk is base because the always-on KOS tools RDEPEND a JRE.
- The client `kas/vendor/*.yml` and `kas/machine/*.yml` files include `complete.yml`
  (deepx boards add `ai.yml` alongside it) so existing targets keep identical content.
- Includes a fix to `meta-avocado-qemu/scripts/run-qemuarm64` so the arm64 image boots
  on a non-Yocto host (needed to validate the boot path).
- 12 per-feature-group feed-validation cases.

## Usage

Minimal dev image, opt into only what you need:

    kas build kas/machine/qemuarm64.yml:kas/feature/system-base.yml:kas/feature/networking.yml:kas/feature/multimedia.yml

Full content for existing clients is unchanged (the vendor/machine files include
`complete.yml`):

    kas build kas/machine/imx8mp-evk.yml

## Validation

- Static composition: every composition (minimal, each single feature, complete)
  resolves with no dangling bbappends and no missing inherited classes, checked before
  building.
- Feed content, real kas build (qemuarm64 + `system-base:networking:multimedia`):
  `gstreamer1.0-*` present in the package feed; `chromium-ozone-wayland` and `qtbase`
  absent. Opt-in proven in both directions.
- Boot: that composition provisioned and booted under QEMU through TF-A, U-Boot and the
  kernel to a login prompt.
- Client manifest preservation: for `imx8mp-evk` and `grinn-astra-1680-sbc` (both
  include `complete.yml`) the resolved kas composition is identical to `scarthgap` (same
  layers and commits, same target, same `bblayers`/`env`), and the feature-group
  partition is exact: base plus all groups RDEPEND the same package set as the old
  monolith, zero packages added or dropped. Default client image content does not change.

## Reviewer notes

Supersedes #222 (ENG-1975). That PR added opt-out overlays to subtract chromium/Qt from
dev images; this reverses to opt-in feature groups, which scales to every capability
rather than the two heavy ones. Per @mobileoverlord's point on #222, clients keep
building the full set by default through `complete.yml`. Implements ENG-2003.


mobileoverlord:
This looks great, I am building locally. In the meantime, could you add something in the top level docs with notes for maintainers on how to build various combinations of this from bare minimal to full featured.

jetm:
> This looks great, I am building locally. In the meantime, could you add something in the top level docs with notes for maintainers on how to build various combinations of this from bare minimal to full featured.

Done, in 7358674. New page at `docs/feature-groups.md`, linked from the README build section.

It walks the composition model first - the `AVOCADO_FEATURE_GROUPS` tokens, the kas colon-overlay syntax, and the `complete.yml` umbrella - then gives copy-paste `kas build` recipes from a bare machine all the way to the full set. There's a per-group table listing each group's vendor layer and its `DISTRO_FEATURES`/`MACHINE_FEATURES`/arch gating. Two gotchas are called out: graphics/Qt/browsers resolve to empty without `opengl wayland`, and `ai` stays out of `complete.yml` because it's deepx-only, so deepx boards stack `complete.yml` + `ai.yml` (grinn-astra already does). If you want board-specific examples or a different cut, say the word.

jetm:
I'm adding you as a reviewer, @nicksinas, since Justin reviewed and requested a change that I have already made. SBOM, /var encryption, and other incoming Yocto changes require this work.

nicksinas (approved):

jetm:
@nicksinas I rebased this onto latest `scarthgap` to clear the `run-qemuarm64` conflict. The old fix carried here was redundant with #240 - that PR merged the same uninative-loader fix - so the rebase just dropped one commit and nothing else changed. The force-push dismissed your approval, so I re-requested it.

nicksinas (approved):

PR #232, opened 2026-06-25
Add TPM2 LUKS sealing, dm-verity, and Secure Boot key management

jetm:
## Problem

The iMX93 FRDM board shipped Phase-1 LUKS2 `/var` encryption with a software Argon2id KDF. The `security` and `secureboot` DISTRO_FEATUREs were wired in name only: dm-verity ran only on QEMU via a build-dir-specific fragment, module signing had no fragment, and the TPM2 stack had no runtime packagegroup, no initramfs split, and abrmd auto-enabled into the initramfs where D-Bus does not exist.

## Solution

Complete Phase-2 hardware binding via `systemd-cryptenroll` (already enabled in the systemd PACKAGECONFIG), add working dm-verity and module-signing kernel fragments gated on DISTRO_FEATUREs, scaffold UEFI SB key generation, and add feed-validation assertions for every security feature.

The TPM2 unsealing follows a two-phase boot progression: the Argon2id recovery keyslot (slot 0) is retained until `cryptsetup open --token-only` confirms the TPM2 slot unseals, then and only then is slot 0 retired.

## Key changes

- **LUKS2 + TPM2**: `cryptsetup-var.sh` enrols a PCR 7-bound TPM2 token on first boot; subsequent boots prefer `--token-only` and fall back to Argon2id; `luksKillSlot` is gated behind a verified unseal
- **dm-verity**: portable `dm-verity.cfg` kernel fragment in the shared layer (replaces the QEMU-only `.cfg.in` template), gated on the `secureboot` DISTRO_FEATURE
- **Module signing**: `modsign.cfg` fragment enabling `MODULE_SIG_FORCE`, gated on the `security` DISTRO_FEATURE
- **SB key management**: idempotent `sb-keys.bb` recipe + `gen-sbkeys.sh`, with private keys never installed to target and key paths excluded from sstate hashes
- **TPM2 packaging**: runtime packagegroup (tpm2-tss + tpm2-tools + tpm2-openssl), trimmed initramfs packagegroup (tools only, no abrmd), abrmd auto-start disabled

## Why not meta-secure-core as a layer dependency?

Wind River's `meta-secure-core` was analyzed as a source of patterns but not added as a layer. Three reasons:

1. **Conflicting TPM2 unlock path.** meta-secure-core uses `cryptfs-tpm2` with a custom `/init.cryptfs` initramfs script. Avocado's existing systemd already enables `systemd-cryptenroll` via `PACKAGECONFIG`. Adding the layer would introduce a second parallel unlock mechanism alongside the one already in place - two competing paths opening the same LUKS container is a bricking risk.

2. **IMA/EVM is bundled and incompatible with TPM PCR sealing.** meta-secure-core ships IMA/EVM appraisal. IMA's measurement rules extend PCRs during boot, which breaks TPM PCR-policy-session unsealing: the PCR values at unseal time no longer match the sealed values. IMA is deferred to a later PR after seal and verity stabilize.

3. **Foreign DISTRO_FEATURE set and bbclass structure.** meta-secure-core carries its own DISTRO_FEATURE names and the `user-key-store.bbclass` central signing engine. Integrating the full layer would require adopting its naming conventions or writing compatibility shims throughout meta-avocado. The patterns were adopted (PCR 7 sealing, two-packagegroup initramfs split, idempotent keygen); the layer was not.

For the TPM2 userspace stack (`tpm2-tss`, `tpm2-abrmd`, `tpm2-tools`) this PR consumes the upstream OE `meta-security`/`meta-tpm` layers, which is where those recipes are actively maintained.

## Reviewer notes

**Not yet verified (requires Yocto build or hardware):**
- `bitbake packagegroup-avocado-tpm2` smoke build (tpm2-tss/abrmd cross-build for ARM + x86)
- `veritysetup format` against a real erofs rootfs image
- `fvl_assert_luks_var_tpm` + `fvl_assert_dm_verity` feed-validation on qemuarm64
- PCR 7 read on a PR-1 iMX93 image with fTPM (determines whether PCR 7 carries a real measurement without active Secure Boot)

**IMA/EVM deferred**: three destructive PCR interactions with the LUKS TPM unseal path; will land after seal + verity stabilize.

**AHAB/UEFI Secure Boot not enabled here**: this PR prepares keys and verity but does not flip the boot chain (PR-2 work). The TPM2 seal is defense-in-depth until PR-2.


jetm:
Appended `e1dd405`: a self-contained fix for `meta-avocado-qemu/scripts/start-swtpm`, which has been broken since `64c30e2` removed the `find-library-path` helper it called. That helper shelled out to `ldd`, which can't execute the uninative-linked native binaries staged under `sysroots-components`, so with it gone the script aborts before launching swtpm.

The fix mirrors the working sibling `run-qemux86-64-swtpm`: it invokes swtpm through the uninative glibc loader with `--library-path` spanning the native sysroot lib dirs, and drops the hand-rolled `libpcre.so.3` symlink the sibling never needed (the loader resolves the SONAME once `libpcre-native`'s lib dir is on the path).

Verified `bash -n` and `shellcheck` clean, and the `run_native` mechanism itself is confirmed working in-tree - qemuarm64 boots to a login prompt through the identical loader pattern. It is not runtime-verified against swtpm directly, since a no-TPM qemuarm64 build doesn't stage `swtpm-native`; exercising that end to end needs a TPM-enabled build.

jetm:
Unified with #236 into #272. As a two-PR stack neither PR showed the whole change, and this branch had drifted from the one built on top of it. All commits from both are in #272, rebased onto current scarthgap, plus the fixes that until now existed only on the wrynose line.

The two commits unique to this branch are accounted for: the initramfs `encrypted-var` gate was already present in the other branch under an equivalent commit, and the var seed restoration was missing from every other branch and is now cherry-picked onto both #271 and #272.

Closing as superseded by #272.

PR #235, opened 2026-06-29
[DRAFT] bitbake-setup: generate Yocto-native build configs from kas

yoctopidg3:
Adds bitbake-setup as a parallel way to set up AvocadoOS alongside kas. We translate the existing kas configuration into upstream Yocto bitbake-setup configuration templates, so the same build can be driven by either  tool and the two never drift. Also includes a build-tree/cache layout for the wrapper and one kernel-config fix. 

kas remains the canonical setup mechanism, but for grumps like me, we can still use bitbake-setup.

Generator (scripts/kas2bitbake-setup.py)
    - Runs kas dump offline per kas/machine/*.yml and maps the flattened result into bitbake-setup templates.
    - meta-avocado itself (the url-less local repo in kas) is emitted as bb-layers-file-relative pointing at the live working tree 

Generated configs (27 machines)
    - One avocado-<machine>.conf.json per machine plus the matching meta-avocado/conf/fragments/avocado-build/<machine>.conf.
    - Vendor sources pinned to the same wrynose/wrynose-avocado revisions as kas.
    - Covers the full machine set.

Wrapper (scripts/bitbake-setup)
    - Pins the build top dir under meta-avocado/bitbake-builds/ (mirrors kas's keep-build-trees-in-checkout convention).
    - Caches DL_DIR/SSTATE_DIR in a shared external cache (default ~/.cache/avocado-bitbake-setup), overridable via AVOCADO_BB_CACHE / AVOCADO_DL_DIR / AVOCADO_SSTATE_DIR, so a rm -rf bitbake-builds no longer nukes downloads/sstate.
    - .gitignores the bootstrap bitbake clone and the build trees.

Docs
    - Both setup paths documented in the top-level README, plus bitbake-setup/README.md
Kernel fix

Kernel: (which should probably be on it's own PR, but I stashed it here)
    - # CONFIG_USB_SERIAL_CONSOLE is not set in the qemu and x86-64 kernel fragments. The common-pc BSP sets it =y, which hard-depends on USB_SERIAL=y. AvocadoOS ships USB_SERIAL=m, so it silently dropped and tripped do_kernel_configcheck.


mobileoverlord:
Is there a way to concatenate config fragments as to de duplicate any shared local config settings between target families?

PR #236, opened 2026-07-03
scarthgap: LUKS2-encrypted /var sealed to the TPM, with a boot-method abstraction

yoctopidg3:
## Target branch

Please apply this to [`avocado-linux/meta-avocado:imx93-secure-boot-and-var-encryption`](https://github.com/avocado-linux/meta-avocado/tree/imx93-secure-boot-and-var-encryption) (Javier's branch), *not* `scarthgap`. These commits sit directly on top of that base and extend it; they are not meant to land on `scarthgap` on their own.

## What this adds

This branch:

1. Brings the encrypted-`/var` + TPM2-seal flow up on the two QEMU reference machines (`qemux86-64`, `qemuarm64`) and the Intel target, end to end: **first-boot format + enroll, and reopen-on-reboot via the TPM**.
2. Introduces a boot-method abstraction (`AVOCADO_BOOTLOADER`) so boot-artifact build dependencies are selected by boot method (`uboot` / `uefi`) instead of by hardcoded machine names. (There is no var I know of in YP, that allows this, and frankly, this is simple)
3. Stands up an **OP-TEE firmware-TPM (fTPM)** on `qemuarm64`, which has no discrete TPM, so the same seal flow works on ARM.
4. Fixes two pre-existing bugs that stopped the sealed `/var` from reopening on reboot (a udev race, and a missing libcryptsetup token plugin).

## Why we're doing this

**Encrypt `/var`, not the whole disk.** The rootfs is read-only and verified as part of the image; the mutable state that actually needs confidentiality at rest lives in `/var`. Encrypting just `/var` keeps the boot/root path simple and measurable while protecting the data that changes.

**Seal the disk key to TPM2 PCR 7.** PCR 7 records the platform's Secure Boot state (Secure Boot on/off and the active keys). Sealing the `/var` key to PCR 7 means the volume only auto-unlocks when the machine boots in the *same trusted Secure Boot configuration* it was enrolled in. Disable Secure Boot, swap firmware keys, or otherwise change that policy and the PCR differs, the unseal fails, and the volume will not silently open. The key is never on disk in the clear; it lives sealed in the TPM.

[...]

PR #237, opened 2026-07-05
systemd, qemu: build native recipes hermetically against host libs

jetm:
## Problem

Two -native recipes pick up optional host libraries during do_configure and
fail a clean rebuild whose recipe sysroot does not carry the matching headers.
The result depends on what the build host has installed, so a build that
succeeds on one machine fails on another:

- systemd-systemctl-native leaves systemd's meson `pam` option at 'auto'.
  Configure probes the host for libpam and enables it, but there are no pam
  headers in the sysroot, so libsystemd-shared's pam-util.c fails on a missing
  security/pam_ext.h.
- qemu-system-native auto-detects the host libkeyutils and enables the
  cryptodev-lkcf backend, so backends/cryptodev-lkcf.c fails on a missing
  keyutils.h.

Both surface on a from-scratch rebuild on a second builder that shares the
sstate cache but does not carry the host packages the first builder had.

## Change

- systemd-systemctl-native: pin -Dpam=disabled. This recipe builds only the
  offline systemctl:executable helper (run on the build host to preset services
  in the target rootfs); it never uses PAM. libpam also cannot be a dependency:
  it gates on ANY_OF_DISTRO_FEATURES "pam systemd" and native builds use
  DISTRO_FEATURES_NATIVE (no pam), so there is no libpam-native provider. Scoped
  to the native helper only -- the target systemd keeps PAM.
- qemu-system-native: add keyutils-native to DEPENDS. Unlike libpam, keyutils
  has a native provider, so the cryptodev-lkcf backend builds from a staged
  header and stays functional.

## Testing

Built core-image-minimal for qemuarm64 from a clean tmpdir on a second builder
sharing the sstate cache over NFS. Before: both recipes fail as above when they
rebuild. After: both build from their recipe sysroot and the image completes
(exit 0), with 99% of tasks restored from shared sstate.

### DIFF
--- a/meta-avocado-distro/recipes-core/systemd/systemd-systemctl-native_%.bbappend
+++ b/meta-avocado-distro/recipes-core/systemd/systemd-systemctl-native_%.bbappend
@@ -1,0 +1,6 @@
+# Host build: the buildtools gcc is sandboxed to its own --sysroot and does not
+# search the host /usr/include. systemd's PAM detection defaults to auto and
+# finds the host libpam, but the compile of src/shared/pam-util.c then fails on
+# the missing security/pam_ext.h. systemctl-native has no use for PAM, so drop
+# it outright rather than pulling in libpam-native.
+EXTRA_OEMESON += "-Dpam=disabled"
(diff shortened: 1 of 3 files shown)


jetm:
@nicksinas assigning this to you - it's blocking my ongoing BayLibre work. Without these native-recipe hermeticity fixes the qemuarm64 native tooling fails to build (qemu-system-native on missing keyutils.h, systemd-systemctl-native on missing pam_ext.h), which stops me verifying the secure-boot / var-encryption stack (#232 + #236). Could you take a review?

nicksinas (approved):

PR #240, opened 2026-07-16
qemuarm64: repair the run script and unblock the nativesdk build

jetm:
Two self-contained fixes for building and booting qemuarm64. Neither introduces nor depends on any feature; each stands alone against `scarthgap`.

### `run-qemuarm64` native invocation
The script still sourced the deleted `scripts/find-library-path` and pointed at a nonexistent `avocado-qemuarm64-rootdisk.img`, with the BIOS under the stone `_build` dir. Commit `64c30e2` moved the run scripts to the uninative-loader pattern and updated `run-qemux86-64` but not the arm64 script, so booting qemuarm64 via the script aborted immediately. Ported the same `run_native()` uninative-loader helper and the real stone artifact names (`avocado-os-qemuarm64.img`, `flash.bin`).

### ganesha nativesdk D-Bus
The nativesdk ganesha (a host-side NFS tool for dev-VM provisioning) fails `do_compile` on a missing `dbus/dbus.h`, and does not use the D-Bus management interface. Dropped the `dbus` PACKAGECONFIG for `class-nativesdk` only; the target build is unchanged.

### Testing
Built qemuarm64 and booted the provisioned image to a login prompt via the fixed run script.

### DIFF
--- a/meta-avocado-qemu/scripts/run-qemuarm64
+++ b/meta-avocado-qemu/scripts/run-qemuarm64
@@ -1,34 +1,47 @@
 #!/usr/bin/env bash
+# Run QEMU ARM64 without TPM.
+# Usage: ./run-qemuarm64
+# Assumes pwd is the build directory (sources/build-qemuarm64).
 
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+set -euo pipefail
+
 BUILD_DIR="$(pwd)/build"
-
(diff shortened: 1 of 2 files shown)


jetm:
@nicksinas assigning this to you. Both fixes are self-contained and stand alone against scarthgap - the qemuarm64 run script (broken since 64c30e2) and the nativesdk ganesha build. They're blocking my /var encryption work: I need them merged to build and boot the encrypted-var verification images on qemuarm64. The encryption itself is a separate, opt-in kas feature (`encrypted-var`, default off) and is deliberately not in this PR - this is just the build/boot plumbing.

nicksinas (approved):

PR #241, opened 2026-07-18
kas/vendor/oe: pin bitbake to race-review-fixes for the cross-node race fixes

jetm:
Repoint the bitbake pin from 2.8-avocado to vendor-bitbake race-review-fixes, which adds the mkdirhier EEXIST/ESTALE retries and the asyncrpc client connect timeout - the cold-review fixes for the cross-node build wedge (a stalled hashserv/prserv connect no longer wedges the cooker in select()). Depends on avocado-linux/vendor-bitbake#race-review-fixes.


kas/vendor/oe.yml
```diff
@@ -15,8 +15,20 @@ repos:
...
+    #
+    # Rebased onto the current 2.8-avocado tip for this merge. The branch had
+    # been cut from an older base, so pinning it while scarthgap moved forward
+    # would have dropped eleven commits that landed on 2.8-avocado in the
+    # meantime - among them the fetch2 deb/ipk member-name and striplevel
+    # validation, the shallow-ref quoting, RPM --no-absolute-filenames, and the
+    # hashserv unihash validation. Keeping the race fixes is not worth losing
+    # those, so the branch moved rather than the pin.
+    branch: "race-review-fixes"
+    commit: 81d1e0e96a7a9a88ffc4171137b581bf97c33236
```
jetm: **The pinned commit ships a broken `mkdirhier`**

At `81d1e0e9`, `mkdirhier()`'s exhausted-EEXIST path does a bare `return`, reporting success with nothing at the path. The fix for it, `f06288b4`, is on this same branch but two commits *after* the pin.

Reproduced against real worktrees rather than inferred:

```
PIN  81d1e0e9 : outcome = RETURNED SUCCESS   elapsed 1.00s   isdir() = False
TIP  7fcf96b2 : raised FileExistsError       elapsed 0.00s
BASE fc13d719 : raised FileExistsError       elapsed 0.00s   <- what scarthgap pins today
```

A path occupied by a dangling symlink is enough - no NFS required. `mkdir(2)` returns EEXIST for the link regardless of target, `os.stat()` follows it and fails ENOENT, the retry loop burns all 5 attempts (`MKDIRHIER_ATTEMPTS = 5`, about 1s of sleeps) and returns success. Every bitbake caller relying on the `mkdir -p` postcondition - WORKDIR, stamp, sstate directory creation - then proceeds against a path with no directory, converting a loud immediate `FileExistsError` into a later unattributable ENOENT elsewhere in the task.

That makes builds strictly worse than the `fc13d719` they run today, which is why this is blocking rather than a nit. Re-pin to `7fcf96b2` or later.

Worth noting for the commit body: "81d1e0e9 confirmed as the current branch tip" was true when written (12:16) and stopped being true at 16:04/16:11 the same day. `git ls-remote` now gives `refs/heads/race-review-fixes = 7fcf96b2e00e7e41c036a4265820c2b4ef95bfa3`. Re-verify at merge rather than trusting the recorded check.

kas/vendor/oe.yml
```diff
@@ -15,8 +15,20 @@ repos:
   bitbake:
     url: https://github.com/avocado-linux/vendor-bitbake
     path: bitbake
-    branch: "2.8-avocado"
-    commit: fc13d7197745017a5c8c274cde82b62a9cc796ce
+    # race-review-fixes = 2.8-avocado + mkdirhier EEXIST/ESTALE retries and the
+    # asyncrpc client connect/handshake budgets - cold-review fixes for the
+    # cross-node build race (a stalled hashserv/prserv connect no longer wedges
+    # the client).
```
jetm, line 21: **The pin does not deliver the fix this comment claims**

The comment says "a stalled hashserv/prserv connect no longer wedges the client". At `81d1e0e9` there is no wall-clock bound, so it does not.

A hashserv/prserv that is alive and listening but blocked - NFS I/O, sqlite lock, GIL starvation - has its handshake completed by the kernel backlog, so `self.socket` is non-None and every failure bills the *request* budget: `MAX_REQUEST_RETRIES = 3` (4 attempts) times `self.timeout`, which is the same total the connect/request split was meant to eliminate. The lever that actually caps it, `deadline = time.monotonic() + self.timeout * RETRY_DEADLINE_TIMEOUTS`, exists only at `7fcf96b2` (`lib/bb/asyncrpc/client.py:226-228`).

Second concrete cost at the pin: `MAX_CONNECT_RETRIES = 1` with no backoff, so both attempts land inside the few-hundred-millisecond window a restarting hashserv spends rebinding. The pinned client gives up on a server restart *faster* than upstream's 4-attempt loop would. The tip restores 4 attempts plus `CONNECT_RETRY_BACKOFF`.

Same one-line fix as the other thread: re-pin to `7fcf96b2`.

kas/vendor/oe.yml
```diff
@@ -15,8 +15,20 @@ repos:
...
+    #
+    # Rebased onto the current 2.8-avocado tip for this merge. The branch had
+    # been cut from an older base, so pinning it while scarthgap moved forward
+    # would have dropped eleven commits that landed on 2.8-avocado in the
+    # meantime - among them the fetch2 deb/ipk member-name and striplevel
+    # validation, the shallow-ref quoting, RPM --no-absolute-filenames, and the
+    # hashserv unihash validation. Keeping the race fixes is not worth losing
+    # those, so the branch moved rather than the pin.
+    branch: "race-review-fixes"
```
jetm, line 51: **Pinning a branch that is rebased by policy**

The commit message states that rebasing `race-review-fixes` is "the expected move rather than re-pinning to an older commit", and it has already been force-pushed once.

The next rebase drops the pinned SHA from the branch and `kas/repos.py:546-550` raises `RepoRefError: Branch "race-review-fixes" ... does not contain commit "81d1e0e9..."` - every kas/bakar invocation on this base fails until someone edits `oe.yml`. Loud, which is good. But recovery is not guaranteed: the pre-rebase tip is not on the remote (`git ls-remote --heads` lists only `race-review-fixes` and `mkdirhier-nfs-race`; the commit message says the old tip lives "in the vendor clone", i.e. one laptop), so GitHub can GC an orphaned SHA. And with `KAS_CLONE_DEPTH` set, kas *skips* the containment check and fetches the bare SHA (`repos.py:735-737`), which then fails at fetch with no fallback at all.

Durable shape: cut an immutable tag on the fork (`avocado-2.8-race-review-1`) and pin that, or land the commits on `2.8-avocado`, which every other consumer already tracks. A tag costs one command and removes the whole failure mode - worth it even if the branch never actually gets rebased.

Related cleanup: a third overlapping branch, `mkdirhier-nfs-race` (`cd8838cf`), also exists on the fork remote. Branch sprawl there makes "which ref is authoritative" harder than it needs to be.

kas/vendor/oe.yml
```diff
@@ -15,8 +15,20 @@ repos:
   bitbake:
     url: https://github.com/avocado-linux/vendor-bitbake
     path: bitbake
-    branch: "2.8-avocado"
-    commit: fc13d7197745017a5c8c274cde82b62a9cc796ce
+    # race-review-fixes = 2.8-avocado + mkdirhier EEXIST/ESTALE retries and the
+    # asyncrpc client connect/handshake budgets - cold-review fixes for the
```
jetm, line 19: **The fork pin has no exit criterion**

The comment records why the branch moved but never says what retires the pin, so a future reader cannot tell whether it is still needed.

None of the 14 fork-only commits carries an `Upstream-Status:` trailer or a lore/bitbake-devel link - I checked every body in `81d1e0e9 ^fc13d719`; the only gesture is one line reading "upstream and is tracked separately". Upstream bitbake master has no NFS retry in `mkdirhier` at all (`lib/bb/utils.py:932-944`).

Combined with "rebasing it is the expected move", this becomes a permanent fork branch by default: at every scarthgap or bitbake bump, whoever is on the hook re-derives from scratch whether all 14 commits are still required. One line naming the submission target and a ticket ID fixes it.

Credit where due: the commit body is unusually good otherwise - it names the mkdirhier/asyncrpc fixes and all 11 `2.8-avocado` commits by content, and the "ahead 12, behind 0" claim checks out.

jetm (commented):
Cold second read of my own change. This is a draft, so posting rather than sitting on it - the first finding is blocking and should not wait for un-drafting.

**The pin is reachable and correctly SHA-locked, but it pins the wrong commit.** `81d1e0e9` is two commits behind the branch tip, and those two commits are the ones that actually fix what this pin exists to fix. As it stands the merge would make builds strictly worse than the `fc13d719` they run today.

What I checked and found fine, so nobody re-litigates it:

- **Not a floating pin.** `commit:` sits alongside `branch:`, and kas validates containment via `branch_contains_ref()` -> `git branch origin/<b> -r --contains <commit>`. The branch name is a guard, not the resolved ref.
- **Both refs exist on the named remote.** They are just not the same commit: `refs/heads/race-review-fixes = 7fcf96b2`, and the pinned `81d1e0e9` is an ancestor two back.
- **No conflicting pins.** `kas/vendor/oe.yml` is the only file in the tree naming `vendor-bitbake` or pinning bitbake - checked all 80 `kas/*` files at this head. This fragment wins uncontested.
- **The fork itself is not new.** `scarthgap` already pins `avocado-linux/vendor-bitbake` at `2.8-avocado`. What is new is the topic-branch pin, which is what the third and fourth findings are about.
- The second hunk deletes a trailing blank line at EOF; the file still ends `      meta:` with a newline and parses identically.

Four findings inline. The first two share one fix - re-pin to `7fcf96b2` - so this is a one-line change plus a tag decision.

kas/vendor/oe.yml
jetm: Confirmed at source, and fixed in 417ab17 - but the pin moved to `7fcf96b2`, the current branch tip, not to `f06288b4`.

The mkdirhier defect reads exactly as described. Tail of the function at `81d1e0e9`:

    if not last_attempt:
        time.sleep(0.1 * (attempt + 1))
        continue
    return

At `7fcf96b2` that same position is `raise e`, with a comment saying mkdirhier must never report success without a directory present. So the pinned tree was strictly worse than the scarthgap bitbake it replaces, which raises there.

One correction to my own finding: I wrote that `f06288b4` was "two commits after" the pin. It is **one** - `81d1e0e9..f06288b4` contains exactly `f06288b47 bb/utils: mkdirhier: never report success without a directory`, nothing else, on the same branch. My error, and it propagated into a second reader's summary before anyone checked it.

Re-verified the tip at push time rather than trusting the value recorded when this was written: `git ls-remote` gives `refs/heads/race-review-fixes = 7fcf96b2e00e...`, which is what is now pinned.

kas/vendor/oe.yml
jetm: Right, and this is why the pin went to `7fcf96b2` rather than to the mkdirhier fix alone - `f06288b4` would have closed the first finding and left this one exactly as written.

Measured at both commits:

|  | `81d1e0e9` (old pin) | `f06288b4` | `7fcf96b2` (new pin) |
|---|---|---|---|
| `MAX_CONNECT_RETRIES` | 1 | 1 | 3 |
| `CONNECT_RETRY_BACKOFF` | absent | absent | 0.2 |
| `RETRY_DEADLINE_TIMEOUTS` | absent | absent | 2 |
| wall-clock `deadline` | absent | absent | `client.py:228` |

So the "no longer wedges the client" sentence was not true of the tree it sat next to, and `MAX_CONNECT_RETRIES = 1` with no backoff means the pinned client gave up on a restarting hashserv *faster* than the upstream loop it was meant to improve on. Both close by moving one further commit.

The comment now states which commit supplies which property instead of describing the branch in the abstract, so the next reader can check the claim against the pin without leaving the file.

417ab17.

kas/vendor/oe.yml
jetm: Agreed, and I have not fixed it - only stopped it being silent.

The comment now records the hazard explicitly: this pins a SHA on a branch the fork rebases by policy, the next rebase takes the SHA off the branch, kas raises `RepoRefError` until someone edits this file, and because the pre-rebase tip is not on the remote an orphaned SHA can be collected. It names the two durable shapes - an immutable tag, or landing the commits on `2.8-avocado` where every other consumer already looks.

I stopped short of cutting the tag. That is a write to `avocado-linux/vendor-bitbake` creating an org-visible ref, and it is a decision about how the fork is run rather than a review fix, so it should be yours rather than something that arrives inside a comment-clearing pass.

Worth noting the risk is live, not theoretical: the tip has already moved from `81d1e0e9` to `7fcf96b2` since this PR was opened. Leaving it as a branch pin means the same edit again next time.

417ab17.

kas/vendor/oe.yml
jetm: Standing, and not addressed in 417ab17.

Re-checked rather than carried over: no `Upstream-Status:` trailer or lore link on any of the fork-only commits, and upstream bitbake master still has no NFS retry in `mkdirhier` at all (`lib/bb/utils.py`, local master clone).

Not doing it here for the same reason as the tag: adding trailers across 14 commits rewrites the fork's history, which is a call about how the fork is maintained, not a fix to this PR's diff. Both that and the tag are on the human's desk.

The thing worth deciding is the exit criterion itself, since a fork pin with none is how a temporary vendor branch becomes permanent: either these go upstream and the pin follows, or they land on `2.8-avocado` and the fork branch disappears. Right now nothing in the tree says which.

kas/vendor/oe.yml
mobileoverlord: Decided: merging without the tag. You were right to escalate rather than cut it inside a comment-clearing pass.

To be explicit about what that accepts, so nobody reads the merge as the hazard being gone: the SHA stays pinned on a branch the fork rebases by policy, the next rebase takes `7fcf96b2` off the branch, and this file needs another edit at that point. Your `KAS_CLONE_DEPTH` finding is the part I had not considered — skipping the containment check and fetching the bare SHA turns the loud `RepoRefError` into a fetch failure with no fallback, which is a worse failure than the one the comment describes.

The durable shape is still an immutable tag on the fork, or landing these on `2.8-avocado` where every other consumer already looks. Not a blocker for this PR.

`mkdirhier-nfs-race` on the fork remote should go once its content is accounted for here.

kas/vendor/oe.yml
mobileoverlord: Decided: merging without the trailers, and you are right that the exit criterion is the real question — a fork pin without one is how a temporary vendor branch becomes permanent.

Not asking you to rewrite 14 commit bodies for it. What matters is that the answer lives somewhere outside a reviewer's memory: either these go upstream to bitbake-devel and the pin follows them, or they land on `2.8-avocado` and this branch disappears. Tracking that separately rather than holding this PR.

Confirmed your check independently: upstream bitbake has no NFS retry in `mkdirhier`, so the fork-only surface here is real and not a backport waiting to be dropped.

mobileoverlord (approved):
Approving. Verified rather than read — the kas change is one line, but merging it ships 14 bitbake commits, so that is where I spent the time.

## Checked and holds up

**The base drift is harmless, though the diff reads like a pin revert.** `scarthgap` now pins `2.8-avocado` @ `fc13d719`, not the `82cf21bc` this diffs against. `race-review-fixes` @ `7fcf96b2` is **ahead of `fc13d719` by 14, behind by 0** — nothing that landed on `2.8-avocado` is lost.

**Commit 2's self-criticism is accurate.** At `81d1e0e9` the spent-retry EEXIST path really is a bare `return`, so that pin would have shipped a `mkdirhier` reporting success with nothing at the path — strictly worse than the scarthgap it replaces. And it really carried `MAX_CONNECT_RETRIES = 1` with no deadline or backoff constant. Both claims true, both fixed at the tip.

**The negative-timeout sentinel is real, not cargo.** `StreamConnection.recv()` does `if self.timeout < 0: unbounded readline`, and all three connect paths now honor it. The `asyncio.TimeoutError` → `ConnectionError` conversion is required: `_send_wrapper`'s catch tuple leads with `OSError`, and pre-3.11 `asyncio.TimeoutError` is not one.

**`self.socket is None` is a sound discriminator.** `connect()` assigns the socket before `setup_connection()`, so a handshake failure bills the request budget — which is what the comment claims and the right call, since it cost a round trip rather than a connect timeout. `timeout=None` is unreachable (hashserv and prserv both take the default 30), so `self.timeout >= 0` cannot `TypeError`. The unix path closes the fd on a failed connect, which would otherwise leak per retry.

**I ran the tests.** 11 asyncrpc + 6 mkdirhier, both suites green on 3.14.6. They are properly hermetic — the websocket tests inject a fake `websockets` module, and I confirmed they pass with the package absent.

## Landing anyway, recorded as follow-ups

**Nothing runs those tests.** `vendor-bitbake` has no `.github/workflows` at all and no check runs on `7fcf96b2`; the checks here are CodeQL/Analyze only — no build, no `bitbake-selftest`. I ran them by hand, which does not scale. A silently-dropped commit on a rebase is exactly what produced the `81d1e0e9` regression, so this is the gap worth closing before more fixes arrive this way.

**Merging without the tag.** Both threads below are answered on the PR rather than in the diff, which was the right call — they are decisions about how the fork is run. The fragility is accepted for now; the next rebase will need another edit to this file.

**`RETRY_DEADLINE_TIMEOUTS = 2` is a calibration worth owning.** At the default `timeout=30` that is a 60s ceiling where upstream's 4 attempts allowed ~120s. It is deliberate and tested both ways — `test_the_deadline_does_not_cut_a_call_that_is_making_progress` keeps all 4 attempts for fast failures, so the cut only bites when each attempt burns a full timeout. But a slow-but-alive hashserv now gets 2 tries instead of 4, and `4` would still bound the wedge at 120s while preserving upstream tolerance. Flagging it because the same commit criticized the earlier iteration for exactly this.

**The unix connect is still synchronous.** `settimeout` plus a blocking `connect()` inside the event loop means a stalled unix accept blocks the cooker's loop for up to 30s rather than forever. Strictly better, and it is the pre-existing shape — but "no longer wedges the client" is bounded, not non-blocking. This file's own `BB_HASHSERVE: auto` and `PRSERV_HOST: localhost:0` mean the unix path is the one this deployment actually uses.

Squash-merging with the 14-commit surface spelled out, since the PR body describes two.

PR #243, opened 2026-07-22
intel-x86-64: NVIDIA dGPU driver + kernel/provisioning fixes

mobileoverlord:
## Summary
Enables the NVIDIA discrete-GPU stack on Intel x86-64 and fixes several bugs that blocked GPU bring-up and provisioning. Validated on hardware (Neousys Nuvo-9000, i9-14900 + RTX 4000 SFF Ada): `nvidia-smi` reports the GPU at 595.84, all NICs come up, and a TensorRT-RTX engine builds on the GPU.

All changes are x86-64-scoped and apply to `intel-x86-64-v2/v3/v4`.

## Commits (each a distinct fix)
1. **NVIDIA driver 595.84 + open kernel modules** — bump firmware/userspace/modules from 580.119.02 and make the open kernel modules build on x86-64 (fetch pinned tag via `nobranch=1`; clear `LDFLAGS` for the raw-`ld` prelink; `ARCH=x86_64`; install via `MODLIB` so `kernel-module-split` packages the `.ko`s).
2. **`CONFIG_BINFMT_SCRIPT`** — was unset, so every `#!/bin/sh` systemd `ExecStart` (sshd keygen, `avocado-grow-var`, …) failed `203/EXEC`.
3. **`br_netfilter` + IPv6 netfilter as modules** — completes the container-networking module group (`kernel-module-*` resolvable).
4. **Install the driver on all x86-64 targets** — `PKG_EXTRA_INSTALL` in the shared `avocado-x86-64.inc`, so a complete build produces the driver in the feed for v2/v3/v4 (replaces the v3-only kas override).
5. **Fix x86-64 disk-image provisioning (v2/v3/v4)** — expand-partition sizing, mtools `/dev/tty` hang, string-valued image `jq`, `sgdisk` off-by-one, and sparse writes + progress.


meta-avocado-x86-64/recipes-graphics/nvidia-gpu/nvidia-gpu-userspace.inc
```diff
@@ -12,7 +12,7 @@ NVIDIA_SRC = "${WORKDIR}/${NVIDIA_ARCHIVE}"
 # unpack=0 prevents bitbake from trying to extract the .run file itself;
 # we handle extraction in do_unpack:append below.
 SRC_URI = "https://download.nvidia.com/XFree86/Linux-x86_64/${NVIDIA_DRIVER_VERSION}/${NVIDIA_ARCHIVE}-no-compat32.run;name=run;unpack=0"
-SRC_URI[run.sha256sum] = "1c4623284d8d3c6b843eadc4a878bff6f894bf99fc48169945e19af470ca6d90"
+SRC_URI[run.sha256sum] = "9e4f5d56e74e1ec12a05b2b0afda893c3187da71cbd8fb14c1a394bbeeeb4148"
```
jetm, line 15: **Re-verify the LICENSE md5 after the 580 -> 595.84 bump.**

The `.run` payload sha256 changed here, but `LIC_FILES_CHKSUM = "file://LICENSE;md5=92aa2e2a..."` is inherited unchanged from the 580.119.02 era (and the NOTE right below says it must match the LICENSE extracted from the archive). If NVIDIA altered the LICENSE text in the 595.84 archive, `do_populate_lic` aborts with a checksum mismatch and the whole x86-64 image build fails. Confirm a real `nvidia-gpu-userspace_595.84` / `nvidia-gpu-firmware_595.84` build passed after the bump.

meta-avocado-x86-64/recipes-kernel/nvidia-gpu-modules/nvidia-gpu-modules_595.84.bb
```diff
@@ -6,11 +6,14 @@ HOMEPAGE = "https://github.com/NVIDIA/open-gpu-kernel-modules"
...
+# nobranch=1: SRCREV below is the 595.84 release tag's commit, which is not
+# reachable from the main branch tip, so a branch= check fails ("Unable to find
+# revision ... in branch main"). Pin the exact commit instead.
 SRC_URI = " \
-    git://github.com/NVIDIA/open-gpu-kernel-modules.git;branch=main;protocol=https \
+    git://github.com/NVIDIA/open-gpu-kernel-modules.git;nobranch=1;protocol=https \
     file://nvidia-gpu-modules-load.conf \
 "
-SRCREV = "de3d54abf3dacfb9293d96ed1aa31b449e0d6518"
+SRCREV = "722ae84526a09ed672fbe75448e2909834ba4cce"
```
jetm, line 16: **Confirm this SRCREV is the 595.84 release tag's commit.**

The userspace/firmware ship `libcuda.so.595.84` (PV-driven), but the module builds from this pinned commit with `nobranch=1`. If `722ae84` is not the `open-gpu-kernel-modules` 595.84 tag, the built `nvidia.ko` version string won't match the userspace libraries and the driver fails to load at runtime with an API-mismatch error. Worth verifying the SHA maps to tag `595.84` before merge (I couldn't check the tag mapping offline).

meta-avocado-x86-64/conf/machine/include/avocado-x86-64.inc
```diff
@@ -12,3 +12,11 @@ MACHINE_ESSENTIAL_EXTRA_RDEPENDS += "v86d"
...
 # Ensure kernel modules are built as packages (loadable, not built-in)
 MACHINE_EXTRA_RRECOMMENDS += "kernel-modules"
+
+# NVIDIA discrete-GPU driver stack (open kernel modules + libcuda + nvidia-smi +
+# GSP firmware). Installed on all x86-64 targets so a *complete* build always
+# builds the driver into the feed (and GPU-equipped boards work out of the box).
+# Harmless on boards without an NVIDIA GPU -- the modules just don't load.
+# Applies to v2/v3/v4 (all require this include). This replaces the ad-hoc
+# per-machine kas local_conf_header that only covered v3.
+PKG_EXTRA_INSTALL:append = " packagegroup-avocado-nvidia-gpu"
```
jetm, line 22: **`packagegroup-avocado-nvidia-gpu` is now installed twice.**

This `PKG_EXTRA_INSTALL:append` adds it, but `packagegroup-avocado-x86-64-extra.bb` already lists it in `RDEPENDS:${PN}`. Harmless to the image, but a future "disable NVIDIA on x86-64" edit now has to touch two places and the two can drift. Consider dropping one reference or noting why both exist.

meta-avocado-x86-64/recipes-kernel/linux/files/x86-efi.cfg
```diff
@@ -155,3 +155,17 @@ CONFIG_CUSE=m
...
 CONFIG_NETFILTER_XT_MATCH_ADDRTYPE=m
+
+# Container networking module group. Built as modules -- matching the other
+# Avocado targets -- so container extensions (docker/podman/k8s) can depend on
+# kernel-module-* uniformly across targets, leaving only genuine module-name
+# drift across kernel versions to the kernel-X.Y.* conditional syntax. The IPv4
+# nf/xt modules are already =m via x86_64_defconfig; these fill the gaps:
+#   br_netfilter (was unset -> no kernel-module-br-netfilter), IPv6 nf tables,
+#   and conntrack netlink.
+CONFIG_BRIDGE_NETFILTER=m
```
jetm, line 169: **These netfilter symbols depend on defconfig-provided parents that merge_config drops silently if absent.**

`CONFIG_BRIDGE_NETFILTER` needs `CONFIG_BRIDGE`; the `CONFIG_IP6_NF_*` need `CONFIG_IP6_NF_IPTABLES`+`CONFIG_IPV6`; `CONFIG_NF_CONNTRACK_NETLINK` needs `CONFIG_NF_CONNTRACK`. The comment says these come from `x86_64_defconfig`, and the symbol names are all spelled correctly - but if any parent is missing from the effective defconfig, `merge_config.sh` only warns and drops the symbol, so `kernel-module-br-netfilter` etc. silently never build and container networking deps fail later. Confirm the six land in the built `.config`.

jetm (commented):
Independent review - no correctness defects; the change is coherent and unusually well-commented. Verified: the proprietary-license flag is handled (LICENSE_FLAGS commercial is accepted in kas/base.yml), the installed packagegroup exists, the kernel .cfg symbol names are all real, the three stone-provision-img.sh copies are byte-identical (no copy-paste divergence), the sgdisk +size_sectors fix is correct (no partition overlap/gap), and the module recipe's ARCH/LDFLAGS/MODLIB overrides are sound. Two MEDIUM verify-by-building items and two LOW cleanups inline. One more LOW, non-blocking: the write_image progress loop in all three stone/intel-x86-64-v{2,3,4}/stone-provision-img.sh reads /proc/PID/fdinfo/1 (dd's stdout under status=none), not the output fd, so it logs 0% for the whole write (image is written correctly; the percentage is just wrong). And an informational note: the nvidia recipes gate on COMPATIBLE_HOST=x86_64 rather than COMPATIBLE_MACHINE=intel-x86-64 - fine for this layer since install is scoped via PKG_EXTRA_INSTALL, just wider parse scope than the stated intent.

mobileoverlord:
Thanks @jetm — went through all five.

**Fixed (pushed):**

- **Duplicate nvidia packagegroup** (`avocado-x86-64.inc:22`): dropped `packagegroup-avocado-nvidia-gpu` from `packagegroup-avocado-x86-64-extra`'s `RDEPENDS`; the `PKG_EXTRA_INSTALL` in `avocado-x86-64.inc` is now the single source of truth (`aa93c4f`).
- **Netfilter symbols silently dropped** (`x86-efi.cfg:166`): correct — the built `.config` has `# CONFIG_IPV6 is not set`, so `CONFIG_IP6_NF_*` and `CONFIG_NF_CONNTRACK_NETLINK` never landed (0 hits in the built `.config`). Kept only `CONFIG_BRIDGE_NETFILTER=m` — which *does* land (parent `CONFIG_BRIDGE=m`) and is the one the docker extension needs — and documented why the IPv6 group is left out (`6ff6b0b`).

**Verified (no change needed):**

- **LICENSE md5** (`nvidia-gpu-userspace.inc:15`): the 595.84 `license.txt` is byte-identical to the 580.119.02 one (both `md5=92aa2e2a…`), and `nvidia-gpu-userspace`/`-firmware` 595.84 built cleanly and are running on hardware (`nvidia-smi` → 595.84), so `do_populate_lic` passes.
- **SRCREV = 595.84 tag** (`nvidia-gpu-modules.bb:16`): `git ls-remote … refs/tags/595.84^{}` → `722ae845…`, and the built `nvidia.ko` loads against the 595.84 userspace on hardware with no API-mismatch.

**Not a defect (write_image progress):** GNU `dd` dup2's `of=` onto fd 1 (`STDOUT_FILENO`), so `/proc/PID/fdinfo/1` `pos:` *is* the output-file offset and advances during the write — verified locally (`1 -> …/out.img`, pos 266M → 269M mid-write). It isn't the terminal stdout, so the percentage is real, not stuck at 0%. Happy to switch to scanning `/proc/PID/fd/` for the `of=` link if you'd prefer belt-and-suspenders.

jetm (approved):

PR #244, opened 2026-07-22
imx95-frdm: enlarge SD var partition base to fit the dev /var image

jetm:
## Problem

`avocado provision` for imx95-frdm SD fails in the fwup `rootdisk` step:

```
fwup: file size assertion failed on '...avocado-image-var-imx95-frdm.btrfs'.
Size is 684720128 bytes. It must be <= 536870912 bytes (1048576 blocks)
Error: Failed to provision runtime
```

The dev runtime's `/var` btrfs image is ~653 MiB, but the SD var partition base is pinned at 512 MiB.

## Root cause

`meta-avocado-nxp/stone/stone-imx95-frdm.json` set the var partition `size: 512` MiB. That becomes `AVOCADO_PARTITION_VAR_BLOCKS=1048576`, enforced by `avocado-imx/rootdisk.conf` `assert-size-lte = ${VAR_PART_BLOCKS}`. `expand: true` grows the partition to fill the media at first boot, but the flashed base must still be `>=` the image, and the dev `/var` image outgrew 512 MiB.

## Fix

Raise the var partition base to 1024 MiB. `expand: true` is unchanged, so runtime behavior is identical (the partition still grows to fill the SD at first boot); this only lifts the fwup ceiling so the larger image is accepted.

## Testing

- `stone-imx95-frdm.json` validates as JSON; the diff is the single `size` line (512 -> 1024).
- Not provision-tested on hardware in this environment (no imx95-frdm board). Needs a validation flash on the board: confirm the fwup `rootdisk` step passes and `/var` expands to the SD at first boot.

## Notes

- `stone-qemuarm64.json` also uses a 512 MiB var base and will hit the same wall as its dev `/var` grows; a follow-up should bump it (or lift the var base into a shared default) so targets do not drift.

### DIFF
--- a/meta-avocado-nxp/stone/stone-imx95-frdm.json
+++ b/meta-avocado-nxp/stone/stone-imx95-frdm.json
@@ -132,8 +132,6 @@
         {
           "name": "var",
           "image": "var",
-          "size": 512,
-          "size_unit": "mebibytes",
           "expand": "true"
         }
       ]


jetm (commented):
Second-reviewer pass on my own change. The bump itself is arithmetically correct - I checked the units through stone (`manifest.rs:358` converts mebibytes to bytes, fwup gets blocks via `provision.rs:615/675`) and the 512 MiB ceiling in the error message reconciles with the 536870912-byte assertion. `var` is the last partition and nothing after `boot-a` carries a fixed offset, so no sibling moves and the layout still fits under a 2 GB card.

The finding is that the constant should not be there at all - this board is the only one of 23 that pins `var`, which is exactly what stops it auto-tracking the image the way the other 22 do. Raising it buys one cycle; removing it ends the class. Detail inline.

I checked for other machines with the same outlier and found none.

meta-avocado-nxp/stone/stone-imx95-frdm.json
```diff
@@ -132,7 +132,7 @@
         {
           "name": "var",
           "image": "var",
-          "size": 512,
+          "size": 1024,
```
jetm: **This constant should be deleted, not raised - it is why the board needs a manual bump at all.**

The arithmetic in the bump is fine (1024 MiB clears the ~653 MiB image, units are consistent, `var` is last so nothing overlaps and no offset has to move). But an explicit `size` on `var` shadows the `--partition-size var=<bytes>` override that `avocado-stone.bb` passes on every bundle and provision, so this is the only board of 23 whose `/var` ceiling does not auto-track the built image.

`stone/src/manifest.rs:390-409` - `resolve_partition_size_bytes` early-returns on `if let Some(size) = p.size` and never consults the override map. The doc comment at `manifest.rs:335-337` says size "May be omitted only on the last partition when `expand: \"true\"` ... supplied at bundle/provision time via `--partition-size`", and `avocado-stone.bb:86-99` states the intent verbatim: "The var partition omits size in stone manifests so its size is computed from the freshly-built var image and rounded up to size_alignment".

Every other manifest - `stone-imx91-frdm.json`, `stone-imx93-frdm.json`, `stone-raspberrypi5.json`, `stone-intel-x86-64-v3.json` and the rest - omits `size`/`size_unit` on `var`. imx95-frdm is the sole outlier, before and after this change.

So the failure recurs: the dev `/var` went 512 to ~653 MiB in one cycle, and the next time it crosses 1024 MiB `do_stone_bundle` fails again with the same `file size assertion failed ... must be <= 1073741824 bytes`, while every other board absorbs identical growth silently. There is a cost today too - the GPT declares 1024 MiB for a 653 MiB payload, making the layout's minimum media requirement ~371 MiB larger than auto-sizing would produce.

Deleting both `"size": 1024` and `"size_unit": "mebibytes"` from the `var` entry is the fix. `var` is last and has `expand: "true"`, so it satisfies the omit precondition and `stone validate` (`manifest.rs:441-471`) accepts it.

jetm: Done in 3d5b924 - deleted `size` and `size_unit` rather than raising them, so `var` now resolves through the `--partition-size` override like the other 22 boards.

Verified the two preconditions before deleting rather than assuming them. `resolve_partition_size_bytes` (`manifest.rs:394`) really does `return` inside `if let Some(size) = p.size` before it ever reads the overrides map, and `avocado-stone.bb:96,108` passes `--partition-size "var=$var_image_bytes"` from `stat -c%s` on the freshly-built image in both `do_stone_bundle` and `do_stone_provision`. `validate_partitions` (`manifest.rs:442`) accepts the omission: `var` is last in `rootdisk`, carries `expand: "true"`, and is named.

Also re-ran the outlier count as a sweep instead of quoting it from memory - 23 manifests define a `var` partition, all 23 set `expand: "true"`, and imx95-frdm was the only one carrying `size`. The entry is now byte-identical to `stone-imx93-frdm.json`'s.

The commit message changed with the fix, since it no longer describes a bump.

mobileoverlord (approved):

PR #245, opened 2026-07-23
Compile and deliver declared device-tree overlays

jetm:
## Problem

Avocado had no declarative way to ship a device-tree overlay. Users hand-compiled `.dtso` sources and manually placed the `.dtbo` on the boot medium.

## Solution

Add an SDK wrapper that compiles declared overlays the way the kernel does (`cpp` against `dt-bindings` when `#include`d, `dtc -@` for overlay symbols), plus per-BSP delivery hooks that place each `.dtbo` on the target's boot medium for stone to bundle.

## Key changes

- Add the SDK device-tree overlay builder (`avocado-dtc-overlay`) in the shared layer
- Add per-BSP delivery hooks for qemuarm64 (boot FAT) and Raspberry Pi
- Relocate `avocado-multikernel` out of the distro layer (fixes an SDK-container build regression)
- Harden the wrapper: require `dt-bindings` only on `#include`, handle multi-token CPP, verify `fdtget` before use

## Reviewer notes

Part of the device-tree-overlay feature (ENG-2134). Targets `scarthgap` (feature groups already merged there). Depends on the stone `files_append` PR - the `stone: bump` commit points at that branch; retarget once stone merges. avocado-cli's DTO commands depend on this PR. Validated end to end on raspberrypi5: overlay compiled, delivered to the boot FAT, and `os-bundle.aos` finalized.


meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
```diff
@@ -0,0 +1,174 @@
...
+        || die "$label: preprocessing failed"
+else
+    dtc_src="$src"
+fi
+
+# -@ (overlay symbols): emits __symbols__ for labels this overlay defines so a
+# sibling overlay or the base can reference them. dtc warnings are expected on
+# real hardware overlays (reg formatting, etc.) and must not fail the build; a
+# genuine syntax/structure error makes dtc exit non-zero and we surface it.
+"$DTC" -@ -I dts -O dtb -o "$out" "$dtc_src" \
```
jetm: **`dtc` invoked with no `-i` for the source directory**

This call omits the `-i <source dir>` include path the kernel's own dtb rule passes, so dtc cannot resolve an `/include/`d `.dtsi` once cpp has moved the compilation unit into the temp dir.

Failure: an overlay containing both `#include <dt-bindings/gpio/gpio.h>` and `/include/ "frag.dtsi"` (frag.dtsi beside the source) sets `needs_bindings=true`, so :144 sets `dtc_src="$work/preprocessed.dts"` and this line runs `"$DTC" -@ -I dts -O dtb -o "$out" "$dtc_src"` with no `-i`. dtc aborts and the wrapper dies "dtc failed to compile the overlay".

Verified on dtc v1.8.1: cpp into a separate tmpdir then `dtc -@ -I dts -O dtb -o out2.dtbo tmpdir/preprocessed.dts` printed `FATAL ERROR: Couldn't open "frag.dtsi": No such file or directory`; adding `-i srcdir` to the identical command returned rc=0.

`linux-stable/scripts/Makefile.dtbs:134` is `$(DTC) -o $@ -b 0 $(addprefix -i,$(dir $<) $(DTC_INCLUDE))` for exactly this reason - its `dtc-tmp` also lives outside the source dir.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
```diff
@@ -0,0 +1,174 @@
...
+    || die "$label: dtc failed to compile the overlay"
+
+# Structural check: a real overlay materialises at least one fragment@ node.
+root_nodes="$(fdtget -l "$out" / 2>/dev/null || true)"
+grep -q '^fragment@' <<<"$root_nodes" \
```
jetm: **Structural check greps the blob for a node named `^fragment@`**

This requires a root node literally named `fragment@...`, but that name is convention only - the kernel and libfdt both select on the presence of an `__overlay__` child - so a valid overlay using any other node name is hard-rejected.

Failure: an overlay whose top-level node is `my_spi_frag { target-path = "/"; __overlay__ { ... }; };` compiles cleanly, but `grep -q '^fragment@' <<<"$root_nodes"` fails and :161 dies "compiled blob has no overlay fragments (is this actually an overlay?)".

Verified: compiled that source on dtc v1.8.1 (rc=0); `fdtget -l out.dtbo /` printed `my_spi_frag` and the grep found no match. `linux-stable/drivers/of/overlay.c` `init_overlay_changeset` iterates `for_each_child_of_node(ovcs->overlay_root, node)` selecting on `of_get_child_by_name(node, "__overlay__")`, never on the name; its kerneldoc says "Any other top level node will be ignored."

Test for an `__overlay__` child instead of the node name.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
```diff
@@ -0,0 +1,174 @@
...
+    || die "$label: compiled blob has no overlay fragments (is this actually an overlay?)"
+
+# If the overlay targets an external label via a fragment (target = <&name>),
+# that reference is resolved at apply time from the base's __symbols__ and MUST
+# be exported as a __fixups__ entry. Its absence means the reference was baked
+# to a literal phandle that can never bind - a silent boot-time failure. Path-
+# targeted and label-free overlays legitimately have no __fixups__, so this
+# only fires when a fragment target phandle is actually present.
+if grep -Eq 'target[[:space:]]*=[[:space:]]*<[[:space:]]*&' "$dtc_src"; then
```
jetm: **Fixups guard greps source text for `target = <&`, and is wrong in both directions**

Gating the `__fixups__` assertion on a line-oriented grep of source text fires on input it should accept and stays silent on input it should catch.

**False positive:** a path-targeted overlay with no `#include` and the comment `/* example: target = <&spi0>; */` takes the raw-source branch (:148 `dtc_src="$src"`), the grep matches the comment, the blob legitimately has no `__fixups__`, and :171 kills the build. Verified on dtc v1.8.1: the grep matched, and `fdtget -l comment.dtbo /` printed only `fragment@0`.

Worse, adding one `#include` to that same file flips the verdict, because cpp strips the comment (`grep -c 'target = <&spi0>' pre.dts` -> 0). So the wrapper judges identical overlay semantics differently based on an unrelated header.

**False negative:** `/dts-v1/; /plugin/; &spi0 { status = "okay"; };` - the standard shorthand - produces a blob with `__fixups__` but contains no `target` token, so the assertion never runs. Verified: `fdtget -l short.dtbo /` printed `fragment@0` and `__fixups__` while the grep found nothing. A newline-split `target\n    = <&spi0>;` behaves the same, since `[[:space:]]` cannot cross a newline in line-based grep.

Testing the compiled blob for a fragment `target` property closes all four cases at once - and it is the same shape of fix as the `fragment@` thread above, so both are one change.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
```diff
@@ -0,0 +1,174 @@
...
+src=""
+out=""
+kdir="${AVOCADO_KDIR:-}"
+kdir_explicit=false
+[[ -n "$kdir" ]] && kdir_explicit=true    # $AVOCADO_KDIR is a deliberate choice
+name=""
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --src)  src="${2:-}"; shift 2 ;;
```
jetm: **Option arms run `shift 2` when the value is absent**

Each arm tolerates a missing value with `${2:-}` but then runs `shift 2`, which returns non-zero with one positional left. Under `set -euo pipefail` that aborts before the required-argument diagnostics at :61-62 can run.

Failure: `avocado-dtc-overlay --src ov.dtso --out` - e.g. an unquoted empty expansion in the calling shell leaving an option as the final token. The `--out` arm at :53 runs `out="${2:-}"; shift 2` with `$#==1`, shift fails, and the script exits 1 with **no message at all** instead of printing usage and "--out is required".

Verified: a reduced copy of the :50-59 loop under `set -euo pipefail` on bash 5.3.15, with `--src` alone and with `--src a.dtso --out`, exited 1 with empty stdout and stderr and never reached the post-loop marker.

The `${2:-}` defaults show fall-through to the required-arg check was the intent. `shift "$(( $# > 1 ? 2 : 1 ))"` or an explicit `[[ $# -ge 2 ]]` guard fixes it.

meta-avocado-raspberrypi/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
```diff
@@ -0,0 +1,119 @@
...
+        die("hook invoked with no overlays to deliver")
+
+    with open(stone_path) as f:
+        stone = json.load(f)
+    dev_key, img_key = find_boot_fat(stone)
+    platform = stone.get("runtime", {}).get("platform", "raspberrypi")
+
+    # Additive delivery: each .dtbo into overlays/, plus the include file.
+    files_append = [
+        {"in": ov["file"], "out": f"overlays/{ov['file']}"} for ov in overlays
```
jetm, line 123: **An overlay name matching a stock `.dtbo` is silently dropped**

This emits `in` as the bare `<name>.dtbo`, byte-identical to how the BSP stone manifests name their stock overlays - so a colliding user overlay hits stone's idempotent-duplicate branch instead of its conflict branch and is discarded, while `claimed_by` still passes the CLI gate.

Failure: on avocado-raspberrypi5 a user declares `device_tree_overlays: [{name: dwc2, src: my-dwc2.dtso}]`. The wrapper compiles `$DTBO_STAGING/dwc2.dtbo`. This line emits `{"in":"dwc2.dtbo","out":"overlays/dwc2.dtbo"}`, identical to `stone-raspberrypi5.json:137`. stone `src/manifest.rs:721-736` sees the same `output_name()`, finds `input_filename()` equal, and `continue`s at :735, dropping the append. `find_file_in_dirs` (stone `bundle.rs:111-119`) then resolves `dwc2.dtbo` from the first `-i` dir, `$AVOCADO_SDK_PREFIX/stone` (avocado-cli `build.rs:2423`), which precedes `$DTBO_STAGING` (avocado-cli `device_tree_overlay.rs:327`) - so the **stock** blob ships. :109 already set `claimed_by`, so the unclaimed check at avocado-cli `device_tree_overlay.rs:238` passes. Build exits 0, `dtoverlay=dwc2` applies the stock overlay, and the user's overlay is nowhere on the image.

Same exposure for `vc4-kms-v3d-pi5`/`bcm2712d0` (pi5), `vc4-kms-v3d-pi4`/`dwc2` (pi4, fr202), `i2c3`/`reTerminal`/`dwc2` (reterminal).

Note stone's conflict branch is structurally unreachable for this shape, so the fix has to be here - namespace the `in` filename, or detect the stock name and refuse.

meta-avocado-qemu/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
```diff
@@ -0,0 +1,97 @@
...
+    fragment = {
+        "storage_devices": {
+            dev_key: {"images": {img_key: {"build_args": {"files_append": files_append}}}}
+        }
+    }
+    with open(fragment_path, "w") as f:
+        json.dump(fragment, f, indent=2)
+
+    for ov in overlays:
+        ov["claimed_by"] = platform
```
jetm, line 100: **This hook sets `claimed_by` though its own header says selection is unwired**

The header at :14-18 concedes boot-time selection is not wired - "wiring u-boot's apply_overlays / fdt_overlays so the overlay is actually applied onto the DTB is a separate step" - yet this line sets `claimed_by` on every overlay. That is the exact signal the RPi hook documents as existing "so the CLI can fail the build if any declared overlay went undelivered" (:29-31), so setting it here disarms the one gate that could have caught the gap.

Failure: build avocado-qemuarm64 with a declared overlay. `claimed_by` is set, the in-container check (avocado-cli `device_tree_overlay.rs:238`) tests only `not o.get("claimed_by")` so it passes, and `avocado runtime build` exits 0 with a normal success line. At boot `meta-avocado-qemu/recipes-bsp/u-boot/u-boot/env/avocado-qemuarm64.txt` runs `avocado_boot=booti ${image_addr} ${ramdisk_addr}:${filesize} ${fdt_addr};` - no `fdt addr`, no `fdt apply`, no load of `overlays/*.dtbo` - so the QEMU-generated DTB reaches Linux unmodified.

The only warning is a parenthetical on stdout at :93, inside a multi-thousand-line build log. The user concludes the overlay is applied and debugs the wrong layer.

Either do not set `claimed_by` until selection is wired, or fail the build outright on this platform. Half-claiming is worse than either.

meta-avocado-qemu/recipes-avocado/packagegroups/packagegroup-avocado-extra.bbappend
```diff
@@ -0,0 +1,7 @@
+# Pull the QEMU device-tree overlay delivery hook into the SDK target sysroot
+# alongside kernel-devsrc, so the avocado CLI finds it at
+# $OECORE_TARGET_SYSROOT/usr/libexec/avocado/device-tree-overlay-deliver when
+# building a qemu runtime that declares device-tree overlays. This bbappend is
+# only parsed for builds that include meta-avocado-qemu, so non-qemu builds
+# never reference the qemu recipe.
+SDK_SYSROOT_DEPENDS:append = " avocado-dtc-overlay-deliver"
```
jetm: **qemux86-64 accepts DT overlays and reports success**

meta-avocado-qemu also owns `avocado-qemux86-64`, an x86 machine that boots via `zboot` with no device tree at all, yet this `SDK_SYSROOT_DEPENDS:append` installs the delivery hook unconditionally - and the hook's `KERNEL_NAMES` tuple contains `bzImage`. So an x86 build declaring device-tree overlays passes every gate.

Failure: build `avocado-qemux86-64` with one declared overlay. dtc compiles the `.dtbo` (arch-agnostic). `find_boot_fat` matches `bzImage` from `meta-avocado-qemu/stone/stone-qemux86-64.json:88`, the fragment appends `overlays/x.dtbo`, the hook sets `claimed_by`, the unclaimed check passes, `stone bundle` succeeds, exit 0. At boot `env/avocado-qemux86-64.txt` runs `avocado_boot=zboot ${kernel_addr_r} 0 ${ramdisk_addr_r} ${filesize}` - no `${fdt_addr}`, no FDT anywhere in the chain.

Green build, dead file, and no message that overlays are meaningless on this target. Gate the append on a machine that actually has an FDT, or have the hook refuse on x86.

meta-avocado-qemu/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
```diff
@@ -0,0 +1,97 @@
...
+        for img_key, img in (dev.get("images") or {}).items():
+            if not isinstance(img, dict):
+                continue
+            if img.get("build_args", {}).get("type") != "fat":
+                continue
+            for entry in img["build_args"].get("files", []):
+                out = entry.get("out") if isinstance(entry, dict) else entry
+                base = str(out).rsplit("/", 1)[-1]
+                if base in KERNEL_NAMES:
+                    return dev_key, img_key
```
jetm: **`find_boot_fat` takes the first match with no multi-match guard**

This returns the first FAT image that matches and never checks whether a second also matched, so the delivery target is decided by textual position in the stone JSON. Same shape in the RPi hook at :57.

Failure: fill the reserved `recovery` slot - `stone-qemuarm64.json:92-97` and `stone-raspberrypi5.json:177-182` both hold `"image": ""` placeholders today - with a real `{"build_args": {"type": "fat", "files": ["Image", ...]}}`, declared above `boot` in the `images` object. Python 3.7+ `json.load` preserves document order, so the loop at :42-52 returns `recovery` and never reaches `boot`. Overlays append to the recovery FAT, `claimed_by` is set, exit 0, and the normal boot slot ships without them.

The stdout line does name `recovery`, but nothing errors and no assertion says exactly one FAT must match. There is `die(...)` at :53 for zero matches and nothing for two - add a counter and assert uniqueness.

meta-avocado-raspberrypi/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
```diff
@@ -0,0 +1,119 @@
...
+    for dev_key, dev in devices.items():
+        images = dev.get("images", {})
+        for img_key, img in images.items():
+            if not isinstance(img, dict):
+                continue
+            build_args = img.get("build_args", {})
+            if build_args.get("type") != "fat":
+                continue
+            for entry in build_args.get("files", []):
+                out = entry.get("out") if isinstance(entry, dict) else entry
```
jetm: **`entry.get("out")` is None for implied-out `files` entries**

Boot-FAT discovery matches only on an explicit `out` key, so a stone `files` entry of the form `{"in": "bootfiles/config.txt"}` that lets `out` default to the basename is never recognised.

Failure: a manifest whose FAT lists `[{"in": "bootfiles/config.txt"}, {"in": "Image"}]`. This line's `entry.get("out") if isinstance(entry, dict) else entry` yields None, :56's `out == "config.txt"` is False for every entry, and :58 hard-errors "no FAT image ships config.txt in the stone manifest" on a valid manifest. Same shape at the QEMU hook :49, where `str(None).rsplit("/", 1)[-1]` is `"None"` and never matches `KERNEL_NAMES`.

Plausible rather than confirmed: the code path is unambiguous, but whether stone permits an implied `out` is unverified - the recipe pins a remote SRCREV (`stone_2.2.0.bb`) rather than vendoring, so I could not read the schema. Settle it by reading stone's `files`/`files_append` schema or running either hook against a manifest using the implied-out form.

Worth noting either way: both test fixtures use explicit `out` or bare strings only, so the suites cannot catch this whichever way it goes.

jetm (commented):
Cold second read of my own change. Two caveats up front, because they change how much weight to put on this.

**This review did not finish.** The pass ran out of budget partway: two finder angles reported and then the verify, dedup and gap-sweep stages never ran, and neither did the cross-file angle. So nothing below has been challenged by a second reader, the list is not deduplicated, and - more importantly - **absence of a finding here means nothing.** I am posting anyway because nine of these carry their own reproduction and a draft with findings beats findings in a file, but this is not a completed review and should not be treated as one.

**Confidence is per-finding, stated on each thread.** Eight are CONFIRMED with a command output or a source citation; one (the implied-`out` key) is PLAUSIBLE because I could not read stone's schema - the recipe pins a remote SRCREV rather than vendoring.

The findings split cleanly in two. The wrapper's four are all the same class: it decides things by grepping source text where it could read the compiled blob it already has in hand. The delivery hooks' five are worse in kind - each ships a green build whose overlay is absent or inert, which is the failure shape a user cannot diagnose. I would fix the hook findings first for that reason, even though the wrapper ones are cheaper.

Three more that did not get their own thread:

**The base-DTB binding check is absent by design, and I want to reconsider that.** The wrapper's comment at :22-26 says it \"deliberately does NOT try to prove the overlay will apply to a particular base\", and :163-172 assert only that a `__fixups__` node exists at all. So a target label the base does not define produces a fully green build and a boot with no effect - e.g. `target = <&spi0>` where `bcm2712-rpi-5-b.dtb` exports it as `spi10` in `__symbols__`. dtc emits a `__fixups__` entry naming `spi0`, the guard is satisfied, the `.dtbo` ships, `dtoverlay=my-spi` is written, `claimed_by` is set, exit 0, and at boot the firmware fails fixup resolution and skips the overlay. The reason to revisit the decision: the base DTBs needed to falsify this are already on the same FAT image at bundle time (`stone-raspberrypi5.json:138`), so the check is much cheaper here than the comment assumes.

**Unguarded dict subscripts on manifest entries** - RPi hook :89, :62, :113 and QEMU hook :75, :90 read entries with bare subscripts rather than the `die()` path the rest of the file uses. An entry `{\"file\": \"my-spi.dtbo\", \"params\": {}}` with no `name` reaches `overlay_line` at :62, which evaluates `overlay['name']` and raises KeyError - dumping a Python traceback into the build log instead of the `[ERROR] device-tree-overlay-deliver (raspberrypi): ...` message the hook emits for every other input defect. The version gate at :76 shows per-entry shape was meant to be validated; only the envelope is. Not executed - a repro script was blocked by a sandbox guard - so treat as plausible.

**A user overlay can capture a stock `dtoverlay=` line on seeed and fr200** (`rpi-config_git.bbappend:19` vs `:40`). Those machines' config.txt carries hard-coded `dtoverlay=` directives for overlay names absent from the stone `files` list, so a user overlay taking one of those names is delivered normally (no merge collision) and then captures the pre-existing directive - and because the appended `include` precedes those directives, the same blob is loaded twice with different parameters. On avocado-reterminal, an overlay named `i2c1` gets `include avocado-overlays.txt` processed first with the user's params, then `dtoverlay=i2c1,pins_2_3` applied again. Exposed names on those machines: `i2c1`, `i2c5`, `i2c6`, `spi4-1cs`, `spi3-1cs`, `spi6-1cs`, `tpm-nuvoton`, `i2c-rtc`, `audremap`, `imx219`, `reComputer-R100x`. The ordering is confirmed from bitbake source (`ast.py:183` -> `data_smart.py:567-574` accumulates `:append` into a flag list, `:832` applies it in registration order). What needs a real Pi to settle is whether the duplicate `dtoverlay` is a double-apply, a duplicate-node failure, or benign - the config.txt docs do not specify.

Re-run the full review before trusting that this list is complete.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
mobileoverlord: **Priority — confirmed.** Reproduced on dtc 1.8.1: an overlay carrying both `#include <dt-bindings/gpio/gpio.h>` and `/include/ "frag.dtsi"` (frag.dtsi beside the source) sets `needs_bindings=true`, so `dtc_src` becomes `$work/preprocessed.dts` and this line runs without `-i`:

```
FATAL ERROR: Couldn't open "frag.dtsi": No such file or directory
[ERROR] avocado-dtc-overlay: withinc.dtso: dtc failed to compile the overlay
```

Adding `-i "$(dirname "$src")"` returns rc=0. Worth doing unconditionally rather than only on the cpp branch, so behaviour does not depend on whether the source happens to `#include` anything — that coupling is the same trap as the fixups guard below.

This is a capability gap rather than a wrong result (the user is blocked with a message, not misled), but it is a one-flag fix that makes the wrapper accept a documented, kernel-supported source layout it currently refuses.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
mobileoverlord: **Priority — confirmed.** This is the clearest case of the wrapper being stricter than the thing it wraps. Reproduced on dtc 1.8.1 with a valid overlay whose top-level node is not named `fragment@N`:

```
$ dtc -@ -I dts -O dtb -o direct.dtbo named.dtso   # rc=0
$ fdtget -l direct.dtbo /
my_spi_frag
$ avocado-dtc-overlay --src named.dtso --out named.dtbo
[ERROR] avocado-dtc-overlay: named.dtso: compiled blob has no overlay fragments (is this actually an overlay?)
```

Your reading of `drivers/of/overlay.c` matches — `init_overlay_changeset` selects on `of_get_child_by_name(node, "__overlay__")` and never on the node name, and the kerneldoc says any other top-level node is ignored. `fragment@N` is dtc-generated convention for the `&label { }` shorthand, not part of the format.

Testing for an `__overlay__` child instead makes the check assert the property that actually matters, and stops the wrapper rejecting overlays the kernel would apply.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
mobileoverlord: **Priority — same fix as the `fragment@` check above.** Agreed it is wrong in both directions, and the direction I care about most is the one that makes identical overlay semantics compile or fail depending on an unrelated `#include`: on the raw-source branch a `/* target = <&spi0>; */` comment matches and kills a valid path-targeted overlay, and adding one header makes cpp strip the comment so the same file passes.

The false negative matters less in kind — a guard that silently does not run leaves you where you were before the guard existed, rather than producing a wrong artifact — but `&spi0 { status = "okay"; };` is the shorthand most overlays in the wild are written in, so in practice the assertion is off for the common case and mis-firing on the uncommon one.

Inspecting the compiled blob for a fragment `target` property closes all four cases and shares its shape with the check above, so I would land the two together.

mobileoverlord (commented):
Reviewed the wrapper, both hooks and the recipes, ran all three shipped suites (18/18 pass on a host with dtc/cpp/fdtget), and exercised the full stack end to end: compiled an overlay with the wrapper, ran the RPi hook against the real `stone-raspberrypi5.json`, fed the resulting fragment to stone 2.3.0, and confirmed `overlays/bbproto.dtbo` plus `avocado-overlays.txt` land in the boot FAT with `config.txt` untouched. `find_boot_fat` also resolves correctly on all six shipped RPi manifests. The architecture is sound and the layering is right — generic wrapper, generic stone, per-BSP hooks holding BSP knowledge.

Of your nine self-review findings, these three are the ones I'd prioritise, and for a specific reason: they are the only ones where the **generic** wrapper hard-rejects input that is valid on every target. The rest are either target-specific knowledge that belongs with the developer, a naming rule we can document, or latent. These three make the tool less flexible than the thing it wraps, which is the opposite of what it is for.

All three assert against the wrong artifact — source text or a node name, rather than the compiled blob. #2 and #3 are one fix. Reproductions inline; each is independently confirmed on dtc 1.8.1, not just a re-read of your notes.

One of the nine I think you can close outright: the `entry.get("out")` implied-out concern, which you flagged as plausible-not-confirmed because the recipe pins a remote SRCREV. I have the source — `FileEntry::Object.output` carries no `#[serde(default)]`, so `out` is mandatory, and a manifest using the implied form fails to parse outright:

```
Failed to parse manifest JSON: data did not match any variant of untagged enum FileEntry at line 161
```

So `{"in": "bootfiles/config.txt"}` cannot appear in a manifest at all, and both hooks' `entry.get("out")` is safe as written. Not a bug.

Nothing here blocks from my side. Note the ruleset enforces review-thread resolution, so the other six threads need a reply before this can merge.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
jetm: Fixed in 820f471. `avocado-dtc-overlay:169` now passes `-i "$(dirname "$src")"`.

Unconditional rather than only on the cpp branch, per your follow-up: whether an overlay happens to `#include` a header should not decide how its `/include/` resolves.

I reproduced it first on dtc 1.8.1 and got your `FATAL ERROR: Couldn't open "frag.dtsi"` exactly. The new test case pairs `#include <dt-bindings/gpio/gpio.h>` with `/include/ "frag.dtsi"`, because it takes the `#include` to relocate the compilation unit and the `/include/` to need the source dir — neither alone reproduces.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
jetm: Fixed in 820f471. The structural check now walks the root nodes and selects a fragment on the presence of an `__overlay__` child (`avocado-dtc-overlay:184-196`), not on the node name.

Your reading of `init_overlay_changeset` is what I verified against: it selects on `of_get_child_by_name(node, "__overlay__")` and the kerneldoc says any other top-level node is ignored. `fragment@N` is just what dtc generates for the `&label { }` shorthand.

Reproduced before fixing — a top-level `my_spi_frag` compiled at rc=0 and `fdtget -l` showed it, while the wrapper rejected it as "no overlay fragments". There is a test for that shape now.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
jetm: Fixed in 820f471, and you were right that it is one change with the `fragment@` check — both now come from a single walk of the blob (`avocado-dtc-overlay:184-196`).

A fragment counts as targeting an external phandle when it carries a `target` property. The match is anchored (`grep -qx 'target'`) so `target-path`, the path-targeted spelling, cannot satisfy it — I checked both blobs directly: the `&spi0` shorthand fragment lists `target`, the path-targeted one lists only `target-path`.

Both directions have tests: the commented-out `target = <&spi0>` case failed on the pre-fix wrapper, and the `&label { }` shorthand is kept as a regression guard since that is the form the source grep silently skipped.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
jetm: Fixed in 820f471. Each arm now shifts through `shift_opt` (`avocado-dtc-overlay:55`), which yields 2 only when a value is actually present.

Your reading of the intent was right — the `${2:-}` defaults were there to fall through to the required-argument checks, and the flat `shift 2` killed the script before they could run. I confirmed the exact symptom you described: `--out` as the final token exited 1 with both stdout and stderr empty.

The test asserts the diagnostic now appears rather than just asserting non-zero, since a bare exit 1 was the bug.

meta-avocado-raspberrypi/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
jetm: Fixed in 820f471. The Pi hook now reads the boot FAT's existing outputs and refuses by name when a declared overlay would land on one of them (`device-tree-overlay-deliver:113-118`).

I confirmed the collision shape you traced: the stock entries in `stone-raspberrypi5.json` are `{"in": "<name>.dtbo", "out": "overlays/<name>.dtbo"}`, which is byte-identical to what the hook emitted, so stone's idempotent-duplicate branch takes it and the append is dropped.

Refusing rather than namespacing, because the delivered filename is what `dtoverlay=<name>` resolves at boot — namespacing the file would silently desynchronise it from the `avocado-overlays.txt` line unless both moved together, and a rename by the user is the honest fix. The error names the colliding overlay and says the stock one would otherwise ship.

Test covers it against a manifest that already ships `overlays/my-spi.dtbo`.

meta-avocado-qemu/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
jetm: Fixed in 820f471, in both hooks — qemu at `device-tree-overlay-deliver:49-66`, Pi at `:59-73`.

Each now collects every matching FAT and refuses when there is more than one, naming them, rather than returning the first. You are right that this was decided by document order: `json.load` preserves it, so a filled-in recovery slot declared above the boot slot would have taken delivery on textual position alone.

Both suites have a test feeding two matching FATs with the recovery slot first.

meta-avocado-qemu/recipes-avocado/packagegroups/packagegroup-avocado-extra.bbappend
jetm: Fixed in 820f471. The dependency is now scoped to the machine (`packagegroup-avocado-extra.bbappend:16`) rather than applied layer-wide.

Your trace is what I confirmed: `KERNEL_NAMES` contains `bzImage`, so on x86 the hook found a "boot" FAT, appended the overlay and set `claimed_by`, and every gate passed on a machine whose boot chain has no FDT anywhere in it.

I took the gating option rather than making the hook refuse. x86 now gets no hook at all, so the CLI's undelivered check fails the build — which puts the error at the point where a device-tree overlay was requested for a machine that has no device tree, rather than after compiling and staging one.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
jetm: Done in 820f471, and I took the unconditional form you asked for — `-i "$(dirname "$src")"` is passed on both branches, so an unrelated `#include` cannot change how `/include/` resolves. Details in the parent thread.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
jetm: Done in 820f471 — the check now selects on an `__overlay__` child rather than the node name, so an overlay the kernel would apply is no longer rejected for what its fragment is called. Details in the parent thread.

meta-avocado-shared/recipes-avocado/avocado-dtc-overlay/files/avocado-dtc-overlay
jetm: Done in 820f471, and folded into the same blob walk as the `fragment@` check, as you suggested. The `#include`-flips-the-verdict case is what the new test pins: a commented-out `target = <&spi0>` no longer trips the assertion on either branch. Details in the parent thread.

meta-avocado-raspberrypi/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
jetm: I settled this the way you suggested — by reading stone's schema — and it comes out the other way, so I have not changed the code. Leaving the thread open in case you read it differently.

`FileEntry` in `stone/src/manifest.rs:288-296` is an untagged enum of exactly two shapes:

```rust
pub enum FileEntry {
    String(String),
    Object {
        #[serde(rename = "in")]  input: String,
        #[serde(rename = "out")] output: String,
    },
}
```

`output` is `String`, not `Option<String>`. An object entry with no `out` matches neither variant, so serde fails the whole manifest at deserialisation — stone rejects it rather than defaulting the basename. The implied-out form is therefore not something a valid manifest can contain, and the two shapes that can are both already handled: a bare string falls to the `else` branch of `entry.get("out") if isinstance(entry, dict) else entry`, and an object always has `out`.

That also disposes of the `str(None)` -> `"None"` note on the qemu hook at :49 — `out` cannot be None from a manifest stone accepted.

Same conclusion for the fixture point: the suites use explicit `out` and bare strings because those are the only two forms, so the gap you flagged is not one.

Caveat worth stating: I read this at the SRCREV the recipe pins (`a620b52`). If stone later makes `out` optional this becomes live again, and the fix would be the basename fallback you described.

meta-avocado-qemu/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
jetm: Agreed that half-claiming is the worst of the three, and I have not changed it yet because the two honest options both un-ship something this PR advertises. Flagging the state plainly first: on the current head qemuarm64 still sets `claimed_by` with selection unwired, so the failure you describe is live.

Your trace is right — `env/avocado-qemuarm64.txt` has no `fdt addr` / `fdt apply` and never loads `overlays/*.dtbo`, so the QEMU-generated DTB reaches Linux untouched while the build reports success.

The options I see, and what each costs:

**A. Stop setting `claimed_by` on this platform.** One line. The CLI's undelivered check then fails the build, so a declared overlay on qemuarm64 becomes a hard error. The delivery code stays in place for when selection lands. The error message comes from the CLI and does not explain *why*, so it reads as a tooling failure rather than an unsupported-platform one.

**B. Have the hook refuse outright on qemuarm64.** Also small, and the diagnostic is ours, so it can say "delivery works, boot-time selection is not wired on this platform yet". Same practical effect as A — no overlays on qemuarm64 — but the user learns the actual reason.

**C. Wire the u-boot selection.** Makes the feature real: load the DTB, `fdt apply` each overlay from `avocado-overlays.txt`'s equivalent, in the qemuarm64 boot env. Materially larger than a review fix and it needs its own boot test, so it is a follow-up PR rather than something to fold in here.

I would take **B now and C as a follow-up**. B costs the same as A and is the difference between a user filing a tooling bug and a user reading one line and moving on. What would change my mind is if qemuarm64 overlay support is meant to land in this PR — then it is C, and A/B are both wrong because they ship a refusal for a feature that was supposed to work.

Happy to push whichever you prefer; say the word and it is a small change either way.

mobileoverlord: Settled: option A, and shipping this PR without it. The qemu hook should stop setting `claimed_by` rather than learn that qemuarm64 is unsupported — teaching a hook a machine name is the platform knowledge we do not want in the tooling, and `claimed_by` already means "I delivered this", so not setting it is the hook being accurate rather than a special case. The generic CLI undelivered check then fails the build on its own with no platform awareness added anywhere.

Not fixed here. Javier has a follow-up underway that owns it, so this merges with the gap known and named: on qemuarm64 a declared overlay is delivered, `claimed_by` is set, and nothing applies it at boot. Resolving so the merge is not blocked, not because the behaviour changed.

meta-avocado-raspberrypi/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
mobileoverlord: Your reading is right and the code needs no change. Verified independently at `a620b52` — the SRCREV this PR pins, since it bumps stone to 2.3.0 — where `FileEntry` is `#[serde(untagged)]` with `output: String`, not `Option<String>`. An object entry without `out` matches neither variant, so serde rejects the whole manifest at deserialisation and the implied-out form cannot reach this code. Both reachable shapes are already handled.

Your caveat is the right one to keep: if stone ever makes `out` optional, the basename fallback becomes live again. Resolving.

mobileoverlord (approved):
Approving. Eight of the nine review items are genuinely closed, and I checked the fixes rather than trusting the resolve checkbox — each is at the line the reply cites:

| Item | Fix |
|---|---|
| `dtc` missing `-i` | `:169` `-i "$(dirname "$src")"`, unconditional |
| `^fragment@` name grep | `:188` walks root nodes for an `__overlay__` child |
| Fixups guard wrong both ways | `:193` anchored `grep -qx 'target'` |
| `shift 2` on absent value | `:55` `shift_opt()` |
| Pi stock `.dtbo` collision | `:117` refuses by name |
| `find_boot_fat` first-match | `:65` refuses on >1 match |
| qemux86-64 accepted overlays | `:16` scoped `:avocado-qemuarm64` |

The `entry.get("out")` thread was argued rather than changed, and the argument is right — I verified `FileEntry` at `a620b52`, the SRCREV this PR pins, and it is `#[serde(untagged)]` with `output: String`, so the implied-out form cannot survive stone's deserialisation. No change needed.

**Merging with one known gap.** On qemuarm64 a declared overlay is delivered, `claimed_by` is set, and nothing applies it at boot - so the build reports success and the overlay never reaches Linux. Not fixed here; a follow-up owns it. The agreed shape is option A: the qemu hook stops setting `claimed_by`, so the generic CLI undelivered check fails the build on its own. Option B was rejected because teaching the hook that qemuarm64 is unsupported is exactly the platform knowledge we do not want in the tooling.

Worth noting the parse guard from #242 ran on this PR and passed both arches - first PR it has actually covered.

PR #246, opened 2026-07-23
meta-avocado-qemu: provision the container-dev trust-store location

jetm:
Container Dev Mode delivers its per-project CA into the device trust store at
session start (never baked). Provision the `/etc/container-dev` location in the
qemu base-files so the runtime delivery has a target; no certificate is
installed at build time.

### DIFF
--- a/meta-avocado-qemu/recipes-core/base-files/base-files_%.bbappend
+++ b/meta-avocado-qemu/recipes-core/base-files/base-files_%.bbappend
@@ -1,3 +1,7 @@
 do_install:append() {
     install -d ${D}/efi
+    # Container Dev Mode (design D8/H4): provision the trust-store LOCATION only.
+    # The per-project CA is delivered into the engine trust store at `up`, never
+    # baked into the image.
+    install -d ${D}${sysconfdir}/container-dev
 }


meta-avocado-qemu/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -1,3 +1,7 @@
 do_install:append() {
     install -d ${D}/efi
+    # Container Dev Mode (design D8/H4): provision the trust-store LOCATION only.
+    # The per-project CA is delivered into the engine trust store at `up`, never
+    # baked into the image.
+    install -d ${D}${sysconfdir}/container-dev
```
jetm, line 6: **Nothing reads `/etc/container-dev`.** The CLI delivers the CA to `/etc/docker/certs.d/<registry>/ca.crt` and the bootstrap to `/var/lib/avocado/container-dev/`, and it creates both directories itself.

- `avocado-cli/src/utils/container_dev/bootstrap.rs:240-245` builds `ca_trust_path` as `format!("/etc/docker/certs.d/{registry}/ca.crt")`.
- `avocado-cli/src/commands/container/dev.rs:736-751` (`deliver_vm_ca`) runs `mkdir -p {ca_dir} && cat > {ca_path} && chmod 0644 {ca_path}` against that parent.
- The other runtime dir is `WRITABLE_PARTITION = "/var/lib/avocado"` (`bootstrap.rs:47`) plus `BOOTSTRAP_RELATIVE_PATH = "container-dev/bootstrap.json"` (`bootstrap.rs:50`), self-created by `mkdir -p` at `dev.rs:702`.

Grepping `etc/container-dev` across `avocado-cli`, `meta-avocado`, `stone`, `avocado-mcp` and `avocadoctl` returns hits in exactly one file, and none of them is a device-side read.

So after `avocado container dev up` on a qemu guest built with this, `/etc/container-dev` is an empty 0755 directory with no writer and no reader.

Two things to decide rather than one. First, whether the intended path is `/etc/docker/certs.d/` - if so this line is aimed one directory tree over. Second, whether pre-provisioning helps at all here: the rootfs is `erofs-lz4` with `read-only-rootfs` (`avocado-image-rootfs.bb:16-18`) and `/etc` is a confext overlay mounted noexec (stated at `rubikpi3-init-services_1.0.bb:49` and `rubikpi3-thermal_1.1.0.bb:20`), so if that overlay has no writable upper on the engine guest, `deliver_vm_ca`'s `mkdir -p` fails with EROFS regardless of this directory existing. Booting the qemu image and running `touch /etc/.rwtest` and `mkdir -p /etc/docker/certs.d/x` over SSH settles that in a minute; if both succeed, the read-only concern drops and only the wrong-path point stands.

meta-avocado-qemu/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -1,3 +1,7 @@
 do_install:append() {
     install -d ${D}/efi
+    # Container Dev Mode (design D8/H4): provision the trust-store LOCATION only.
```
jetm, line 3: The only thing observing this change is a grep of this file's own source text, so the check passes for any path string containing `container-dev` and cannot catch a wrong location.

`avocado-cli/docs/container-dev/verify-vm-write-path.sh:100-104`:

```sh
if grep -q 'container-dev' "$BBAPPEND"; then
  ok "the overlay provisions the /etc/container-dev trust-store location"
```

and `$BBAPPEND` defaults to this exact file at `:36`.

Falsifier: change the line to `install -d ${D}${sysconfdir}/container-dev-typo`, or move it to `${localstatedir}/container-dev`, or delete the `install` and leave only the comment. Step 4 prints PASS in all three cases.

The script's own real assertion at step 1/4 (`:129`) tests `$GUEST_CA` = `/etc/docker/certs.d/${VM_REGISTRY}/ca.crt` - a different path that step 4 never mentions, which is the same mismatch as the finding below.

jetm (commented):
Second-reviewer pass on my own change, and this one needs a rethink rather than a tweak: I cannot find anything that reads the path it provisions. The CLI puts the CA at `/etc/docker/certs.d/<registry>/ca.crt` and the bootstrap at `/var/lib/avocado/container-dev/`, and creates both itself. Two findings inline - the path, and the fact that the only check covering this is a grep of the bbappend text, which passes for any string containing `container-dev`.

Mechanics that are fine and not worth re-litigating: permissions (`install -d` gives 0755 root:root and `deliver_vm_ca` chmods the cert itself), no `FILES:${PN}` addition needed (upstream `base-files_3.0.14.bb:161` sets `FILES:${PN} = "/"`), the layer is right (the `%` wildcard covers both qemuarm64 and qemux86-64, and the verify script points at exactly this file), and the line matches the style of the pre-existing `install -d ${D}/efi` beside it. No other layer provisions a container-dev path, so there is no peer left un-updated.

meta-avocado-qemu/recipes-core/base-files/base-files_%.bbappend
jetm: Confirmed, and the evidence turned out stronger than the finding: I'm closing this PR rather than repointing the path.

I said two things needed deciding - whether the path was wrong, and whether pre-provisioning helps at all given the read-only rootfs. The second collapses the first. `deliver_vm_ca` (`dev.rs:747`) runs:

```
mkdir -p {ca_dir} && cat > {ca_path} && chmod 0644 {ca_path}
```

`mkdir -p` creates the **whole chain**. So when `/etc` is writable the CLI needs no pre-baked parent, and when it isn't, nothing can be written there regardless of which directories the image ships. Pre-provisioning cannot help in either world - which also means "aim it at `/etc/docker/certs.d/` instead" is not a repair either, since the CLI would still have to create `<registry>` beneath it.

That makes the boot check I proposed unnecessary. I did not run it, because no outcome changes the answer.

Since this commit is the PR's entire content, there is nothing left to repurpose. Closing.

meta-avocado-qemu/recipes-core/base-files/base-files_%.bbappend
jetm: Fixed in avocado-linux/avocado-cli@2c8e854, on the `container-dev-mode` branch where the script lives - it could not ride along here, since this PR is being closed (see the thread above).

Deleted the self-referential assertion rather than repairing it. Re-ran the three mutations from the finding against the fixed step: a typo'd path, the directory moved to `${localstatedir}`, and the install line deleted leaving only a comment. All three used to print `PASS: the overlay provisions the /etc/container-dev trust-store location`; none now makes any positive claim.

Kept the first assertion, which is a negative - no cert baked - and verified it still discriminates: a fixture that installs `ca.crt` reports `FAIL: ... a CA is baked`.

The comment now says why there is no positive counterpart, so it does not get re-added: build metadata cannot tell you where a file landed at run time, and step 1/4 already asserts that against the live guest via `openssl x509 -in $GUEST_CA` over ssh.

jetm:
Closing: the change cannot do what it was written to do, at any path.

`avocado container dev up` delivers the per-project CA with `mkdir -p {ca_dir} && cat > {ca_path}` (`avocado-cli src/commands/container/dev.rs:747`), where `ca_dir` is `/etc/docker/certs.d/<registry>` (`bootstrap.rs:245`). Because `mkdir -p` creates the whole chain, a directory baked into the image is redundant when `/etc` is writable and useless when it is not - so `/etc/container-dev` is not merely the wrong path, there is nothing for the image to provision at all. Grepping `container-dev`, `meta-avocado`, `stone`, `avocado-mcp` and `avocadoctl` finds no device-side reader for it either.

The verification that reported this as working was checking its own source text - a grep of this bbappend for the literal `container-dev`. That is fixed separately in avocado-linux/avocado-cli@2c8e854, since the script lives in that repo; the real assertion (step 1/4, `openssl x509` against `$GUEST_CA` on the live guest) was already correct and is untouched.

No follow-up PR here - there is no residual change to make in meta-avocado.

PR #247, opened 2026-07-23
Default boards to minimal, opt into the full feature set

jetm:
## Problem

Every non-qemu board included `kas/feature/complete.yml` (the umbrella of all
feature groups except `ai`) by default, so a plain `kas build <machine>`
produced the maximal image. That default OOM-hung a build box on chromium and
forced a subtractive `AVOCADO_FEATURE_GROUPS:remove` hack just to get a minimal
boot.

## Solution

Remove the umbrella include from the 12 board files that carried it (9 vendor, 3
machine) so a board defaults to base content. Full-featured becomes an explicit
`:kas/feature/complete.yml` append. CI now builds each labeled board in both a
minimal and a full variant so the new minimal default is exercised per board,
and the docs are updated to match.

## Key changes

- Drop the `complete.yml` umbrella include from 9 vendor + 3 machine files;
  board-specific fragments (virtualization, ai, multi-kernel) are kept.
- Full-featured is now an explicit `:kas/feature/complete.yml` append; add a
  `--complete` flag to `scripts/build-all-machines.sh` (kas needs the fragment
  glued to the config arg, which `$ARGS` forwarding cannot do).
- CI (`pr-build-labeled.yml`) builds a `minimal` + `full` variant matrix per
  labeled board for durable per-board coverage of the minimal default.
- Rewrite `docs/feature-groups.md` and the `complete.yml` header to
  minimal-by-default and the explicit opt-in.

## Reviewer notes

Targets `scarthgap` (feature groups already merged there, #229). **BREAKING**
for anyone relying on full-by-default: append `:kas/feature/complete.yml` to
restore. `--target avocado-complete` is a BitBake target, unrelated to the
`kas/feature/complete.yml` fragment.

### Tested (real `bakar build` on raspberrypi5, `--target avocado-complete`)

| Board yaml | Variant | Result |
|---|---|---|
| `raspberrypi5.yml` | bare-base (default) | PASS - SDK packaged, `os-bundle.aos` finalized, feed has zero chromium/webkit/qt |
| `raspberrypi5.yml:...:python.yml` | dev-append (`system-base`+`networking`+`python`) | PASS - NetworkManager + python composed, `os-bundle.aos` finalized, still no heavy GUI groups |
| `raspberrypi5.yml:complete.yml` | full | resolves via CI full variant |
| `orangepi-5-plus.yml` (rockchip) | dependency graph | resolves |
| `reterminal-dm.yml` | dependency graph | resolves |

### Still needs testing (auto-covered by the new CI minimal+full matrix)

Only `raspberrypi5.yml` got real image builds. Every other board yaml that lost
the umbrella - or inherits it via one of the 9 edited vendor files - is
exercised by the CI `minimal`+`full` matrix on its next `build-<machine>`
labeled PR. That is the 12 edited files and every machine that includes them:

- **Directly edited machine yamls:** `grinn-astra-1680-sbc.yml`,
  `imx8mp-evk.yml`, `imx93-evk.yml`.
- **Edited vendor yamls** (`compulab`, `intel`, `nvidia`, `nxp-frdm`,
  `raspberrypi`, `renesas`, `rockchip`, `rubikpi`, `synaptics`) fan out to their
  consuming machines: the jetson/recomputer/icam boards (nvidia), the
  raspberrypi family incl. fr202/reterminals (raspberrypi), intel-x86-64-v2/v3/v4
  (intel), the imx*-frdm and ucm/compulab boards (nxp-frdm/compulab),
  rzv2n-sr-som (renesas), orangepi-5-plus (rockchip), rubikpi3 (rubikpi), and the
  synaptics/deepx boards (synaptics).

Watch the BSP-packagegroup boards (reterminal, synaptics, tegra, renesas) for a
base packagegroup RDEPENDing on removed feature content - that is the one way a
board could regress. grinn-astra is deepx (keeps `ai.yml`); a `dx-*` failure
there is a pre-existing deepx issue, not this change.

### Unrelated pre-existing issue surfaced during validation

The scarthgap SDK/target `libyaml`/`python3-pyyaml` `do_package_qa` "no shlib
provider" bug hit both the nativesdk and target `python3-pyyaml` on warm sstate,
cleared each time by `bitbake -c cleansstate libyaml python3-pyyaml`. It is
independent of this change (`python3-pyyaml` comes from the unchanged `python`
feature group) and is being filed separately.

### DIFF
--- a/.github/workflows/pr-build-labeled.yml
+++ b/.github/workflows/pr-build-labeled.yml
@@ -42,10 +42,11 @@ jobs:
       fail-fast: false
       matrix:
         machine: ${{ fromJSON(needs.set-matrix.outputs.matrix) }}
+        variant: [minimal, full]
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
-      - name: build ${{ matrix.machine }}
+      - name: build ${{ matrix.machine }} (${{ matrix.variant }})
         shell: bash
         env:
           BB_GIT_SHALLOW: "1"
@@ -57,11 +58,37 @@ jobs:
           SOURCE_MIRROR_URL: "file:///avocado/dl"
           SSTATE_MIRRORS: "file:///avocado/sstate/PATH;downloadfilename=PATH"
         run: |
-          echo "TARGET: avocado-distro"
+          echo "TARGET: avocado-complete  VARIANT: ${{ matrix.variant }}"
           echo "SDKMACHINE: x86_64"
-          . scripts/init-build kas/machine/${{ matrix.machine }}.yml
-          # Append kas/feature/complete.yml so labeled PR builds ship the full
-          # image (boards default to minimal now). NOTE: --target avocado-complete
-          # is a BitBake target, unrelated to the kas/feature/complete.yml fragment.
+          # The variant MUST be in the build directory name. init-build derives
+          # "build-<machine>" from the basename alone, so without this both cells
+          # for one machine share a build dir, symlink, TMPDIR and bitbake.lock -
+          # and DL_DIR/SSTATE_DIR too, since kas/ci/mirrors.yml puts them under
+          # ${TMPDIR}, inside the colliding directory. Every matrix cell used to
+          # have a distinct directory by construction; adding the variant
+          # dimension is what removed that, so restore it explicitly.
+          . scripts/init-build kas/machine/${{ matrix.machine }}.yml \
+            "build-${{ matrix.machine }}-${{ matrix.variant }}"
+          # Build each labeled machine twice: minimal (board default, no umbrella)
+          # and full (umbrella appended). The minimal cell is durable per-board A4
+          # coverage - a base packagegroup RDEPENDing on removed feature content
+          # fails here.
+          #
+          # The full cell preserves the pre-existing umbrella coverage for the
+          # boards that already had it, but for alif-e8-devkit, stm32mp25-dk,
+          # imx8mp-var-dart, qemuarm64 and qemux86-64 it is NEW coverage that has
+          # never run: none of the five pulled complete.yml at base, directly or
+          # transitively. packagegroup-avocado-feature-system-base and
+          # -networking do not gate on DISTRO_FEATURES, so those boards will now
+          # build cockpit, nodejs, redis, curl, networkmanager, bluez5 and
+          # wireguard-tools for the first time. A failure in the full cell for
+          # one of them is a first run, not a regression.
+          #
+          # NOTE: --target avocado-complete is a BitBake target, unrelated to the
+          # kas/feature/complete.yml fragment.
+          COMPLETE=""
+          if [ "${{ matrix.variant }}" = "full" ]; then
+            COMPLETE=":meta-avocado/kas/feature/complete.yml"
+          fi
           # Use CI overlay to enable mirrors and avoid directory locking.
-          kas build meta-avocado/kas/machine/${{ matrix.machine }}.yml:meta-avocado/kas/feature/complete.yml:meta-avocado/kas/ci/mirrors.yml --target avocado-complete
+          kas build "meta-avocado/kas/machine/${{ matrix.machine }}.yml${COMPLETE}:meta-avocado/kas/ci/mirrors.yml" --target avocado-complete
--- a/.github/workflows/pr-build-labeled.yml
+++ b/.github/workflows/pr-build-labeled.yml
@@ -60,5 +60,8 @@ jobs:
           echo "TARGET: avocado-distro"
           echo "SDKMACHINE: x86_64"
           . scripts/init-build kas/machine/${{ matrix.machine }}.yml
-          # Use CI overlay to enable mirrors and avoid directory locking
-          kas build meta-avocado/kas/machine/${{ matrix.machine }}.yml:meta-avocado/kas/ci/mirrors.yml --target avocado-complete
+          # Append kas/feature/complete.yml so labeled PR builds ship the full
+          # image (boards default to minimal now). NOTE: --target avocado-complete
+          # is a BitBake target, unrelated to the kas/feature/complete.yml fragment.
+          # Use CI overlay to enable mirrors and avoid directory locking.
+          kas build meta-avocado/kas/machine/${{ matrix.machine }}.yml:meta-avocado/kas/feature/complete.yml:meta-avocado/kas/ci/mirrors.yml --target avocado-complete
--- a/kas/vendor/intel.yml
+++ b/kas/vendor/intel.yml
@@ -1,8 +1,5 @@
 header:
   version: 16
-  includes:
-  - repo: meta-avocado
-    file: kas/feature/complete.yml
 
 repos:
   meta-intel:
--- a/kas/vendor/nvidia.yml
+++ b/kas/vendor/nvidia.yml
@@ -1,8 +1,5 @@
 header:
   version: 16
-  includes:
-  - repo: meta-avocado
-    file: kas/feature/complete.yml
 
 repos:
   meta-tegra:
--- a/docs/feature-groups.md
+++ b/docs/feature-groups.md
@@ -28,11 +28,13 @@ kas build kas/machine/<machine>.yml:kas/feature/<group>.yml:kas/feature/<group>.
 - At parse time `packagegroup-avocado-extra` expands every token into a
   `packagegroup-avocado-feature-<token>` runtime dependency.
 
-With no feature fragment, `AVOCADO_FEATURE_GROUPS` is empty and the image gets
-the always-on base content only.
+Boards are **minimal by default**: with no feature fragment appended,
+`AVOCADO_FEATURE_GROUPS` is empty and the image gets the always-on base content
+only.
 
 `kas/feature/complete.yml` is an umbrella that includes every group fragment
-except `ai`. Client vendor and machine files include it to ship the full set.
+except `ai`. No board file includes it; append it explicitly to ship the full
+set (see §3).
 
 ## 2. Feature groups
 
@@ -59,7 +61,7 @@ content builds against, not image packages directly): `clang.yml`,
 
 ### Bare minimal (base content only)
 
-A machine with no feature fragment. The smallest bring-up image.
+Any machine with no feature fragment. This is now the default for every board.
 
 ```bash
 kas build kas/machine/qemux86-64.yml
@@ -67,10 +69,12 @@ kas build kas/machine/qemux86-64.yml
 
 ### Minimal plus a few groups
 
-Stack only the groups you want. Always-on groups need no vendor layer:
+Stack only the groups you want. Always-on groups need no vendor layer. A
+workflow that needs a booted, reachable board (networking, ssh, python) composes
+those essentials explicitly:
 
 ```bash
-kas build kas/machine/qemuarm64.yml:kas/feature/system-base.yml:kas/feature/networking.yml:kas/feature/multimedia.yml
+kas build kas/machine/raspberrypi5.yml:kas/feature/system-base.yml:kas/feature/networking.yml:kas/feature/python.yml
 ```
 
 ### Adding graphics / Qt / browsers
@@ -84,13 +88,14 @@ kas build kas/machine/<gpu-machine>.yml:kas/feature/graphics.yml:kas/feature/qt.
 
 ### Full featured
 
-Pull every group through the umbrella (this is what client targets do):
+Every board is minimal by default, so ship the full set by appending the
+umbrella explicitly:
 
 ```bash
-kas build kas/machine/imx8mp-evk.yml
+kas build kas/machine/imx8mp-evk.yml:kas/feature/complete.yml
 ```
 
-To compose it onto a bare machine manually:
+The same append works on any machine:
 
 ```bash
 kas build kas/machine/qemuarm64.yml:kas/feature/complete.yml
@@ -98,15 +103,16 @@ kas build kas/machine/qemuarm64.yml:kas/feature/complete.yml
 
 ### DeepX (NPU) boards
 
-`ai` is excluded from `complete.yml` because it is `MACHINE_FEATURES`-gated. A
-deepx board stacks both (the grinn-astra machine file already includes
-`complete.yml` and `ai.yml`):
+`ai` is excluded from `complete.yml` because it is `MACHINE_FEATURES`-gated.
+Deepx machine files keep their own `ai.yml` include (deepx is machine-specific),
+but no board includes the umbrella. grinn-astra defaults to base + ai; append
+the umbrella for the full set:
 
 ```bash
-kas build kas/machine/grinn-astra-1680-sbc.yml
+kas build kas/machine/grinn-astra-1680-sbc.yml:kas/feature/complete.yml
 ```
 
-To compose it manually onto another deepx machine:
+On a deepx machine whose file does not already include `ai.yml`, append both:
 
 ```bash
 kas build kas/machine/<deepx-machine>.yml:kas/feature/complete.yml:kas/feature/ai.yml
@@ -122,6 +128,6 @@ kas build kas/machine/<deepx-machine>.yml:kas/feature/complete.yml:kas/feature/a
 - The SDK mirrors the target: `avocado-pkg-sdk-extra` pulls the Qt5 nativesdk
   toolchain only when the `qt` token is present, so an SDK build that opts out
   of `qt` does not require meta-qt5.
-- When adding a new client machine that should ship the full set, include
-  `kas/feature/complete.yml` (and `ai.yml` if it is a deepx board) rather than
-  listing groups individually.
+- Board files do not include the umbrella; a board is minimal by default. To
+  ship the full set, append `:kas/feature/complete.yml` at build time (and
+  `:kas/feature/ai.yml` on a deepx board) rather than adding an include.
--- a/kas/feature/complete.yml
+++ b/kas/feature/complete.yml
@@ -3,9 +3,10 @@ header:
 
 # Umbrella aggregator: includes every per-group kas/feature/<group>.yml
 # fragment so that including this one file pulls every vendor layer and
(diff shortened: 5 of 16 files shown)


kas/vendor/intel.yml
```diff
@@ -1,8 +1,6 @@
 header:
   version: 16
   includes:
```
jetm: **This breaks the build for three machines: the emptied `includes:` key is YAML null and kas rejects it.**

`complete.yml` was this file's only `header.includes` entry. Removing the list item but leaving the bare `includes:` key makes YAML parse it as `None`, which fails kas's schema validation - and validation runs on *included* files, not just the entry point, so it aborts at config-parse time before any layer fetch.

Reproduced against kas 5.4 with a minimal pair:

```
$ kas dump machine.yml
ERROR - Config file validation Error:
None is not of type 'array'
ERROR - Error(s) occured while validating the config file: .../vendor.yml
```

Deleting the `includes:` line makes the same pair dump cleanly.

This file takes down `intel-x86-64-v2`, `-v3` and `-v4`; `kas/vendor/nvidia.yml` has the identical shape and takes down six more. Both the new `minimal` and `full` CI cells fail for all nine.

Fix is deleting the now-empty `includes:` line - only `version` is required. The other seven vendor files (`compulab`, `nxp-frdm`, `raspberrypi`, `renesas`, `rockchip`, `rubikpi`, `synaptics`) each had a second entry and are fine.

kas/vendor/nvidia.yml
```diff
@@ -1,8 +1,6 @@
 header:
   version: 16
   includes:
```
jetm: Same defect as `kas/vendor/intel.yml:3` - `complete.yml` was the only `includes` entry, so the bare key now parses as YAML null and kas fails schema validation on the included file.

Machines this takes down: `jetson-agx-orin-devkit`, `jetson-orin-nano-devkit`, `jetson-orin-nx`, `recomputer-j301x`, `recomputer-j401x`, `icam-540`.

Delete the empty `includes:` line.

.github/workflows/pr-build-labeled.yml
```diff
@@ -57,8 +58,18 @@ jobs:
           SOURCE_MIRROR_URL: "file:///avocado/dl"
           SSTATE_MIRRORS: "file:///avocado/sstate/PATH;downloadfilename=PATH"
         run: |
-          echo "TARGET: avocado-distro"
+          echo "TARGET: avocado-complete  VARIANT: ${{ matrix.variant }}"
           echo "SDKMACHINE: x86_64"
           . scripts/init-build kas/machine/${{ matrix.machine }}.yml
```
jetm: The new `variant` dimension is not in the build directory path, so both cells for one machine target the same directory.

`. scripts/init-build kas/machine/${{ matrix.machine }}.yml` derives `BUILD_DIR="$(pwd)/build-${BASENAME}"` from the machine basename alone - so `minimal` and `full` share a build dir, symlink, `TMPDIR`, and `bitbake.lock`. With `fail-fast: false`, no `max-parallel` and no `concurrency:` group, both cells are eligible immediately.

If two `molcajete` runner processes share a workspace root, the second bitbake hits `Another instance of bitbake is running` - or worse, the two interleave in one `tmp/` and the `minimal` cell's sstate-reused rootfs picks up the `full` cell's feature content, destroying exactly the A4 signal the minimal cell exists to provide.

Marked as a should-check rather than a certainty because runner topology is not in-tree: with a single runner process the jobs serialize, and `actions/checkout@v4` with `clean: true` then `git clean -ffdx`s the gitignored `build-*/` between them - safe, but it throws away the first cell's sstate. Either way, this removes the property that previously made it safe by construction, since every matrix cell used to have a distinct directory.

`kas/ci/mirrors.yml`'s "avoid locking issues when building in parallel" comment does not cover this - `DL_DIR`/`SSTATE_DIR` are set under `${TMPDIR}`, which is *inside* the colliding directory. `init-build` already accepts an explicit build-dir second argument, so it is a one-argument fix.

.github/workflows/pr-build-labeled.yml
```diff
@@ -57,8 +58,18 @@ jobs:
...
-          kas build meta-avocado/kas/machine/${{ matrix.machine }}.yml:meta-avocado/kas/ci/mirrors.yml --target avocado-complete
+          # Build each labeled machine twice: minimal (board default, no umbrella)
+          # and full (umbrella appended). The minimal cell is durable per-board A4
+          # coverage - a base packagegroup RDEPENDing on removed feature content
+          # fails here; the full cell keeps full-image coverage. NOTE:
+          # --target avocado-complete is a BitBake target, unrelated to the
+          # kas/feature/complete.yml fragment.
+          COMPLETE=""
+          if [ "${{ matrix.variant }}" = "full" ]; then
+            COMPLETE=":meta-avocado/kas/feature/complete.yml"
```
jetm, line 91: The comment says the `full` cell "keeps full-image coverage", but for five machines it is new coverage that has never run.

`alif-e8-devkit`, `stm32mp25-dk`, `imx8mp-var-dart`, `qemuarm64` and `qemux86-64` never pulled the umbrella at base, so the `full` cell is not preserving anything for them.

Concretely for `alif-e8-devkit`: its `local_conf_header` block is empty - no `DISTRO_FEATURES_EXTRA`, so no opengl/wayland/virtualization. The graphics, qt, browsers and containers packagegroups self-gate on those features and resolve empty (I checked). But `packagegroup-avocado-feature-system-base` and `-networking` do **not** gate, so the full cell would newly try to build cockpit, nodejs, redis, curl, networkmanager, bluez5 and wireguard-tools on a board that has never built any of it. Same shape for `stm32mp25-dk`.

It may well build fine. The point is narrower: the comment presents this cell as preserving existing coverage when for these five boards it is adding untried coverage, so a failure there will read as a regression rather than as a first run.

jetm (commented):
Second-reviewer pass on my own change, cold, and it found a hard blocker: **`kas/vendor/intel.yml` and `kas/vendor/nvidia.yml` are left with a bare `includes:` key, which YAML parses as null and kas rejects at config-parse time.** Nine machines fail before any layer fetch, in both the new `minimal` and `full` CI cells. Reproduced against kas 5.4. Details inline; the fix is deleting one line in each file.

First, what I verified as correct, so it does not get re-litigated:

- **The removal sweep is complete.** Resolving the `header.includes` graph across all 30 `kas/machine/*.yml`: 25 machines transitively pulled `complete.yml` at base, 0 do at head. No board or vendor file was missed, and no sibling within a family (imx*, jetson*, raspberrypi*, intel-x86-64-*) was treated inconsistently.
- **No hard build failure from a base packagegroup.** `packagegroup-avocado-rootfs.bb` RDEPENDS only on `os-release base-files base-passwd netbase avocadoctl avocado-users btrfs-tools` plus the `VIRTUAL-RUNTIME_*` set; `-initramfs`, `-container` and `-sdk*` touch nothing feature-gated. `packagegroup-avocado-x86-64-extra` (tpm2-tools) and `-tegra-extra` (nvidia-docker) survive because those machines include `tpm.yml`/`virtualization.yml` directly rather than through the umbrella.
- All three umbrella path spellings are right in their own context, `docs/feature-groups.md` is internally consistent (the section pointer resolves, the grinn-astra claim matches HEAD, and the three fragments in the raspberrypi5 example all exist), there is no `local_conf_header` key collision from moving the umbrella to a command-line append, and `--target avocado-complete` was already the real target - the old `echo "TARGET: avocado-distro"` was simply wrong.

Four findings inline. Four more have no anchorable line because they live in files this diff does not touch - which is itself the pattern worth noticing: a default flip lands in the kas files, and the fallout is in the docs and scripts that assumed the old default.

**`docs/feed-staging-runbook.md:105,108-111,121,124-126,331`** - the runbook's all-targets produce step is `bash meta-avocado/scripts/build-all-machines.sh` with no `--complete`, and its verify step is `find <machine-build>/tmp/deploy/rpm -name 'zeromq-*.rpm'`. `zeromq` lives only in `packagegroup-avocado-feature-networking.bb:45`, so post-PR that find returns nothing for every machine, where previously imx8mp-evk, imx93-evk, grinn-astra and the nine vendor-backed families all produced it. Lines 124-126 then hand the operator the wrong diagnosis: "the shared packagegroup did not reach that target - record a corpus case and switch that target to an explicit dependency." The real cause is a missing `:kas/feature/complete.yml`. The "Useful flags (passed through to kas)" list also omits `--complete`, which is not actually passed through - the script's `case` consumes it - and the signature row at 331 still reads `build-all-machines.sh [--clean] [--sdkmachine=<arch>]`. Concrete cost is a 27-machine farm sweep run bare, then re-run.

**`scripts/feed-validation-lib.sh:105`** - this is a third build entry point and it got neither the `--complete` switch nor the CI `variant`. `fvl_build` runs `$AVOCADO_LOCAL_BUILD_CMD build "$kas_machine" --target avocado-complete` with the bare machine yml, and `scripts/feed-validation-cases` carries one case per feature group (`system-base|vim`, `networking|networkmanager`, `python|python3-pip`, `graphics|weston`, `qt|qtbase`, `browsers|chromium-ozone-wayland`, `containers|podman`, plus `zeromq`). Post-PR there is no machine argument for which `run-feed-validation.sh -m <machine>` can produce a feed containing those packages. `run-feed-validation.sh -m imx8mp-evk` worked before this PR and does not now. The `-m qemuarm64` default was already minimal at base, so that part is pre-existing - the loss of *any* working machine is what this introduces.

**`docs/adding-a-machine-target.md:175-178, 210, 672`** - the new-machine guide still mandates meta-virtualization on a rationale that is now false: it says "meta-virtualization is REQUIRED -- packagegroup-avocado-extra in the base meta-avocado layer RDEPENDs on docker", with symptom `Nothing RPROVIDES 'docker'`. At HEAD, `packagegroup-avocado-extra.bb`'s static RDEPENDS has no docker; docker is in `packagegroup-avocado-feature-containers.bb:13`, reachable only via the `containers` token, which no board now sets. So someone onboarding a board adds `virtualization.yml` and `DISTRO_FEATURES_EXTRA:append = " opengl wayland seccomp virtualization"` to dodge a parse error that can no longer occur, paying a meta-virtualization fetch and a docker build for nothing. Worse in the other direction: this file never mentions `kas/feature/complete.yml`, and its step 13 is a bare `kas build distro/kas/machine/<board>.yml` - so the board they hand off ships base-only with no indication that is a choice. This is where the guidance removed from `docs/feature-groups.md` ("When adding a new client machine that should ship the full set, include `kas/feature/complete.yml`") should have landed.

**`docs/local-package-feed.md:33`** - the running example is `./scripts/dev-repo.sh 2024 imx8mp-evk` after building `build-imx8mp-evk/`, and that machine lost its `complete.yml` include here. A developer follows the doc as written and gets a browsable feed missing Qt, chromium, the python3 set, openssh/NetworkManager and zeromq - a feed whose stated purpose ("install from the feed") is unserviceable for essentially anything they would reach for. Nothing hints the build command needs an appended fragment now.

**One design point worth taking seriously, `packagegroup-avocado-extra.bb:38`.** Every other opt-in mistake fails loudly - I checked: a typo'd fragment path dies with a kas config-not-found error, a space instead of a colon reaches bitbake as `Nothing PROVIDES 'kas/feature/complete.yml'`, and a typo'd token like `netwoking` becomes a literal RDEPENDS on `packagegroup-avocado-feature-netwoking` and fails `Nothing RPROVIDES`. Simply *omitting* the umbrella is the one exception, because `or ''` expands an empty variable to the empty string with no diagnostic. So a release engineer runs yesterday's command, gets a green build indistinguishable from the full one, and publishes a feed missing about fourteen groups - no log line, no QA check, no `AVOCADO_FEATURE_GROUPS` echo anywhere in the output. The only signal is a package absence nobody looks for until a device tries to install one. A one-line `bb.note` or `bb.warn` when the variable is empty is the cheapest mitigation for the whole opt-in design, and I would want it before this ships.

Two notes, not findings: `scripts/build-all-machines.sh:33`'s `ARGS="${PASSTHRU_ARGS[*]}"` is now vestigial - its only remaining consumer is the `echo "Arguments passed: $ARGS"` on line 40. And the repo's `shfmt -i 2 -ci -bn` pre-commit hook rewrites that entire script (it uses 4-space indentation throughout); the non-conformance is pre-existing, but touching the file is what makes the hook fire, so expect a formatting churn commit. `shellcheck` is clean.

kas/vendor/intel.yml
jetm: Fixed in 9480f30 - the empty `includes:` line is deleted, so `header` carries only `version`.

Reproduced your repro before fixing rather than taking it on trust, against kas 5.4:

```
$ kas dump machine.yml
ERROR - Config file validation Error:
None is not of type 'array'
ERROR - Error(s) occured while validating the config file: .../vendor.yml
```

and the same pair dumps cleanly with the line removed. Confirmed the mechanism too: `yaml.safe_load` gives `header.includes = None`, and the failure is reported against the *included* file, so it aborts at config-parse time.

Swept every kas config for the same shape - `intel.yml` and `nvidia.yml` were the only two left holding an empty `includes`, matching your read that the other seven vendor files each kept a second entry.

Landed in the commit that emptied the key (`kas: default boards to minimal...`) rather than as a fixup on top, so no commit in the series is broken in isolation.

kas/vendor/nvidia.yml
jetm: Fixed in the same commit, 9480f30 - same one-line deletion.

Confirmed the machine list you gave: `jetson-agx-orin-devkit`, `jetson-orin-nano-devkit`, `jetson-orin-nx`, `recomputer-j301x`, `recomputer-j401x`, `icam-540` all reach `complete.yml` only through this file, so all six were down in both cells. Nine machines across the two vendor files.

.github/workflows/pr-build-labeled.yml
jetm: Fixed in 725ae64. The build dir is now passed explicitly:

```
. scripts/init-build kas/machine/${{ matrix.machine }}.yml \
  "build-${{ matrix.machine }}-${{ matrix.variant }}"
```

Verified `init-build` honours the second argument - sourcing it twice with `build-m-minimal` and `build-m-full` lands in two distinct directories.

Took this as a fix rather than a should-check. Whichever way the runner topology falls, both outcomes you named are bad: a lock collision, or serialized cells where the checkout wipes the previous cell's sstate. And the property that made it safe before was structural - every matrix cell had a distinct directory - so restoring it is cheaper than depending on out-of-tree topology.

Your note about `kas/ci/mirrors.yml` is in the commit body: `DL_DIR`/`SSTATE_DIR` sit under `${TMPDIR}`, which is inside the colliding directory, so that comment does not cover this case.

.github/workflows/pr-build-labeled.yml
jetm: Fixed in 725ae64 - the comment now says the full cell is new coverage for those boards, not preserved coverage, and names them.

Verified the list independently by resolving the include chain transitively at the merge-base, since 27 machines lack `complete.yml` in their own machine yml and most of them pull it through a vendor file. Following `header.includes` through gives exactly five with no umbrella at base: `alif-e8-devkit`, `imx8mp-var-dart`, `qemuarm64`, `qemux86-64`, `stm32mp25-dk`. Your five.

The comment also records why it may still be fine and why it might not - the graphics/qt/browsers/containers packagegroups self-gate on `DISTRO_FEATURES`, but `packagegroup-avocado-feature-system-base` and `-networking` do not - so a failure there reads as a first run rather than a regression, which was your point.

mobileoverlord (approved):
Verified against `scarthgap` tip: exactly 12 `kas/feature/complete.yml` includes exist on base (9 vendor + 3 machine) and this PR removes exactly those 12 — no stragglers, and none of the 8 commits that landed on base since the fork point added a new one.

Checked the regression paths:

- **No orphaned layer.** `kas/feature/virtualization.yml` pulls meta-virtualization at the same repo/commit as `containers.yml`, so the boards that keep the virtualization include still get the layer without the umbrella. The jetson/recomputer/icam machine files include `virtualization.yml` directly, so dropping it from `nvidia.yml` (which had no other include) leaves nothing dangling.
- **Feature-layer references outside the fragments are inert when the layer is absent.** `rubikpi.yml`'s `PACKAGECONFIG:remove:pn-qtbase`/`pn-qtwayland`, `avocado-jetson.inc`'s `pn-wpewebkit`/`pn-chromium-ozone-wayland` overrides, and the reterminal/fr202 `pn-qtbase` appends are all `pn-`-scoped, so they no-op rather than error. meta-avocado-distro routes qt5/webkit/wayland/aws/ros content through `BBFILES_DYNAMIC` collections, which is exactly the absent-layer-tolerant design #229 put in.
- **CI build-dir collision is correctly handled.** `init-build` takes an optional second positional arg for the build dir and the workflow uses it, so the two variant cells no longer share a build dir, `meta-avocado` symlink, TMPDIR, `bitbake.lock`, or the `${TMPDIR}`-rooted DL_DIR/SSTATE_DIR from `kas/ci/mirrors.yml`. The `TARGET:` echo also stops claiming `avocado-distro` when the build is `--target avocado-complete`.
- **build-all-machines.sh quoting is a strict improvement.** `"${PASSTHRU_ARGS[@]}"` preserves multi-word args that the old unquoted `$ARGS` re-split; `ARGS` is still used for the summary echo, so it isn't dead.
- Only `pr-build-labeled.yml` invokes kas in CI, so no release path silently flips to minimal. `docs/feature-groups.md` is the canonical doc and is updated to match.

Two non-blocking notes:

1. The workflow change itself has never executed — this PR carries no `build-*` label, so only CodeQL ran. First labeled PR after merge is where the minimal+full matrix gets its first exercise.
2. The matrix now doubles concurrent jobs per labeled board on `molcajete` with no `max-parallel`. Given the OOM that motivated this change, worth watching the first multi-board labeled run.

Separately (pre-existing, from #229, not this PR): `docs/adding-a-machine-target.md` still states meta-virtualization is required because `packagegroup-avocado-extra` RDEPENDs on `docker` — `docker` now lives in the `containers` feature group and is no longer in that packagegroup's base RDEPENDS.

PR #249, opened 2026-07-28
kas/vendor/oe: bump bitbake and openembedded-core

hiagofranco:
Current "kas build" fails with:
    ERROR - Branch "2.8-avocado" in repository "bitbake" does not
    contain commit "82cf21bc6d1a2941c2dd292ecea327031a247a8b"

Update both repos at the current branch heads.

- bitbake 82cf21bc -> fc13d719 (2.8-avocado): keeps the fetch2/git LFS subdir-.gitattributes fix, now on top of 10 more upstream 2.8 backports - wget HTTP 308 and auth-on-redirect handling, hashserv unihash validation, fetch2 striplevel and deb/ipk data member validation, RPM unpack with --no-absolute-filenames, and a varflag exclusion fix in data.py.

- openembedded-core e8e93ce1 -> 06b1c4f6 (scarthgap-avocado): keeps the staging identical-symlink and sstatesig nativesdk manifest fallback patches, rebased over 176 upstream scarthgap commits. Mostly CVE fixes (expat, python3, binutils, glib-2.0, vim, util-linux, gzip, bzip2, socat, libcap, python3-urllib3), plus the wic nvme fstab fix and routine upgrades (tzdata 2026c, ca-certificates 20260601, wireless-regdb 2026.05.30).

### DIFF
--- a/kas/vendor/oe.yml
+++ b/kas/vendor/oe.yml
@@ -16,7 +16,7 @@ repos:
     url: https://github.com/avocado-linux/vendor-bitbake
     path: bitbake
     branch: "2.8-avocado"
-    commit: 82cf21bc6d1a2941c2dd292ecea327031a247a8b
+    commit: fc13d7197745017a5c8c274cde82b62a9cc796ce
     layers:
       .: disabled
 
@@ -24,7 +24,7 @@ repos:
     url: https://github.com/avocado-linux/vendor-openembedded-core
     path: openembedded-core
     branch: scarthgap-avocado
-    commit: e8e93ce1a7a952caf0859f1952eaa9fcb08856f6
+    commit: 06b1c4f636f21d29d0786c79bc325000aa80547e
     layers:
       meta:
 


hiagofranco:
@jetm: I noticed my kas build was failing today, so I decided to send the PR. Let me know if something else is needed, thanks.

jetm:
@hfranco Looking at the review. Please, next time assign it to my GH @jetm username, as I only look at the reviews that are assigned to me. Thanks

kas/vendor/oe.yml
```diff
@@ -16,15 +16,15 @@ repos:
...
+    commit: fc13d7197745017a5c8c274cde82b62a9cc796ce
     layers:
       .: disabled
 
   openembedded-core:
     url: https://github.com/avocado-linux/vendor-openembedded-core
     path: openembedded-core
     branch: scarthgap-avocado
-    commit: e8e93ce1a7a952caf0859f1952eaa9fcb08856f6
+    commit: 06b1c4f636f21d29d0786c79bc325000aa80547e
```
jetm, line 27: **This pin is a repair, not a bump - the old one was orphaned**

This is why your build broke, and it is worth recording because the failure mode does not point at the cause.

`vendor-openembedded-core:scarthgap-avocado` was **force-pushed**. My fetch reported it:

```
+ e8e93ce1a7a...06b1c4f636f scarthgap-avocado -> origin/scarthgap-avocado  (forced update)
```

And the old pin is no longer on the branch:

```
git merge-base --is-ancestor e8e93ce1a7a origin/scarthgap-avocado   # exit 1
git merge-base e8e93ce1a7a origin/scarthgap-avocado                 # ece80784b493
```

So kas could not check out the pinned commit on the named branch. Nothing was wrong with your tree - the pin was dangling.

**The two avocado carries survived the rebase.** I checked this rather than assuming, because a rewrite is exactly where a downstream carry gets dropped:

```
e8e93ce1a7a  patch-id=b18ba864a9742ac2  staging: Allow identical symlinks...
06b1c4f636f  patch-id=b18ba864a9742ac2  staging: Allow identical symlinks...
7e0e1097209  patch-id=66ce012cde6d1ee7  sstatesig: Add fallback manifest lookup...
1e0500f5d7e  patch-id=66ce012cde6d1ee7  sstatesig: Add fallback manifest lookup...
```

Identical patch-ids, and `git diff e8e93ce1a7a 06b1c4f636f -- meta/lib/oe/sstatesig.py meta/classes-global/staging.bbclass` is empty. Both carries were re-applied on top of the rebased branch, so `git log origin/scarthgap-avocado..e8e93ce1a7a` listing them as "dropped" is a hash artifact, not content loss.

`06b1c4f6` is the branch tip and brings 178 commits, overwhelmingly CVE fixes (glib-2.0, vim, expat x2, python3 x3, binutils x3, socat, bzip2, libcap, util-linux, gzip, openssh, python3-urllib3) plus `wic: Fix updating fstab for nvme devices`. Good to have.

kas/vendor/oe.yml
```diff
@@ -16,15 +16,15 @@ repos:
     url: https://github.com/avocado-linux/vendor-bitbake
     path: bitbake
     branch: "2.8-avocado"
-    commit: 82cf21bc6d1a2941c2dd292ecea327031a247a8b
+    commit: fc13d7197745017a5c8c274cde82b62a9cc796ce
```
jetm, line 19: **Clean fast-forward, and the reason #241 has to rebase onto it**

Verified: `fc13d719` is an ancestor of `origin/2.8-avocado`, and `82cf21bc..fc13d719` is 11 commits with no rewrite involved.

```
fc13d7197 fetch2/git: detect LFS in subdir .gitattributes and propagate lfs/ store
0880963fe data: fix issue with varflag exclusion
2b0f7fb5f tests/fetch: cover checkstatus redirect auth handling
348edecf9 fetch2/wget: limit auth on checkstatus redirects
5ca465fc4 fetch2/wget: handle HTTP 308 Permanent Redirect
a4daa1431 hashserv: validate unihash values
d0bc6eafb hashserv/tests: use valid 64-character unihashes
37beb06ba fetch2: Unpack RPMs with --no-absolute-filenames
6d3f8bd4d fetch2/git: quote shallow extra ref arguments
3a8937cc4 fetch2: validate striplevel parameter
a32064d0f fetch2: validate deb/ipk data member names
```

Seven of those are input validation or auth hardening, so this is worth taking on its own merits independent of the oe-core repair.

The collision: my #241 replaces these same two lines with `branch: "race-review-fixes"` / `commit: 999cc46a`, and `race-review-fixes` at `42ff7daf` does **not** contain `fc13d719` - it forked at `82abbfcd` and is 11 behind. So if #241 landed after this, all 11 above would silently disappear from the composition, including the hashserv unihash validation.

No change needed here. Recording it so the ordering is deliberate: this merges first, then I rebase and re-pin.

jetm (commented):
Both pins check out and I verified them against the actual repos rather than eyeballing the hashes. This is a repair, not an optional bump - details inline on the oe-core pin.

Short version: the change is correct, nothing avocado-specific is lost, and I have no objection to the content. The one thing that needs a decision is ordering, because this PR and my own #241 edit the same two lines from the same base blob.

## Ordering vs #241

[#241](https://github.com/avocado-linux/meta-avocado/pull/241) repoints the bitbake entry to `branch: "race-review-fixes"` at `999cc46a`. Same file, same two lines, same starting blob (`index 341ef45b` in both diffs). So they conflict textually, and whichever merges second silently reverts the other's pin.

**This one should go first.** If #241 merged after, the tree would drop the 11 bitbake commits below - `race-review-fixes` at `42ff7daf` is 11 behind `2.8-avocado` and does not contain `fc13d719`. That set includes `hashserv: validate unihash values`, which is the same subsystem my race fixes touch, so losing it is the worst possible trade.

That is my problem to fix, not yours: after this lands I rebase `race-review-fixes` onto `fc13d719` and re-point #241. My pin is stale anyway - it names `999cc46a` while the PR head is now `42ff7daf`.

## Underlying process note

`vendor-openembedded-core:scarthgap-avocado` was force-pushed, which is what broke your build. Worth deciding as a team whether that branch should be rebased at all, or only ever fast-forwarded - a rewrite orphans every pin in every downstream composition at once, and the failure surfaces as an unrelated-looking kas fetch error rather than as anything pointing at the rewrite. Not something to solve in this PR.

## Scope note

I did not run the full review pipeline on this one - it is a two-line commit-hash bump, and the analysis that mattered was cross-repo ancestry and patch-id comparison, which reading the diff cannot answer. The verification commands are in the inline comments so you can re-run them.

kas/vendor/oe.yml
jetm: Correcting myself: "no rewrite involved" above is wrong, and the count is off by one. `2.8-avocado` was force-pushed too, so this pin is a repair for the same reason the oe-core one is - not an optional bump.

```
git merge-base --is-ancestor 82cf21bc6 origin/2.8-avocado   # exit 1 - orphaned
git merge-base 82cf21bc6 origin/2.8-avocado                 # 82abbfcdbda9
```

Both branches forked from `82abbfcd`. Relative to that base, the old pin carried exactly one commit and the new pin carries eleven - and the old one's single commit is present on the new branch by patch-id, re-landed with a new hash:

```
OLD 37ec676bc9647d50  82cf21bc6  fetch2/git: detect LFS in subdir .gitattributes...
NEW 37ec676bc9647d50  fc13d7197  fetch2/git: detect LFS in subdir .gitattributes...
```

So **10 genuinely new commits, not 11** - my list above double-counted the LFS one, which is the old pin rebased rather than something added. Nothing was dropped on this side either, same as oe-core.

Nothing changes about the verdict or the ordering: still correct, still merge this first. What it does change is the process note in the main review - both vendor branches were rewritten, not just `vendor-openembedded-core`, so whatever the team decides about rebasing vendor branches applies to `vendor-bitbake` as well.

It also means my own rebase is across a rewrite rather than a fast-forward: `race-review-fixes` forked at that same `82abbfcd`, so re-pinning #241 is a real rebase onto `fc13d719`, not a trivial move.

jetm (approved):

jetm:
@hiagofranco heads up on what happened to this one, because it is going to get closed and the reason has nothing to do with your change.

Your change is correct. I checked both pins against the repos rather than taking the hashes on faith, and your commit message had already documented the cause better than my review did: both vendor branches were force-pushed, so `82cf21bc` and `e8e93ce1` are orphaned and `kas` cannot check them out. I also confirmed nothing avocado-specific was lost in either rebase - the staging identical-symlink and `sstatesig` nativesdk patches come back with identical patch-ids, and the two trees agree exactly on `sstatesig.py` and `staging.bbclass`.

What blocks this PR is mechanical. It is a fork PR and you are a first-time contributor here, and the repo's Actions policy is `fork-pr-contributor-approval: all_external_contributors`, so GitHub creates no workflow run for it at all - zero check runs on the head, and nothing sitting in `action_required` for a maintainer to approve either. The `scarthgap` ruleset carries a `code_scanning` rule requiring a CodeQL result, so `mergeStateStatus` is `BLOCKED` with no path forward. Waiting would not have cleared it.

So I re-landed your commit unchanged from a branch in this repo as #250 - same commit, you as author, verified identical by `git patch-id`. CodeQL ran there and is green, which is the proof of the cause: the same one-file change gets its four checks on an in-repo branch and none on a fork. #250 carries `Closes #249`, so this closes automatically when that merges.

Sorry for the detour on your first one here. The policy is the real problem and I have raised it separately - it will hit every external fork PR the same way, so it is worth a team decision rather than a workaround each time. Nothing for you to redo.

jetm:
Hi Hiago,

I opened #250 with this same change from a branch in this repo, and CodeQL passes there. The two diffs are byte-identical, both bumping the same two pins in `kas/vendor/oe.yml`:

```
bitbake: 82cf21bc -> fc13d719
oe:      e8e93ce1 -> 06b1c4f6
```

The reason this one cannot merge is not the change. `code_scanning` is a required rule in the branch ruleset, and the org's `fork-pr-contributor-approval` policy withholds every workflow run on a fork PR from a first-time contributor. No runs means the required check never reports, so the rule stays permanently unsatisfiable no matter how many times it is re-triggered. Re-landing it from a repo branch was the way around it.

Could you close this one and follow #250 instead? Your approval is already on this PR, and #250 still needs a codeowner review, so a look there is what unblocks it.

Worth noting the underlying policy is still in place, so the next external fork PR will hit exactly the same wall. That is a separate thing to fix and I have not raised it yet.

hiagofranco:
@jetm sure, let me close this one and continue there. Thanks.

PR #251, opened 2026-08-03
meta-avocado-x86-64: build nvidia-gpu-modules into the x86-64 feed

mobileoverlord:
## Problem

After #243, `packagegroup-avocado-nvidia-gpu` is pulled into every x86-64 *complete* build via `PKG_EXTRA_INSTALL` (in `avocado-x86-64.inc`), but the driver never actually lands in the feed. Installing anything that needs it fails:

```
nothing provides kernel-module-nvidia needed by packagegroup-avocado-nvidia-gpu-rtx-50
nothing provides kernel-module-nvidia-drm ...
nothing provides kernel-module-nvidia-modeset ...
nothing provides kernel-module-nvidia-uvm ...
```

The packagegroup **metadata** RPM builds (and carries the `Requires`), but the `kernel-module-nvidia*` RPMs it requires are never produced.

## Root cause

- `kernel-module-nvidia*` are **dynamic** packages (`PACKAGES_DYNAMIC`, via `nvidia-gpu-modules`' `inherit module`), so their names don't match the provider recipe's PN.
- The feed images (`avocado-pkg-extra`) inherit `image-packages-only`, which sets `do_rootfs[noexec]`.
- With no running `do_rootfs`, the runtime closure still builds **static** providers (efibootmgr, etc.) via `recrdeptask`, but never pulls a **dynamic** package's out-of-tree provider recipe into the build graph.

Confirmed with `bitbake -g avocado-pkg-extra`: `nvidia-gpu-modules` is **absent** from the build list even though `packagegroup-avocado-nvidia-gpu` is in `IMAGE_INSTALL` (verified via `bitbake -e`). The recipe parses fine (`bitbake-layers show-recipes nvidia-gpu-modules` → `meta-avocado-x86-64 595.84`); it's simply never scheduled.

This is why the driver appeared to "work" during bring-up (we had explicitly run `bitbake nvidia-gpu-modules`, so the RPMs were cached in the feed) and then vanished after a clean.

## Fix

Add a hard build-order dependency in the packagegroup so the provider's RPMs are always written whenever the packagegroup is built:

```bitbake
do_package_write_rpm[depends] += "nvidia-gpu-modules:do_package_write_rpm"
```

Single-file change, kept in the packagegroup that already owns the nvidia RDEPENDS. Applies to every x86-64 target (v2/v3/v4), which all pull the packagegroup via `PKG_EXTRA_INSTALL`. After this, a plain `bitbake avocado-complete` produces `kernel-module-nvidia{,-drm,-modeset,-uvm}` into `deploy/rpm` — no reliance on `do_rootfs` resolving a dynamic out-of-tree provider.

### DIFF
--- a/meta-avocado-x86-64/conf/machine/include/avocado-x86-64.inc
+++ b/meta-avocado-x86-64/conf/machine/include/avocado-x86-64.inc
@@ -20,3 +20,11 @@ MACHINE_EXTRA_RRECOMMENDS += "kernel-modules"
 # Applies to v2/v3/v4 (all require this include). This replaces the ad-hoc
 # per-machine kas local_conf_header that only covered v3.
 PKG_EXTRA_INSTALL:append = " packagegroup-avocado-nvidia-gpu"
+
+# packagegroup-avocado-nvidia-gpu RDEPENDS the kernel-module-nvidia* packages,
+# which are produced dynamically (PACKAGES_DYNAMIC) by the out-of-tree
+# nvidia-gpu-modules recipe. The feed image (avocado-pkg-extra) is packages-only
+# with do_rootfs[noexec], which does not pull dynamic out-of-tree providers into
+# the build -- so name the provider here and image-packages-only.bbclass forces
+# its RPMs into the feed. See that class for the mechanism.
+IMAGE_PACKAGES_ONLY_FORCE_PROVIDERS += "nvidia-gpu-modules"
--- a/meta-avocado/classes/image-packages-only.bbclass
+++ b/meta-avocado/classes/image-packages-only.bbclass
@@ -25,3 +25,34 @@ IMAGE_FEATURES = ""
 
 # Ensure these recipes are excluded from world builds
 EXCLUDE_FROM_WORLD = "1"
+
+# --- Force dynamic out-of-tree package providers into the feed ---------------
+#
+# do_rootfs[noexec] above suppresses the dynamic runtime-provider resolution a
+# real do_rootfs performs. Static providers in the image's runtime closure are
+# still pulled in via do_rootfs[recrdeptask] (this is why e.g. efibootmgr, a plain
+# RDEPENDS of a PKG_EXTRA_INSTALL packagegroup, still builds), but a *dynamic*
+# PACKAGES_DYNAMIC provider that lives in its own out-of-tree recipe -- e.g.
+# kernel-module-nvidia, produced by nvidia-gpu-modules via `inherit module` -- is
+# never added to the build graph. Its RPMs never reach DEPLOY_DIR_RPM, so the
+# packagegroup metadata RPM (carrying the Requires) lands in the feed while the
+# modules it needs do not, and `avocado install` later fails with "nothing
+# provides kernel-module-...".
+#
+# Any packages-only image that (transitively) pulls such a provider lists the
+# provider recipe(s) in IMAGE_PACKAGES_ONLY_FORCE_PROVIDERS -- typically set in
+# the machine/vendor conf next to the PKG_EXTRA_INSTALL that adds the packagegroup.
+# The class turns each into a hard build+deploy dependency of do_build. do_build
+# is deliberately chosen over the packagegroup's own do_package_write_rpm: it is
+# never sstate/setscene-covered, so a shared-sstate build that restores the
+# packagegroup from setscene cannot prune the dependency, and the provider is
+# always built or setscene-restored (either path deploys its RPMs into
+# DEPLOY_DIR_RPM, which is what the feed index scans).
+IMAGE_PACKAGES_ONLY_FORCE_PROVIDERS ??= ""
+
+python () {
+    providers = (d.getVar('IMAGE_PACKAGES_ONLY_FORCE_PROVIDERS') or '').split()
+    if providers:
+        deps = ' '.join('%s:do_package_write_rpm' % p for p in providers)
+        d.appendVarFlag('do_build', 'depends', ' ' + deps)
+}


meta-avocado-x86-64/recipes-avocado/packagegroups/packagegroup-avocado-nvidia-gpu.bb
```diff
@@ -50,3 +50,16 @@ RDEPENDS:${PN} = " \
...
+# (PACKAGES_DYNAMIC, via nvidia-gpu-modules' `inherit module`), so their names do
+# not match the provider recipe's PN. The Avocado feed images (avocado-pkg-extra)
+# inherit image-packages-only, which sets do_rootfs[noexec]; without a running
+# do_rootfs, the runtime closure builds *static* providers but never pulls a
+# *dynamic* package's out-of-tree provider recipe into the build graph. The
+# result is the packagegroup RPM (with its Requires) landing in the feed while
+# kernel-module-nvidia* silently do not -- so `avocado install` later fails with
+# "nothing provides kernel-module-nvidia". A hard build-order dependency on the
+# provider's package-write task guarantees the modules are always produced.
+do_package_write_rpm[depends] += "nvidia-gpu-modules:do_package_write_rpm"
```
jetm: **Worth confirming this holds on a shared-sstate build, not just from scratch.**

The feed index is built by scanning `DEPLOY_DIR_RPM` at `avocado-distro:do_create_repo_map` time (`meta-avocado/classes/avocado-repo-map.bbclass:24-40`) - I verified that premise, and writing the RPMs there genuinely is sufficient for them to be indexed. But a `[depends]` on the real `do_package_write_rpm` is not honored on the setscene path, and it does not re-run if the RPM is removed after it was written. Two concrete paths:

1. A build with warm shared sstate for this packagegroup but no sstate object for `nvidia-gpu-modules:do_package_write_rpm`. BitBake covers the packagegroup via `do_package_write_rpm_setscene`, the real task never runs, and its `[depends]` is pruned from the must-build set - so the modules are absent from `deploy/rpm` and `do_create_repo_map` indexes a feed missing them.
2. The stale-sstate prune this repo already documents at `avocado-multikernel.bbclass:70-73`: "bitbake's stale-sstate prune (`sstate_eventhandler_stalesstate` ...) routinely empties exactly these common RPMs from the default deploy/rpm on a kernel re-sign". That removes already-written `kernel-module-nvidia*` RPMs after the dependency was satisfied, and nothing re-writes them.

Both land the original "nothing provides kernel-module-nvidia" symptom. A from-sstate-mirror build is the check that would settle it.

Separate and smaller: the hardcoded `_rpm` task name silently no-ops if `PACKAGE_CLASSES` ever moves off rpm. `avocado.inc:37` sets it as a weak default (`?=`), and setting a flag on a task that does not exist is not an error in BitBake - so with `package_ipk` the guard vanishes with no diagnostic and the feed quietly regresses. Nothing in-tree overrides it today and the whole feed stack is rpm/dnf, so this is low severity; it is the silence that is worth a line of defence.

meta-avocado-x86-64/recipes-avocado/packagegroups/packagegroup-avocado-nvidia-gpu.bb
```diff
@@ -50,3 +50,16 @@ RDEPENDS:${PN} = " \
     ${PN}-rtx-40 \
```
jetm, line 50: **If the diagnosis in this comment is right, this is the first of five identical bugs.**

The root cause you name is `do_rootfs[noexec]` in the shared `meta-avocado/classes/image-packages-only.bbclass`, which means every `PKG_EXTRA_INSTALL` packagegroup that `RDEPENDS` on a `PACKAGES_DYNAMIC` out-of-tree kernel module has the identical exposure. Four siblings do, and none gets the guard:

- `packagegroup-avocado-reterminal-extra.bb:12-15` - `kernel-module-ili9881d`, wired via `kas/machine/reterminal.yml:23`
- `packagegroup-avocado-imx-extra.bb:10` - `kernel-module-nxp-wlan`, via `kas/vendor/nxp.yml:33`
- `packagegroup-avocado-solidrun-extra.bb:17-20` - via `kas/machine/rzv2n-sr-som.yml:28`
- `packagegroup-avocado-tegra-extra.bb:42` - `kernel-module-r8168`, via `kas/vendor/nvidia.yml:32`

Taking reterminal as the concrete case: build `avocado-pkg-extra` for `avocado-reterminal` and the packagegroup RPM lands in the feed carrying `Requires: kernel-module-ili9881d`, but that dynamic package's provider recipe is never pulled into the graph, so no `kernel-module-ili9881d` RPM reaches `DEPLOY_DIR_RPM` and `avocado install` on the board fails with "nothing provides kernel-module-ili9881d" - the same symptom your comment describes.

Not asking you to fix four other boards in this PR. But the fix belongs in `image-packages-only.bbclass` rather than repeated per recipe, and if this lands as-is it is worth a follow-up issue so the other four do not each get discovered the hard way on a customer board.

The mechanism itself is right and idiomatic - `do_<task>[depends] += "<recipe>:do_<task>"` is the established idiom here (~20 uses, e.g. `avocado-stone.bb:39-43`, `tegraflash-bsp.bb:54-60`), and the comment style matches its siblings.

jetm (commented):
The one-line fix is correct and idiomatic for this tree, and the diagnosis in the comment checks out - I verified the feed index really is built from a live `DEPLOY_DIR_RPM` scan (`avocado-repo-map.bbclass:24-40`), so forcing the RPMs there is genuinely sufficient for them to be indexed. Two findings inline.

The substantive one is that if the root cause is `do_rootfs[noexec]` in the shared `image-packages-only.bbclass`, this is the first of five identical bugs rather than a one-recipe problem - four sibling `PKG_EXTRA_INSTALL` packagegroups have the same exposure and no guard. The other asks whether the guard survives a shared-sstate build, since `[depends]` on a real task is not honored on the setscene path.

Things I checked that came back clean: no arch guard or `COMPATIBLE_MACHINE` is needed (`meta-avocado-x86-64` is only in `BBLAYERS` for the x86-64 machine ymls, and `nvidia-gpu-modules` has a matching `COMPATIBLE_HOST`); the newly-forced recipe introduces no new licence gate (`nvidia-gpu-modules_595.84.bb` is `MIT & GPL-2.0-only` with no `LICENSE_FLAGS`, and the proprietary userspace half is pre-existing and already accepted at `kas/base.yml:130`); no duplicate or conflicting nvidia recipe (`meta-avocado-nvidia` is Tegra/L4T, a disjoint stack); and omitting an `mc:` prefix on the `[depends]` is right here, since no x86-64 machine yml pulls a multi-kernel feature.

mobileoverlord:
Reworked per review — thanks @jetm, both findings were on point. Force-pushed; the per-recipe `do_package_write_rpm[depends]` is replaced by a general mechanism in `image-packages-only.bbclass`.

### Finding #1 — survives shared-sstate / setscene

Moved the force off the packagegroup's `do_package_write_rpm` and onto the **image recipe's `do_build`**, which is never sstate/setscene-covered — so a setscene-restored packagegroup can't prune it. The provider is always built *or* setscene-restored, and either path deploys its RPMs into `DEPLOY_DIR_RPM` (which `do_create_repo_map` scans).

Verified on exactly the build you asked for — a **clean-from-sstate** build:

```
# nvidia-gpu-modules came from setscene this build:
tmp/work/.../nvidia-gpu-modules/.../temp/log.do_package_write_rpm_setscene
# ...and its RPMs are present anyway:
kernel-module-nvidia-6.6.127-yocto-standard-595.84-r0.0...rpm
kernel-module-nvidia-drm-...  -modeset-...  -uvm-...  -peermem-...   (5/5)
```

On the stale-prune (`sstate_eventhandler_stalesstate` on kernel re-sign): that's the multi-config kernel-re-sign path handled by `avocado-multikernel.bbclass`, and no x86-64 machine pulls a multi-kernel feature, so `kernel-module-nvidia*` here isn't subject to it. The class mechanism doesn't touch multikernel merge behavior.

### Finding #2 — 1 of 5, fix in the class

Agreed. The fix now lives in `image-packages-only.bbclass` as `IMAGE_PACKAGES_ONLY_FORCE_PROVIDERS` (data-driven, one place). x86-64 is wired in `avocado-x86-64.inc` next to its `PKG_EXTRA_INSTALL`. The other four (`reterminal`, `imx`, `solidrun`, `tegra`) are the same exposure and become a one-line `IMAGE_PACKAGES_ONLY_FORCE_PROVIDERS +=` in their vendor/machine conf — but since forcing a previously-absent module can surface a latent build break, I'd rather each land validated on its own board than blind in this PR. Filing a follow-up issue to track the four.

### Minor — hardcoded `_rpm`

Kept `do_package_write_rpm`: the whole feed stack is rpm/dnf and `do_create_repo_map` scans `DEPLOY_DIR_RPM`, so a move off rpm would need broader changes here anyway. Noted as the low-severity item it is.

PR #254, opened 2026-08-04
intel-x86-64: reject a non-512 block size

jetm:
## Problem

The three `intel-x86-64` provisioning scripts read `block_size` from the stone
manifest and then never use it. Every commit touching one of them fails the
`shellcheck` pre-commit hook with SC2034, so merges from `scarthgap` have been
landing with the hook skipped.

Deleting the read would silence the warning and hide a real constraint. That
constraint is *not* the sector arithmetic: `sgdisk` operating on a regular file
numbers LBAs in 512-byte units unconditionally, never reads `block_size`, and is
passed no sector size anywhere in this tree, so the hardcoded `* 2048`
conversions are always internally consistent no matter what the manifest
declares. `meta-avocado-stm/stone/stm32mp25-dk/build-disk-image.sh:88` states
this outright.

What actually breaks is the product, not the arithmetic. The GPT these scripts
emit is addressed in 512-byte LBAs, and a 4Kn rootdisk addresses in its own
logical sectors — so it cannot consume that table. That is a property of the
target device, and refusing it is the honest response.

## Solution

Adopt the guard the rockchip sibling already uses
(`meta-avocado-rockchip/stone/orangepi-5-plus/build-disk-image.sh:34`) and
refuse the unsupported case outright. The 512-byte assumption is real and
load-bearing, so stating it beats emitting a table the target cannot read. Every
in-tree manifest declares 512, so nothing that works today changes.

## Key changes

- Refuse a non-512 `block_size` in `intel-x86-64-v2/v3/v4/stone-provision-img.sh`
- Comment records *why* the constraint exists (a 512-LBA GPT is unusable on a
  4Kn rootdisk) and explicitly notes that the `* 2048` arithmetic is not the
  reason, cross-referencing the stm script so the two files stop contradicting
  each other
- Clears SC2034 on all three files, so they no longer need a skipped hook

## Reviewer notes

No production behavior change for any current target: all in-tree manifests
declare `block_size: 512`, and an absent key still defaults to 512.

Verified behaviorally, since the repo has no shell test harness. With a manifest
declaring 4096 the scripts previously computed a full layout
(`offset=1MiB size=64MiB`) before failing later for an unrelated reason; they now
exit 1 at the guard before any layout work. Re-checked that 512 and an absent key
both proceed untouched, and mutation-tested the guard (inverted condition, dropped
`exit 1`, removed block) to confirm the check actually detects its absence.
`shellcheck` is clean on all three files and `pre-commit run --files` passes, so
this commit was made without `--no-verify`.

Tested on Nuvo-9000 (intel-x86-64-v3).

One caveat on the evidence, worth stating plainly: the checks above prove the
guard *fires*, not that its absence corrupts an image. No two images were built
and compared, and `sgdisk` is not installed on the box used here. The reason to
refuse rests on documented `sgdisk` behavior plus the stm cross-reference rather
than on a measurement.

Two related findings, both left out to keep this reviewable:

1. Four more scripts read `BLOCK_SIZE` without using it, and the handling across
   the tree is now inconsistent four ways. Filed as #257 rather than widening a
   PR whose title names one board family.
2. The `shfmt` hook currently enforces nothing. `.pre-commit-config.yaml` sets
   `args: [-i, "2", -ci, -bn]`, which replaces the upstream hook's default
   `args: [--write]`; without `--write` or `-d`, shfmt prints to stdout and always
   exits 0. Verified against `scop/pre-commit-shfmt@v3.13.1-1`. Adding `-d` would
   make it real, but the existing tree is not 2-space-clean, so that is its own
   change.

No CI exercises this: the only workflow triggers on the `build` label, and a build
run does not run provisioning scripts.

### DIFF
--- a/meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
+++ b/meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
@@ -23,6 +23,14 @@ BUILD_DIR="$AVOCADO_STONE_BUILD_DIR"
 PLATFORM=$(jq -r '.runtime.platform' "$MANIFEST")
 BLOCK_SIZE=$(jq -r '.storage_devices.rootdisk.block_size // 512' "$MANIFEST")
 
+# Every sector calculation below converts MiB with a hardcoded "* 2048", which
+# is only correct for 512-byte sectors. Refuse anything else rather than emit a
+# partition table whose offsets are wrong by the block-size ratio.
+if [ "$BLOCK_SIZE" != "512" ]; then
+    echo "ERROR: only 512-byte blocks supported (manifest declares ${BLOCK_SIZE})" >&2
+    exit 1
+fi
+
 echo "=== Creating disk image for ${PLATFORM} ==="
 echo "  Data directory: ${DATA_DIR}"
 echo "  Build directory: ${BUILD_DIR}"
--- a/meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
+++ b/meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
@@ -23,9 +23,12 @@ BUILD_DIR="$AVOCADO_STONE_BUILD_DIR"
 PLATFORM=$(jq -r '.runtime.platform' "$MANIFEST")
 BLOCK_SIZE=$(jq -r '.storage_devices.rootdisk.block_size // 512' "$MANIFEST")
 
-# Every sector calculation below converts MiB with a hardcoded "* 2048", which
-# is only correct for 512-byte sectors. Refuse anything else rather than emit a
-# partition table whose offsets are wrong by the block-size ratio.
+# The "* 2048" conversions below are always correct: sgdisk numbers LBAs in
+# 512-byte units for an image file regardless of block_size, which is stone
+# metadata rather than an sgdisk input (same note at
+# meta-avocado-stm/stone/stm32mp25-dk/build-disk-image.sh:88). What breaks is the
+# product, not the arithmetic - GPT addresses in the disk's own logical sectors,
+# so a 4Kn rootdisk cannot consume the 512-LBA table this emits. Refuse instead.
 if [ "$BLOCK_SIZE" != "512" ]; then
     echo "ERROR: only 512-byte blocks supported (manifest declares ${BLOCK_SIZE})" >&2
     exit 1
--- a/meta-avocado-x86-64/stone/intel-x86-64-v3/stone-provision-img.sh
+++ b/meta-avocado-x86-64/stone/intel-x86-64-v3/stone-provision-img.sh
@@ -23,6 +23,14 @@ BUILD_DIR="$AVOCADO_STONE_BUILD_DIR"
 PLATFORM=$(jq -r '.runtime.platform' "$MANIFEST")
 BLOCK_SIZE=$(jq -r '.storage_devices.rootdisk.block_size // 512' "$MANIFEST")
 
+# Every sector calculation below converts MiB with a hardcoded "* 2048", which
+# is only correct for 512-byte sectors. Refuse anything else rather than emit a
+# partition table whose offsets are wrong by the block-size ratio.
+if [ "$BLOCK_SIZE" != "512" ]; then
+    echo "ERROR: only 512-byte blocks supported (manifest declares ${BLOCK_SIZE})" >&2
+    exit 1
+fi
+
 echo "=== Creating disk image for ${PLATFORM} ==="
 echo "  Data directory: ${DATA_DIR}"
 echo "  Build directory: ${BUILD_DIR}"
--- a/meta-avocado-x86-64/stone/intel-x86-64-v3/stone-provision-img.sh
+++ b/meta-avocado-x86-64/stone/intel-x86-64-v3/stone-provision-img.sh
@@ -23,9 +23,12 @@ BUILD_DIR="$AVOCADO_STONE_BUILD_DIR"
 PLATFORM=$(jq -r '.runtime.platform' "$MANIFEST")
 BLOCK_SIZE=$(jq -r '.storage_devices.rootdisk.block_size // 512' "$MANIFEST")
 
-# Every sector calculation below converts MiB with a hardcoded "* 2048", which
-# is only correct for 512-byte sectors. Refuse anything else rather than emit a
-# partition table whose offsets are wrong by the block-size ratio.
+# The "* 2048" conversions below are always correct: sgdisk numbers LBAs in
+# 512-byte units for an image file regardless of block_size, which is stone
+# metadata rather than an sgdisk input (same note at
+# meta-avocado-stm/stone/stm32mp25-dk/build-disk-image.sh:88). What breaks is the
+# product, not the arithmetic - GPT addresses in the disk's own logical sectors,
+# so a 4Kn rootdisk cannot consume the 512-LBA table this emits. Refuse instead.
 if [ "$BLOCK_SIZE" != "512" ]; then
     echo "ERROR: only 512-byte blocks supported (manifest declares ${BLOCK_SIZE})" >&2
     exit 1
--- a/meta-avocado-x86-64/stone/intel-x86-64-v4/stone-provision-img.sh
+++ b/meta-avocado-x86-64/stone/intel-x86-64-v4/stone-provision-img.sh
@@ -23,6 +23,14 @@ BUILD_DIR="$AVOCADO_STONE_BUILD_DIR"
 PLATFORM=$(jq -r '.runtime.platform' "$MANIFEST")
 BLOCK_SIZE=$(jq -r '.storage_devices.rootdisk.block_size // 512' "$MANIFEST")
 
+# Every sector calculation below converts MiB with a hardcoded "* 2048", which
+# is only correct for 512-byte sectors. Refuse anything else rather than emit a
+# partition table whose offsets are wrong by the block-size ratio.
+if [ "$BLOCK_SIZE" != "512" ]; then
+    echo "ERROR: only 512-byte blocks supported (manifest declares ${BLOCK_SIZE})" >&2
+    exit 1
+fi
+
 echo "=== Creating disk image for ${PLATFORM} ==="
 echo "  Data directory: ${DATA_DIR}"
 echo "  Build directory: ${BUILD_DIR}"
--- a/meta-avocado-x86-64/stone/intel-x86-64-v4/stone-provision-img.sh
+++ b/meta-avocado-x86-64/stone/intel-x86-64-v4/stone-provision-img.sh
@@ -23,9 +23,12 @@ BUILD_DIR="$AVOCADO_STONE_BUILD_DIR"
 PLATFORM=$(jq -r '.runtime.platform' "$MANIFEST")
 BLOCK_SIZE=$(jq -r '.storage_devices.rootdisk.block_size // 512' "$MANIFEST")
 
-# Every sector calculation below converts MiB with a hardcoded "* 2048", which
-# is only correct for 512-byte sectors. Refuse anything else rather than emit a
-# partition table whose offsets are wrong by the block-size ratio.
+# The "* 2048" conversions below are always correct: sgdisk numbers LBAs in
+# 512-byte units for an image file regardless of block_size, which is stone
+# metadata rather than an sgdisk input (same note at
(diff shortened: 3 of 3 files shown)


meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
```diff
@@ -23,6 +23,14 @@ BUILD_DIR="$AVOCADO_STONE_BUILD_DIR"
 PLATFORM=$(jq -r '.runtime.platform' "$MANIFEST")
 BLOCK_SIZE=$(jq -r '.storage_devices.rootdisk.block_size // 512' "$MANIFEST")
 
+# Every sector calculation below converts MiB with a hardcoded "* 2048", which
```
jetm: **Comment's stated mechanism is not what the code does**

This comment says the `* 2048` conversions would be "wrong by the block-size ratio" for a non-512 block size. That is not accurate for this code path. `sgdisk` operating on a regular file numbers LBAs in 512-byte units unconditionally - it never reads `block_size` from the manifest, and nothing here passes it a sector size. Lines 193-194 feed it `MIB * 2048`, so the emitted partition table is internally consistent no matter what the manifest declares.

`meta-avocado-stm/stone/stm32mp25-dk/build-disk-image.sh:88-89` states this outright:

```
# sgdisk works in 512-byte sectors regardless of block_size (BLOCK_SIZE
# in the manifest is for stone metadata, not sgdisk).
```

There is still a real reason to refuse, but it is a different one: the GPT we emit is 512-LBA, and a 4Kn rootdisk cannot consume a 512-LBA GPT. That is a statement about the target device, not about arithmetic in this script. Worth rewording so the next person to touch the sector math is not told to look for a ratio bug that isn't there.

Applies identically to the v3 and v4 copies of this hunk.

meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
```diff
@@ -23,6 +23,14 @@ BUILD_DIR="$AVOCADO_STONE_BUILD_DIR"
...
+# is only correct for 512-byte sectors. Refuse anything else rather than emit a
+# partition table whose offsets are wrong by the block-size ratio.
+if [ "$BLOCK_SIZE" != "512" ]; then
```
jetm, line 32: **Five board families, four different behaviors for block_size**

The guard itself matches existing precedent - `meta-avocado-rockchip/stone/orangepi-5-plus/build-disk-image.sh:34-35` already has these exact two lines, same wording. But across the tree the handling is now inconsistent four ways:

- rockchip: guards (identical to this)
- x86-64 v2/v3/v4: guards, after this PR
- renesas `rzv2n-sr-som/build-disk-image.sh:25,81-82`: reads `BLOCK_SIZE`, ignores it, same hardcoded `* 2048`
- alif `alif-e8-devkit/build-disk-image.sh:28`: reads `BLOCK_SIZE`, never uses it at all
- stm `stm32mp25-dk`: documents that the guard is unnecessary by design

The eMMC/SD boards are where a 4Kn rootdisk is actually plausible; x86-64 is the least likely to see one. So the two families that would benefit most are the two still unguarded. I am not proposing to widen this PR - flagging it so the inconsistency gets settled deliberately in one place rather than drifting further apart one board at a time.

jetm (commented):
Second read of my own change, cold. The shell mechanics are fine - the guard fires before any sector math, the `>&2` + `exit 1` shape matches the existing error idiom in these scripts, and every manifest in the tree declares an integer `512` so the string comparison is safe.

Two things worth fixing before this lands, both inline. The short version: the guard is defensible and matches existing precedent in the rockchip script, but the comment explains it with a mechanism that isn't real, and the tree now handles `block_size` four different ways across five board families.

meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
jetm: You're right, and I checked it rather than taking it on faith: no script in the tree passes a sector size to `sgdisk` - the only match for that pattern anywhere is the stm comment you quoted. So `* 2048` is unconditionally correct here and a 4096 manifest would emit the same bytes. The ratio bug I described cannot happen.

Reworded in 5ade114 across all three copies. The comment now leads with the fact that the arithmetic is always right, says `block_size` is stone metadata rather than an `sgdisk` input, cross-references `meta-avocado-stm/stone/stm32mp25-dk/build-disk-image.sh:88` so the two files stop contradicting each other, and gives the reason that survives checking: GPT addresses in the disk's own logical sectors, so the 512-LBA table this emits is unusable on a 4Kn rootdisk.

The guard body and its behavior are unchanged - this is a comment-only commit. Verified after the edit: all three files still byte-identical (same md5), `bash -n` parses, `shellcheck` clean, the guard still refuses 4096 and still does not fire on 512 or on an absent key.

Worth recording why my original claim was wrong, since it is the more useful lesson: my check proved the guard *fires*, and I let that stand in for proving that its absence corrupts anything. It doesn't. I never produced two images and compared them. I could not run that comparison now either - `sgdisk` isn't installed on this box - which is exactly why the documented behavior plus your stm cross-reference is the evidence here rather than a measurement.

meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
jetm: Agreed on all of it, including the direction of the risk - x86-64 is the family least likely to meet a 4Kn rootdisk, and the eMMC/SD boards where it is plausible are the two still unguarded. That is worth fixing, and not here.

Taking the narrow option deliberately: this PR keeps its stated scope (three x86-64 scripts) and the inconsistency gets settled in a follow-up rather than by widening a PR whose title names one board family. Widening it would touch renesas and alif and reword the stm comment, none of which a reviewer of this diff signed up to review.

I am leaving this thread open rather than resolving it, because the follow-up is not filed yet - resolving on a promise is how this kind of thing drifts. It will carry the four-way table as written, and the real question it needs to answer is whether the guard is the right shape at all: refusing 4Kn is cheap, but making the sector math honor `block_size` is the only option that supports such a target instead of rejecting it. Nothing in the tree needs that today, which is an argument for the guard now and against generalizing the math across five families speculatively.

One correction to the table for whoever picks it up: stm does not merely document that the guard is unnecessary, it is the file that documents *why* - `build-disk-image.sh:88` is now cross-referenced from the x86-64 comment, so that note is load-bearing for two families rather than one.

jetm: Filed as #257, so this can close now - I left it open specifically because resolving on a promise is how a follow-up disappears.

The issue carries the four-way table as written, plus the part that makes the choice tractable: the guard is not arithmetically forced, since `sgdisk` numbers LBAs in 512-byte units for a regular file and is passed no sector size anywhere in the tree. The real constraint is that a 512-LBA table cannot be consumed by a 4Kn rootdisk, which is a property of the target device.

It lists three options - guard everywhere, drop the unused reads, or make the sector math honor `block_size` - and recommends either of the first two over the third until a board actually needs 4Kn. The value is applying one choice to all five families rather than which one wins.

meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
```diff
@@ -23,6 +23,17 @@ BUILD_DIR="$AVOCADO_STONE_BUILD_DIR"
 PLATFORM=$(jq -r '.runtime.platform' "$MANIFEST")
 BLOCK_SIZE=$(jq -r '.storage_devices.rootdisk.block_size // 512' "$MANIFEST")
 
+# The "* 2048" conversions below are always correct: sgdisk numbers LBAs in
+# 512-byte units for an image file regardless of block_size, which is stone
+# metadata rather than an sgdisk input (same note at
+# meta-avocado-stm/stone/stm32mp25-dk/build-disk-image.sh:88). What breaks is the
+# product, not the arithmetic - GPT addresses in the disk's own logical sectors,
+# so a 4Kn rootdisk cannot consume the 512-LBA table this emits. Refuse instead.
```
lee-reinhardt, line 31: FYI: this comment and the PR body give different reasons.

The body says the `* 2048` arithmetic "is only correct for 512-byte sectors" and produces offsets "wrong by the block-size ratio". This says the conversions are always correct, and what actually breaks is that a 4Kn disk cannot consume the 512-LBA table. `5ade114` ("state the real reason") is what changed it, and this version is the accurate one.

Only worth an edit because the repo squash-merges: the body becomes the commit message, so the superseded explanation lands in history next to code that contradicts it.

meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
```diff
@@ -23,6 +23,17 @@ BUILD_DIR="$AVOCADO_STONE_BUILD_DIR"
 PLATFORM=$(jq -r '.runtime.platform' "$MANIFEST")
 BLOCK_SIZE=$(jq -r '.storage_devices.rootdisk.block_size // 512' "$MANIFEST")
 
+# The "* 2048" conversions below are always correct: sgdisk numbers LBAs in
+# 512-byte units for an image file regardless of block_size, which is stone
+# metadata rather than an sgdisk input (same note at
```
lee-reinhardt, line 28: FYI, unrelated to this change — noticed while confirming the citation.

The stm script uses `+$(( size_sectors - 1 ))` at `:92`, which is the off-by-one this file's own comment at `:198-201` records as fatal for `var`: `+N` means N sectors, so subtracting one makes every partition a sector short, and btrfs sized to the exact partition then reports "device total_bytes should be at most ...". renesas and alif look the same.

I did not check whether those targets have an exactly-sized `var`, so I can't say it bites today — but it is the same three files as the `BLOCK_SIZE` sweep you offered to file, so probably worth folding in rather than a separate pass.

lee-reinhardt (approved):

mobileoverlord:
Tested on Nuvo-9000 (intel-x86-64-v3): guard does not fire for the in-tree manifest (`block_size: 512`) and provisioning is unchanged.

Rebased onto current `scarthgap` (was 2 behind; merge base `2c42026`). No conflicts, and the net diff is identical to the pre-rebase head — 3 files, 33 insertions.

PR body updated per @lee-reinhardt's note: it still carried the superseded "wrong by the block-size ratio" explanation that `5ade114` retracted. Since the repo squash-merges, that text would have become the commit message next to code contradicting it.

Second finding from that review (the `sgdisk` `+N` off-by-one) recorded on #257 — including `rockchip/orangepi-5-plus`, which the review thread missed and which is the file cited as this PR's guard precedent.

mobileoverlord (approved):
Approving. Verified independently rather than on the diff alone:

- `sgdisk` is passed no sector size anywhere in the tree, so `* 2048` is unconditionally correct and the reworded comment is the accurate one.
- x86-64 v2/v3/v4 hunks are byte-identical to each other.
- Guard fires on 4096, does not fire on 512 or an absent key.
- Rebased onto current `scarthgap` with no conflicts and an identical net diff.
- Tested on Nuvo-9000 (intel-x86-64-v3).

Scope kept narrow deliberately; the tree-wide inconsistencies are on #257.

meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
mobileoverlord: Fixed — PR body rewritten. The Problem section now states that the `* 2048` arithmetic is unconditionally correct (sgdisk numbers LBAs in 512-byte units for a regular file and is passed no sector size anywhere in the tree) and that the real constraint is a 512-LBA GPT being unusable on a 4Kn rootdisk. Confirmed the retracted "wrong by the block-size ratio" wording is gone, so the squash commit message will match the code comment.

meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
mobileoverlord: Good catch, and confirmed. Recorded on #257 rather than folded in here, to keep this PR to its stated scope.

Verified the mechanism: x86-64 documents it at intel-x86-64-v3/stone-provision-img.sh:198-201 and carries the fix (`+${size_sectors}`); rockchip:111, renesas:83, alif:88 and stm:92 all still use the `- 1` form. Note rockchip/orangepi-5-plus has it too — not just renesas and alif — which matters because that is the file cited as this PR guard precedent. stm32mp25-dk is not an actively supported target, so the live candidates are rockchip, renesas and alif.

PR #255, opened 2026-08-04
intel-x86-64: unbreak the build after the 6.6.142 kernel bump

mobileoverlord:
Two independent breakages from #250's kernel bump (6.6.127 → **6.6.142**), both validated on hardware (Neousys Nuvo-9000, RTX 4000 SFF Ada): kernel builds, driver builds and loads, `nvidia-smi` reports the GPU.

Reworked per @jetm's review — the MTK-enable approach is replaced by a local backport, and the BPF cgroup change is split into its own PR (see below).

### 1. Kernel `-Werror=unused-function` in `btmtk.h`
6.6.142's `CONFIG_BT_MTK`-disabled `#else` stubs are plain `static`, so `btusb.c` trips `-Werror=unused-function` on `btmtk_fw_get_filename`. **Backport upstream `23e88450bb04`** ("btmtk: Mark all stub functions as inline") as a local patch — not yet in 6.6.y. This fixes only the warning; it does **not** enable MediaTek BT (so no `linux-firmware-mediatek` needed, no `btmtk` glue in `btusb.ko`). Drops itself when 6.6.y backports it (patch stops applying).

### 2. `nvidia-gpu-modules` objtool can't load `libelf.so.1`
6.6.142 runs objtool on external modules; its RPATH into the kernel's `recipe-sysroot-native` is stale (rm_work) by the time the driver builds. `DEPENDS += "elfutils-native"` + prepend the native libdir to `LD_LIBRARY_PATH` (guarded so no trailing `:`).

### Split out
The container **BPF cgroup** kernel config (`CONFIG_CGROUP_BPF`) is not a 6.6.142 regression and affects other board families, so it moved to its own PR: (link to follow).

### Review items addressed
- MTK firmware / inert `BT_MTK=m` / "nobody else pays" → all moot; switched to the header backport.
- Trailing-colon `LD_LIBRARY_PATH` → guarded.
- BPF fragment x86-64-only + "wants its own change" → split out.

### DIFF
--- a/meta-avocado-x86-64/recipes-kernel/nvidia-gpu-modules/nvidia-gpu-modules_595.84.bb
+++ b/meta-avocado-x86-64/recipes-kernel/nvidia-gpu-modules/nvidia-gpu-modules_595.84.bb
@@ -19,6 +19,15 @@ S = "${WORKDIR}/git"
 
 inherit module
 
+# The x86-64 kernel runs objtool (ORC/retpoline validation) on external modules.
+# objtool is a host tool the kernel recipe built against its OWN recipe-sysroot-
+# native; that RPATH is stale during this build (rm_work has cleaned the kernel's
+# native sysroot), so objtool aborts at runtime with
+# "libelf.so.1: cannot open shared object file". Depend on elfutils-native so
+# libelf is staged into our native sysroot, and put it on the loader path in
+# do_compile below.
+DEPENDS += "elfutils-native"
+
 # The open kernel modules are built from the kernel-open/ subdirectory.
 # The top-level Makefile dispatches into kernel-open/ when modules= target is used.
 MODULES_MODULE_SYMVERS_LOCATION = "kernel-open"
@@ -46,6 +55,11 @@ EXTRA_OEMAKE += " \
 
 # Build the open modules via the top-level Makefile target
 do_compile() {
+    # Let the kernel's objtool find libelf.so.1 (from elfutils-native) -- its own
+    # RPATH into the kernel's native sysroot no longer resolves here. Guard the
+    # append so an unset LD_LIBRARY_PATH does not leave a trailing ':' (an empty
+    # element the loader treats as the CWD).
+    export LD_LIBRARY_PATH="${STAGING_LIBDIR_NATIVE}${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
     oe_runmake KERNEL_UNAME=${KERNEL_VERSION} modules
 }
 
--- a/meta-avocado-x86-64/recipes-kernel/linux/files/0001-Bluetooth-btmtk-mark-all-stub-functions-as-inline.patch
+++ b/meta-avocado-x86-64/recipes-kernel/linux/files/0001-Bluetooth-btmtk-mark-all-stub-functions-as-inline.patch
@@ -1,0 +1,74 @@
+From a49f567bec75d39a4014c4511c98cdc03d0304db Mon Sep 17 00:00:00 2001
+From: Justin Schneck <j.schneck@peridio.com>
+Date: Tue, 4 Aug 2026 17:47:39 -0400
+Subject: [PATCH] Bluetooth: btmtk: mark all stub functions as inline
+
+Backport of upstream commit 23e88450bb04 ("Bluetooth: btmtk: Mark all
+stub functions as inline"), which is not yet in linux-stable 6.6.y.
+
+The CONFIG_BT_MTK-disabled #else stubs in btmtk.h are plain `static`, so
+btusb.c -- which includes the header unconditionally -- trips
+-Werror=unused-function on btmtk_fw_get_filename when MTK support is off,
+breaking the kernel build. Mark the stubs `static inline`, matching the
+already-inline btmtk_set_bdaddr stub directly above them.
+
+Upstream-Status: Backport [23e88450bb04]
+Signed-off-by: Justin Schneck <j.schneck@peridio.com>
+---
+ drivers/bluetooth/btmtk.h | 14 +++++++-------
+ 1 file changed, 7 insertions(+), 7 deletions(-)
+
+diff --git a/drivers/bluetooth/btmtk.h b/drivers/bluetooth/btmtk.h
+index 3055b97..ae26b5f 100644
+--- a/drivers/bluetooth/btmtk.h
++++ b/drivers/bluetooth/btmtk.h
+@@ -196,39 +196,39 @@ static inline int btmtk_set_bdaddr(struct hci_dev *hdev,
+ 	return -EOPNOTSUPP;
+ }
+ 
+-static int btmtk_setup_firmware_79xx(struct hci_dev *hdev, const char *fwname,
++static inline int btmtk_setup_firmware_79xx(struct hci_dev *hdev, const char *fwname,
+ 				     wmt_cmd_sync_func_t wmt_cmd_sync)
+ {
+ 	return -EOPNOTSUPP;
+ }
+ 
+-static int btmtk_setup_firmware(struct hci_dev *hdev, const char *fwname,
++static inline int btmtk_setup_firmware(struct hci_dev *hdev, const char *fwname,
+ 				wmt_cmd_sync_func_t wmt_cmd_sync)
+ {
+ 	return -EOPNOTSUPP;
+ }
+ 
+-static void btmtk_reset_sync(struct hci_dev *hdev)
++static inline void btmtk_reset_sync(struct hci_dev *hdev)
+ {
+ }
+ 
+-static int btmtk_register_coredump(struct hci_dev *hdev, const char *name,
++static inline int btmtk_register_coredump(struct hci_dev *hdev, const char *name,
+ 				   u32 fw_version)
+ {
+ 	return -EOPNOTSUPP;
+ }
+ 
+-static int btmtk_process_coredump(struct hci_dev *hdev, struct sk_buff *skb)
++static inline int btmtk_process_coredump(struct hci_dev *hdev, struct sk_buff *skb)
+ {
+ 	return -EOPNOTSUPP;
+ }
+ 
+-static void btmtk_fw_get_filename(char *buf, size_t size, u32 dev_id,
++static inline void btmtk_fw_get_filename(char *buf, size_t size, u32 dev_id,
+ 				  u32 fw_ver, u32 fw_flavor)
+ {
+ }
+ 
+-static int btmtk_usb_hci_wmt_sync(struct hci_dev *hdev,
++static inline int btmtk_usb_hci_wmt_sync(struct hci_dev *hdev,
+ 				  struct btmtk_hci_wmt_params *wmt_params)
+ {
+ 	return -EOPNOTSUPP;
+-- 
+2.55.0
+
--- a/meta-avocado-x86-64/recipes-kernel/linux/linux-yocto_%.bbappend
+++ b/meta-avocado-x86-64/recipes-kernel/linux/linux-yocto_%.bbappend
@@ -11,6 +11,7 @@ SRC_URI:append:avocado-x86-64 = " \
     file://tpm.cfg \
     file://nvidia-gpu.cfg \
     file://x86-efi.cfg \
+    file://0001-Bluetooth-btmtk-mark-all-stub-functions-as-inline.patch \
 "
 
 inherit avocado-kernel-feed


meta-avocado-x86-64/recipes-kernel/linux/files/avocado-extra.cfg
```diff
@@ -6,6 +6,16 @@ CONFIG_CPUSETS=y
...
 CONFIG_NAMESPACES=y
 CONFIG_NET_CLS_CGROUP=y
+# BPF cgroup device controller: on cgroup v2, runc/docker gate per-container
+# device access with a BPF_CGROUP_DEVICE program. Without these, container start
+# fails at runc create with:
+#   bpf_prog_query(BPF_CGROUP_DEVICE) failed: function not implemented
+# (x86_64_defconfig leaves BPF_SYSCALL/CGROUP_BPF unset). Deps verified:
+# BPF_SYSCALL selects BPF (already =y); CGROUP_BPF depends on BPF_SYSCALL;
+# BPF_JIT needs MODULES (=y) and HAVE_EBPF_JIT (x86).
+CONFIG_BPF_SYSCALL=y
```
jetm: **Only x86-64 gets these; nine sibling layers set neither**

There are ten `avocado-extra.cfg` fragments in the tree - alif, nvidia, nxp, qcom, qemu, raspberrypi, renesas, rockchip, stm, x86-64 - and after this change x86-64 is the only one that sets `CONFIG_BPF_SYSCALL`/`CONFIG_CGROUP_BPF`. I grepped the whole tree at this head: nothing outside `meta-avocado-x86-64` sets either symbol anywhere.

The runc failure this fixes is not x86-specific. Any board on cgroup v2 whose kernel leaves `CGROUP_BPF` off gets the same `bpf_prog_query(BPF_CGROUP_DEVICE) failed: function not implemented` at `runc create`. Whether a given family is actually broken depends on its base defconfig rather than on this fragment - `x86_64_defconfig` happens to leave BPF off, while some vendor kernels (rpi, tegra) enable it - so this is a "check each family" rather than "nine broken boards".

The one I would check first is qemu: Container Dev Mode is landing right now across avocado-cli#184, avocado-os#46 and docs#473, its documented happy path is a QEMU target, and running containers on the device is the entire point of that feature. If `linux-yocto`'s qemu config leaves `CGROUP_BPF` off, that feature is broken on the exact target its own guide tells people to use.

meta-avocado-x86-64/recipes-kernel/nvidia-gpu-modules/nvidia-gpu-modules_595.84.bb
```diff
@@ -46,6 +55,10 @@ EXTRA_OEMAKE += " \
 
 # Build the open modules via the top-level Makefile target
 do_compile() {
+    # Let the kernel's objtool find libelf.so.1 (from elfutils-native) -- its own
+    # RPATH into the kernel's native sysroot no longer resolves here.
+    LD_LIBRARY_PATH="${STAGING_LIBDIR_NATIVE}:$LD_LIBRARY_PATH"
```
jetm: **Trailing colon puts the build CWD on the loader path**

`LD_LIBRARY_PATH` is not set in a BitBake task environment, so this expands to `"/path/to/staging/lib:"` with a trailing colon. An empty element in `LD_LIBRARY_PATH` means the current directory, and during `do_compile` that is the fetched NVIDIA source root - so the dynamic loader searches unpacked third-party sources for shared libraries before falling through to the system paths.

Harmless in practice today, but it is a real search-path widening for no benefit, and the fix is one substitution:

```
LD_LIBRARY_PATH="${STAGING_LIBDIR_NATIVE}${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
```

meta-avocado-x86-64/recipes-kernel/linux/files/x86-efi.cfg
```diff
@@ -129,6 +129,14 @@ CONFIG_BT_HCIBTUSB=m
...
 CONFIG_BT_BCM=m
 CONFIG_BT_RTL=m
+# MediaTek BT as a loadable module (btmtk.ko + the MTK glue in btusb.ko). Built
+# as a module so users with MTK radios can bring it in, and nobody else pays for
+# it. Also sidesteps a 6.6.142 header bug: btmtk.h's CONFIG_BT_MTK-disabled
+# `#else` stubs are plain `static` (not `static inline`), so btusb.c tripped
+# -Werror=unused-function on btmtk_fw_get_filename when MTK was off. Enabling
+# CONFIG_BT_MTK takes the real `#if IS_ENABLED(CONFIG_BT_MTK)` branch instead.
+CONFIG_BT_MTK=m
+CONFIG_BT_HCIBTUSB_MTK=y
```
jetm: **MTK BT enabled without linux-firmware-mediatek**

This turns on the MediaTek protocol support, but `meta-avocado-x86-64/recipes-avocado/packagegroups/packagegroup-avocado-x86-64-extra.bb` has no `linux-firmware-mediatek` entry - it carries `linux-firmware-ibt` for Intel BT (line 29) plus iwlwifi/ath10k/ath11k/rtl*/i915, and that's it.

OE does not derive RDEPENDS from a module's `MODULE_FIRMWARE` declarations, so nothing pulls the blobs in implicitly. Concrete failure: a board with an MT7922 radio loads `btmtk.ko`, the driver requests `mediatek/BT_RAM_CODE_MT7922_1_1_hdr.bin`, the request fails, and the adapter never comes up - a worse user-visible state than the driver simply being absent, because the device now enumerates and then dies.

If the intent is "available for those who want it", the firmware package needs to land alongside it.

meta-avocado-x86-64/recipes-kernel/linux/files/x86-efi.cfg
```diff
@@ -129,6 +129,14 @@ CONFIG_BT_HCIBTUSB=m
 CONFIG_BT_INTEL=m
 CONFIG_BT_BCM=m
 CONFIG_BT_RTL=m
+# MediaTek BT as a loadable module (btmtk.ko + the MTK glue in btusb.ko). Built
+# as a module so users with MTK radios can bring it in, and nobody else pays for
+# it. Also sidesteps a 6.6.142 header bug: btmtk.h's CONFIG_BT_MTK-disabled
+# `#else` stubs are plain `static` (not `static inline`), so btusb.c tripped
+# -Werror=unused-function on btmtk_fw_get_filename when MTK was off. Enabling
+# CONFIG_BT_MTK takes the real `#if IS_ENABLED(CONFIG_BT_MTK)` branch instead.
+CONFIG_BT_MTK=m
```
jetm: **CONFIG_BT_MTK=m is inert here**

`BT_MTK` is a promptless select-only symbol in 6.6.y:

```
config BT_MTK
	tristate
	select FW_LOADER
```

No prompt string, so Kconfig will not honor a value written in a fragment - the symbol's value comes entirely from `select BT_MTK` in `BT_HCIBTUSB_MTK` (which is `default n`, so the `=y` on the next line is the line actually doing the work, and it clamps to `m` because `CONFIG_BT=m`).

So the `=m` here reads as load-bearing when it isn't. That matters for the reason this hunk exists: someone trimming what looks like a redundant line later would reasonably delete the `=y` instead and silently reopen the `-Werror=unused-function` build break.

meta-avocado-x86-64/recipes-kernel/linux/files/avocado-extra.cfg
```diff
@@ -6,6 +6,16 @@ CONFIG_CPUSETS=y
...
+CONFIG_CGROUP_BPF=y
+CONFIG_BPF_JIT=y
```
jetm: **BPF_JIT isn't required for the device controller**

The comment says "Without these, container start fails", which is true of `BPF_SYSCALL` and `CGROUP_BPF` but not of `BPF_JIT` - a `BPF_CGROUP_DEVICE` program runs fine under the interpreter, and the JIT is a throughput optimization. Enabling it is reasonable, it just isn't part of the fix.

Worth separating for the same reason as the `BT_MTK` line on the other fragment: when someone later audits which of these three lines the runc failure actually needed, the comment as written says all three, and the honest answer is two.

The dependency note is correct for 6.6 specifically, which is worth recording since it does not survive a kernel bump - `BPF_JIT depends on MODULES` is real in 6.6.y, but upstream dropped that dependency when JIT allocation moved to `EXECMEM` (it is gone by 7.1), so the parenthetical will go stale rather than wrong.

meta-avocado-x86-64/recipes-kernel/linux/files/x86-efi.cfg
```diff
@@ -129,6 +129,14 @@ CONFIG_BT_HCIBTUSB=m
 CONFIG_BT_INTEL=m
 CONFIG_BT_BCM=m
 CONFIG_BT_RTL=m
+# MediaTek BT as a loadable module (btmtk.ko + the MTK glue in btusb.ko). Built
+# as a module so users with MTK radios can bring it in, and nobody else pays for
```
jetm: **"nobody else pays for it" isn't right**

`CONFIG_BT_HCIBTUSB_MTK=y` compiles the MediaTek glue into `btusb.ko` for every user, and makes `btmtk.ko` a hard module dependency of `btusb.ko` - so it is loaded on any machine that has a USB Bluetooth adapter at all, MediaTek or not. The cost is small, but it is not zero and it is not opt-in.

The diagnosis in the rest of this comment is solid, for what it's worth - I checked the 6.6.y tip and the `#else` stubs in `btmtk.h` are still plain `static`, and mainline's fix (`23e88450bb04 Bluetooth: btmtk: Mark all stub functions as inline`) has not been backported to 6.6.y. Requesting that backport would decouple "ship MTK BT" from "dodge a compiler warning", which is the part of this I would rather see fixed upstream than worked around in our fragment.

jetm (commented):
Reviewed at `9d4905c`. All three diagnoses hold up - I verified them against the 6.6.y source rather than taking them on faith.

The btmtk.h header bug is real: the `#else` stubs are still plain `static` at the 6.6.y tip (6.6.146), `btusb.c` includes the header unconditionally, and `CONFIG_WERROR` defaults to `y`, so the `-Werror=unused-function` break on `btmtk_fw_get_filename` is exactly as described. Mainline fixed it in `23e88450bb04` ("Bluetooth: btmtk: Mark all stub functions as inline") and that commit is not in 6.6.y - a stable backport request would be worth filing, since it separates "we want MTK BT" from "we need to dodge a compiler warning."

Every dependency claim in the BPF comment checks out against 6.6.y, including the one that surprised me: `BPF_JIT depends on MODULES` really is there in 6.6 (upstream dropped it later when JIT allocation moved to `EXECMEM`). `CGROUP_BPF depends on BPF_SYSCALL`, `BPF_SYSCALL` is `default n` and selects `BPF`, and `x86_64_defconfig` carries no BPF lines at all. I also checked the security question that enabling `bpf()` raises and it is fine: `BPF_UNPRIV_DEFAULT_OFF` is `default y` in 6.6, so unprivileged BPF stays off without any extra sysctl work.

The objtool/libelf diagnosis is right too, and the BT Kconfig dependencies resolve - `BT_HCIBTUSB=m` and `FW_LOADER=y` are both present, and `BT_MTK` lands at `m` clamped by `CONFIG_BT=m`, so nothing is silently dropped. I also checked that `avocado-kernel-builtin-provides.bbclass` keys off `modules.builtin` rather than CONFIG symbols, so the new `=y` bool does not produce a bogus `Provides:`.

Six findings inline. The first two are the ones I would want resolved before merge - the BPF fragment landing on x86-64 only, and the missing MediaTek firmware package. The rest are a loader-path nit and three accuracy corrections to comments.

One process note: this PR now carries three unrelated concerns (BT kernel config, container BPF config, nvidia objtool fix). Not asking you to split it, but if any one of them needs reverting later they will have to come apart, and the BPF hunk in particular reads like it wants to be its own change given it likely affects other board families too.

mobileoverlord:
Thanks for the thorough review @jetm — all six addressed, and the diagnoses matched what I saw building.

**btmtk (findings on firmware / inert `BT_MTK=m` / "nobody else pays"):** dropped the MTK-enable approach entirely. Instead I backported the upstream fix you pointed at — `23e88450bb04` ("btmtk: Mark all stub functions as inline") — as a local patch (`Upstream-Status: Backport`). That fixes only the `-Werror` and enables no MediaTek BT, so all three concerns are moot: no `linux-firmware-mediatek`, no glue in `btusb.ko`, no promptless `BT_MTK=m`. We're not waiting on the 6.6.y backport — we carry it and it drops itself when 6.6.y picks it up (patch stops applying).

**LD_LIBRARY_PATH trailing colon:** fixed with your suggested guard — `${STAGING_LIBDIR_NATIVE}${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}` (the inner `$` keeps bitbake from expanding the shell construct).

**BPF fragment x86-64-only + "wants its own change":** split into #256. It enables `BPF_SYSCALL`/`CGROUP_BPF` on x86-64 (validated on hardware) **and** qemu (Container Dev Mode's target), with the `BPF_JIT` comment corrected to note it's a throughput optimization, not part of the fix, and the 6.6-specific `MODULES` dep flagged as such. Other families left as per-family verification, as you noted.

So this PR is now just the two 6.6.142 build regressions (btmtk backport + nvidia objtool). Re-requesting review.

meta-avocado-x86-64/recipes-kernel/linux/files/avocado-extra.cfg
jetm: Followed this down to the mechanism, and it is worse than "nine layers need the same two lines" - the reason x86-64 needed them at all is an upstream gap that affects every machine.

The containers feature group pulls `meta-virtualization`, whose `recipes-kernel/linux/linux-%.bbappend` includes `linux-yocto_6.6_virtualization.inc` when `DISTRO_FEATURES` carries `virtualization`. That include adds `cfg/docker.scc` to `KERNEL_CACHE_FEATURES`, and `docker.scc` pulls exactly one fragment, `docker.cfg`. In yocto-kernel-cache `origin/yocto-6.6`, that fragment's last line is:

```
CONFIG_CGROUP_BPF=y
```

and it never sets `CONFIG_BPF_SYSCALL`. But `CGROUP_BPF depends on BPF_SYSCALL` in `init/Kconfig`, and `BPF_SYSCALL` is `default n`. Across all of yocto-kernel-cache `yocto-6.6`, `CONFIG_BPF_SYSCALL=y` appears only in `cfg/debug-kselftest.cfg` and `features/bpf/bpf.cfg` - and nothing anywhere references `features/bpf`. So on a stock build Kconfig drops the `CGROUP_BPF` request as an unmet dependency, silently, and `runc create` then fails with precisely the `bpf_prog_query(BPF_CGROUP_DEVICE) failed: function not implemented` in your comment.

Which means this PR's hunk is itself the evidence: if `docker.cfg` were sufficient, you would never have hit that error to fix. Your two lines are correct - they are just also the only place in the tree that satisfies the dependency.

The one I care about most is qemu, because Container Dev Mode's documented happy path is a QEMU target. `meta-avocado-qemu/recipes-kernel/linux/files/avocado-qemux86-64/defconfig` has no BPF symbols at all. It does set `CONFIG_CGROUP_DEVICE=y`, but that is the cgroup **v1** devices controller and it does nothing on v2, which is what we boot - there is no `systemd.unified_cgroup_hierarchy=0` or `cgroup_no_v1` anywhere in the tree. The qemu `avocado-extra.cfg` carries container networking only, no cgroup block.

I have a `kernel_configme` build running for `avocado-qemux86-64` with the containers feature to confirm the merged `.config` rather than leave this as inference, and I will report the result here either way.

If it comes back without `CGROUP_BPF`, then the fix probably wants to live once where the feature group is defined rather than per-board - `packagegroup-avocado-feature-containers.bb` already exists as that seam - so enabling containers implies a kernel that can actually run them. Worth an upstream report against `docker.cfg` too, since every scarthgap consumer of `meta-virtualization` has this hole, not just us.

jetm: Correcting myself: I built it, and qemu is fine. Do not act on my "check qemu first" worry.

`bakar bitbake virtual/kernel meta-avocado/kas/machine/qemux86-64.yml:meta-avocado/kas/feature/containers.yml -c kernel_configme` on `avocado-qemux86-64`, 496 tasks, 0 errors. The merged `.config` at `linux-yocto/6.6.127+git/linux-avocado_qemux86_64-standard-build/.config` has both:

```
CONFIG_BPF_SYSCALL=y
CONFIG_CGROUP_BPF=y
```

plus `CONFIG_BPF_JIT=y`, `CONFIG_BPF_JIT_DEFAULT_ON=y`, `CONFIG_BPF_EVENTS=y` and `CONFIG_BPF_UNPRIV_DEFAULT_OFF=y`. So Container Dev Mode's QEMU path is not blocked, and the part of my earlier comment that inferred otherwise from the absence of BPF symbols in qemu's fragments was wrong. Reading the fragments was not a substitute for reading the merged config.

What still stands is narrower but not nothing. `cfg/docker.cfg` really does request `CONFIG_CGROUP_BPF=y` without `CONFIG_BPF_SYSCALL`, and your PR is evidence that on the x86-64 machines that is not sufficient - you hit the runc failure and these two lines fix it. So the two machine families genuinely differ.

What I could not determine is why, and I would rather flag that than guess. In the kernel-cache actually staged for this build, `CONFIG_BPF_SYSCALL` appears only in `cfg/debug-kselftest.cfg` and `features/bpf/bpf.cfg`, and nothing references `features/bpf`. meta-avocado sets no `KERNEL_FEATURES` anywhere and no BPF symbol outside your new x86-64 hunk. qemu's `avocado-qemux86-64/defconfig` has no BPF lines, and the x86-64 layer ships no defconfig at all - only `.cfg` fragments. Yet qemu ends up with `BPF_SYSCALL=y`. Something supplies it that is not declared in either layer.

That unknown is the part I would still chase, because an undeclared dependency that happens to be satisfied is the kind that disappears on a kernel or kernel-cache bump with no diff to point at - and the symptom would be containers breaking on qemu, which is our own documented starting point for the feature. Declaring the two symbols wherever the containers feature group is turned on would make that robust for every machine rather than repairing one, and it costs the same two lines you already wrote.

That is a suggestion about where the lines live, not a request to change this PR. Your hunk is correct for the machines it targets and I have no objection to it landing as-is.

PR #11, opened 2025-04-16
Add build support for Raspberry Pi 4

mobileoverlord:
Adding 64Bit build support for Raspberry Pi 4 which includes device trees for Compute Modules

PR #259, opened 2026-08-05
imx93-frdm: move to the NXP wrynose-6.18.20-2.0.0 BSP

jetm:
## Problem

`avocado-imx93-frdm` does not boot on i.MX93 **Rev1.2** silicon. The 6.6.36-2.1.0 SPL brings up the PCA9451A, trains LPDDR4X at 3733MTS and stages the Cortex-M33, then stops:

```
U-Boot SPL 2024.04+gde16f4f1722+p0 (Sep 02 2024 - 10:44:35 +0000)
SOC: 0xa1009300
PMIC: PCA9451A / Over Drive Voltage Mode
DDR: 3733MTS
M33 prepare ok
                      <- nothing further, on either UART
```

It never prints `Normal Boot`, i.e. it stalls *before selecting a boot device*. That makes the failure identical whether the boot container arrives from an SD card or over USB serial download, which is why it initially presents as a provisioning problem.

## This is the BSP, not this layer

Two independent observations place the fault upstream.

**NXP's own image reproduces it.** `LF_v6.6.36-2.1.0_images_FRDM_4.0_IMX93` written to SD hangs at the same line, and its SPL is the same build this layer ships: both report `gde16f4f1722`, built `Sep 02 2024 - 10:44:35`. `uuu -b emmc_all` with NXP's own bootloader fails the same way, `HID(W): LIBUSB_ERROR_TIMEOUT (-7)` at roughly 14% of the container. 14% is approximately the size of container 1, so the ROM boots SPL and the SDPS endpoint disappears before container 2 can be sent.

**Swapping only the bootloader fixes it.** Feeding `uuu` the 6.18.2 bootloader extracted from the board's factory eMMC, on the same host, cable, port and boot-mode switch setting, completes the handoff:

```
U-Boot SPL 2025.04-g99518e6b6f20
found DRAM 2CS_2GB DRAM matched
M33 prepare ok
Normal Boot
Trying to boot from BOOTROM
Boot Stage: USB boot
Download 1881088, Total size 1882112       <- container 2 via ROM_API
NOTICE:  BL31: v2.12.0(release):lf-6.18.2-1.0.0
U-Boot 2025.04 ... CPU: NXP i.MX93(52) Rev1.2 A55 at 1700 MHz
Detect USB boot. Will enter fastboot mode!
```

A matching report is open upstream with the same tool version, same 14% cutoff and same SPL stall: https://community.nxp.com/t5/i-MX-Processors/HID-W-LIBUSB-ERROR-TIMEOUT-7-FRDM-i-MX-93/m-p/2378347

## Change

- New `kas/vendor/nxp-frdm-imx93.yml` pinning `meta-imx` to `wrynose-6.18.20-2.0.0` (`52516b92`, identical to upstream `nxp-imx/meta-imx`)
- `kas/machine/imx93-frdm.yml` includes it instead of the shared `nxp-frdm.yml`
- `meta-imx-frdm` dropped for this board

**Why a separate vendor file.** `nxp-frdm.yml` is shared by `imx91-frdm`, `imx93-frdm` and `imx95-frdm`. Bumping it would move two boards with no validation on 6.18, and `imx95-frdm` pins `linux-imx` to 6.6.52 by `SRCREV` in `meta-avocado-nxp/recipes-kernel/linux/linux-imx_%.bbappend`, which a 6.18 base would contradict. `nxp.yml` states this per-family split as the intent and `compulab.yml` already follows it. **`imx91-frdm` and `imx95-frdm` keep their 6.6.36 `meta-imx` pin.** They are not untouched, though - see *What this does to the sibling machines* below.

**Why `meta-imx-frdm` is dropped rather than re-pinned.** Its only branch is `lf-6.6.36-2.1.0`, already at upstream HEAD. From whinlatter onward the FRDM machine configurations, including the `imx93-11x11-lpddr4x-frdm.conf` that `avocado-imx93-frdm.conf` requires, live in `meta-imx-bsp` itself, and the i.MX93 FRDM device trees that layer carried as kernel patches are native to the 6.18 kernel.

## What this does to the sibling machines

`kas/vendor/freescale.yml` is shared by all six i.MX machines, so the `meta-freescale` / `meta-freescale-3rdparty` bump reaches every one of them, not just this board. That bump changes NXP WiFi firmware packaging, and the shared `packagegroup-avocado-imx-extra` named a package the wrynose layer no longer builds.

None of these machines is regressed by this PR - all six already fail on the base branch, at the layer sanity check, on the `wic/wks files ... need to be moved to files/wic` abort the bump exists to clear. What changes is *where* they stop:

| Machine | Where it stops on this branch |
| --- | --- |
| `imx93-frdm` | boots (see Verification) |
| `imx93-evk` | would abort on `Nothing RPROVIDES 'firmware-nxp-wifi-nxp8997-pcie'`; fixed here by dropping that entry from the shared default, which wrynose `meta-freescale` does not package at all |
| `imx8mp-evk`, `imx91-frdm`, `imx95-frdm` | layer compat - their 6.6.36 `meta-imx` declares `LAYERSERIES_COMPAT_fsl-bsp-release = "mickledore nanbield scarthgap"`, so they cannot parse against wrynose OE-core regardless of firmware packaging. Out of scope here; each needs its own BSP move. |
| `avocado-ucm-imx8m-plus` | unaffected - its machine conf drops this packagegroup and substitutes the CompuLab one |

Read from the pinned trees rather than from a build: only `imx93-frdm` was built.

## Known upstream gap, not fixed here

wrynose `meta-imx` `layer.conf:277` adds the `nxpaw693-sdio` machine feature for `mx93-nxp-bsp`, but wrynose `meta-freescale` packages `firmware-nxp-wifi-nxpaw693-pcie` and no `-nxpaw693-sdio`. Whoever first fits an AW693 module to an i.MX93 board will land on this. Worth a report upstream; it is not this PR's to fix.

**Why wrynose and not whinlatter.** Only `wrynose-6.18.20-2.0.0` lists `wrynose` in `LAYERSERIES_COMPAT_fsl-bsp-release`; `whinlatter-6.18.2-1.0.0` stops at `whinlatter`.

## Verification

Hardware: FRDM-IMX93 board V1.0, silicon `i.MX93` revision `1.2`, read from `/sys/devices/soc0/soc_id` and `/sys/devices/soc0/revision` on the running board.

The build is green (`5412/5412 tasks, 0 errors`) and an Avocado image built from this branch boots to a login prompt on that hardware, from **both** eMMC and SD:

```
U-Boot SPL 2026.04-lf_v2026.04+g6eeef838dac+p0
Normal Boot                                        <- 6.6.36 never reached this
NOTICE:  BL31: v2.14.1(release):lf-6.18.20-2.0.0
U-Boot 2026.04-lf_v2026.04+g6eeef838dac+p0
Loading Environment from MMC... Reading from redundant MMC(1)... OK
[    0.000000] Linux version 6.18.20-2.0.0-gb096ce610e95 (aarch64-avocado-linux-gcc 15.2.0)
[    9.835687] erofs (device mmcblk1p6): mounted with root inode @ nid 36.
Avocado OS 0.0.0 avocado-imx93-frdm ttyLP0
```

## Follow-ups, resolved during bring-up

1. **Does SD boot work on 6.18?** Yes, and it needs no environment patching. `mmc_get_env_dev()` in `arch/arm/mach-imx/imx9/soc.c` asks the boot ROM which device it booted and reads the environment from that instance, so `CONFIG_ENV_MMC_DEVICE_INDEX` only applies to a non-SD/MMC boot. Device-tree aliases pin the numbering (`mmc0` = `usdhc1`, non-removable, 8-bit = eMMC; `mmc1` = `usdhc2`, `cd-gpios`, 4-bit = microSD), so the built-in `devnum=1` / `mmcblk=1` / `rootdev=/dev/mmcblk1p6` already target the card whether or not eMMC is populated. That is why `stone-provision-sd.sh` carries no `fw_setenv` call. The `sd` profile stays.
2. **`stone-provision-uuu-emmc.sh` and the eMMC boot partition.** Not the problem it looked like. The failure that motivated the question was the environment being rejected outright (`bad CRC, using default environment`), which sent U-Boot to NXP's built-in defaults where `mmcpart=1` names the raw `imx-boot` partition. Root cause was a Kconfig rename, fixed in this series (see below); with a valid environment the eMMC path boots.

## Environment fix included in this series

U-Boot renamed `SYS_REDUNDAND_ENVIRONMENT` to `ENV_REDUNDANT`. `env-mmc.cfg` still set the old name, which `merge_config.sh` merges without warning and `make oldconfig` then silently drops, taking `ENV_OFFSET_REDUND` with it. `MKENVIMAGE_EXTRA_ARGS = "-r"` still emitted a redundant blob (`CRC32 | flags | data`), so a U-Boot built without redundancy checksummed from the wrong offset and could never match.

Both spellings are now carried, because this fragment is shared by every NXP board and imx8mp-evk / imx91-frdm / imx93-evk / imx95-frdm are still on the 2024.04 BSP where the old name is the live one. The silent-drop behaviour that caused the bug is what makes carrying both safe. Confirmed on hardware by the `Reading from redundant MMC(1)... OK` line above.

## Known, not caused by this PR

`[DEPEND] Dependency failed for Reload manager and retrigger activation targets after Avocado extensions are merged` appears twice per boot. It is a `BindsTo=` on a condition-skipped unit and fires on every Avocado device with no extensions installed, on any machine. Fixed separately in #263 against `scarthgap`.


meta-avocado-nxp/recipes-avocado/packagegroups/packagegroup-avocado-imx-extra.bb
```diff
@@ -16,6 +16,32 @@ NXP_WIFI_FIRMWARE = " \
...
+# linux-firmware carries the identical firmware as linux-firmware-nxp8997-sdio
+# and -pcie, and each already RDEPENDS on linux-firmware-nxp8997-common, so the
+# two entries below are a complete substitution. meta-freescale sets no
+# RPROVIDES or RCONFLICTS on its 8997 packages, so nothing shadows them.
+#
+# Keyed on the machine rather than mx93-nxp-bsp because imx93-evk is still on
+# 6.6.36, where the firmware-nxp-wifi packages do exist. Fold this into the
+# default once every i.MX board is on a BSP that removes them.
+NXP_WIFI_FIRMWARE:avocado-imx93-frdm = " \
+  linux-firmware-nxp8997-sdio \
```
jetm: **The Wi-Fi firmware substitution installs no WLAN firmware**

The two packages substituted in carry no firmware files at all, so the 88W8997 WLAN blobs are silently dropped from the image. oe-core wrynose `meta/recipes-kernel/linux-firmware/linux-firmware_20260410.bb`:

```
1520 FILES:${PN}-nxp8997-common = " \
1521     ${nonarch_base_libdir}/firmware/nxp/uartuart8997_bt_v4.bin* \
1522     ${nonarch_base_libdir}/firmware/nxp/helper_uart_3000000.bin* \
1523 "
1524 ALLOW_EMPTY:${PN}-nxp8997-pcie = "1"
1525 ALLOW_EMPTY:${PN}-nxp8997-sdio = "1"
```

`-nxp8997-sdio` and `-nxp8997-pcie` have no `FILES` - they are `ALLOW_EMPTY` shims. The commit that created them is `4a6cf0b533 linux-firmware: add firmware files for NXP **BT** chipsets`, and the only real content reachable through `RDEPENDS` is `-nxp8997-common`'s two Bluetooth files. The WLAN blobs live elsewhere: `linux-firmware-sd8997` (`sdsd8997_combo_v4.bin`, `sd8997_uapsta.bin`) and `linux-firmware-pcie8997`.

Worse, the filename the NXP driver actually loads is in oe-core under no package at all. meta-freescale wrynose `recipes-bsp/firmware-imx/firmware-nxp-wifi_1.1.bb:95` ships `mrvl/sdiouart8997_combo_v4.bin`; grepping oe-core's linux-firmware for `sdiouart8997` returns nothing. So the comment at :31-33 claiming "a complete substitution" is not achievable from `linux-firmware`.

Failure: the build succeeds (both packages resolve, `ALLOW_EMPTY` suppresses the empty-package error), the image boots, then `moal`/`mwifiex_sdio` probes and the kernel logs `Direct firmware load for nxp/sdiouart8997_combo_v4.bin failed with error -2`. No `wlan0`, and zero build-time signal.

Direction: the layer already patches this recipe (`meta-avocado-nxp/recipes-bsp/firmware-imx/firmware-nxp-wifi_%.bbappend`), so re-add `${PN}-nxp8997-sdio` to `PACKAGES` and restore the `mrvl` symlink that meta-imx's `do_install` override drops.

One thing to resolve before merge: meta-imx wrynose `meta-imx-bsp/conf/layer.conf:172-173` sets `MACHINE_FEATURES_IMX_REMOVALS ?= "mrvl8997 nxp8997-pcie nxp8997-sdio rvgpu-emu"` and :277 adds `nxpaw693-sdio` for `mx93-nxp-bsp`. NXP's own 6.18 BSP drops 8997 on i.MX9 and moves to AW693 - so the right fix may be AW693 rather than restoring 8997. Which module is physically on the FRDM board needs the hardware.

meta-avocado-nxp/recipes-avocado/packagegroups/packagegroup-avocado-imx-extra.bb
```diff
@@ -16,6 +16,32 @@ NXP_WIFI_FIRMWARE = " \
...
+#   #-----------------don't upstream, keep in imx ------------------------
+#   PACKAGES:remove = " ${PN}-nxp8997-sdio "
+#
+# leaving no firmware-nxp-wifi-nxp8997-* package to depend on. oe-core's
+# linux-firmware carries the identical firmware as linux-firmware-nxp8997-sdio
+# and -pcie, and each already RDEPENDS on linux-firmware-nxp8997-common, so the
+# two entries below are a complete substitution. meta-freescale sets no
+# RPROVIDES or RCONFLICTS on its 8997 packages, so nothing shadows them.
+#
+# Keyed on the machine rather than mx93-nxp-bsp because imx93-evk is still on
```
jetm: **The machine-keyed scoping rests on a false claim, and the siblings are broken**

The comment says "imx93-evk is still on 6.6.36, where the firmware-nxp-wifi packages do exist". `kas/machine/imx93-evk.yml` includes only `base.yml`, `vendor/freescale.yml`, `feature/virtualization.yml` and `target/distro.yml` - no `nxp.yml` - so there is no 6.6.36 anything on that machine. Its `firmware-nxp-wifi` comes from wrynose meta-freescale, where `firmware-nxp-wifi_1.1.bb:41-57` lists only `${PN}-nxp8997-sdio` among the 8997 packages; `-nxp8997-common` and `-nxp8997-pcie` are gone (they exist on `origin/scarthgap` at :58-60). The 6.6.36 meta-imx bbappend does not re-add `-nxp8997-pcie` to `PACKAGES` either - it only sets `FILES`/`RDEPENDS`.

Failure: `bitbake avocado-image-rootfs` for `avocado-imx93-evk` - and also imx8mp-evk, ucm-imx8m-plus, imx91-frdm, imx95-frdm - aborts at taskgraph time with `Nothing RPROVIDES 'firmware-nxp-wifi-nxp8997-pcie'`.

The default block is unchanged from base `b7bea88` (`kas/vendor/freescale: track the wrynose branches`), so the breakage predates this PR. What this diff adds is a comment asserting those machines are fine, which sends the next reader looking in the wrong place.

Decide: fix the default in this series, or correct the comment to say the other machines are known-broken pending their own move. Leaving the false claim is the one option that is not acceptable.

meta-avocado-distro/recipes-avocado/distro/avocado-stone.bb
```diff
@@ -11,6 +11,14 @@ PACKAGE_ARCH = "${MACHINE_ARCH}"
 
 DEPENDS += " stone-native avocado-img-var avocado-img-rootfs"
 
+# fw_setenv, for the provision scripts that patch a U-Boot environment inside
+# the disk image before writing it - the uuu-emmc profile sets devnum and
+# mmcblk that way. jq and fwup already arrive through stone-native; fw_setenv
+# had no provider, so do_stone_provision died with "fw_setenv: command not
+# found" the first time a profile that needs it was selected. libubootenv is
+# the recipe that ships it (it also PROVIDES u-boot-fw-utils).
+DEPENDS += " libubootenv-native"
```
jetm: **NXP/`uuu-emmc`-only dependency added to the machine-agnostic recipe**

`fw_setenv` is called only from `meta-avocado-nxp/stone/avocado-imx/stone-provision-uuu-emmc.sh:55-56`, and `uuu-emmc` appears in `STONE_PROVISIONING` for the five `meta-avocado-nxp` machines only - the other ~25 use `img`/`sd`/`usb`/`tegraflash`/`ufs`/`serial`/`emmc`/`direct`. So this lands in the layer-agnostic base recipe for a tool most machines can never invoke.

The justification in the comment is also wrong: "jq and fwup already arrive through stone-native" - `meta-avocado-distro/recipes-devtools/stone/stone_2.2.0.bb` declares no `DEPENDS` at all (only `inherit cargo cargo-update-recipe-crates` and `BBCLASSEXTEND`). They arrive through `meta-avocado-nxp/recipes-avocado/distro/avocado-stone.bbappend:7` (`DEPENDS += " jq-native mkfat-native fwup-native"`), which is exactly the file this line belongs in. Or `DEPENDS:append:stone-uuu-emmc`.

Cost: every non-NXP machine build fetches and builds `libubootenv-native` plus `zlib-native` and `libyaml-native` for nothing.

jetm (commented):
Cold second read of my own change. Draft, but posting - the first finding ships an image with no WLAN firmware and no build-time signal, and that should not wait for un-drafting.

Static review only, no build. Everything below was checked against the local vendor clones (`meta-imx` at `52516b92a7`, `meta-freescale` at `2e2fc69da5`, oe-core `wrynose`) rather than reasoned from the diff.

Three findings inline. A fourth could not be anchored because the file is not in this diff - it is only re-exposed by the kas change:

**`meta-avocado-nxp/conf/machine/include/fsl-imx-preferred-env.inc:6-12` and `:22-28` - the gstreamer pins did not move with the BSP.** `PREFERRED_VERSION_gstreamer1.0* = "1.24.0.imx"` is now unsatisfiable: meta-imx at `52516b92a7` ships `1.28.1.imx` across base/good/bad and `1.28.1` for ugly/libav; oe-core wrynose ships `1.28.2`; and `fsl-imx-preferred-env.inc:38` `BBMASK`s meta-freescale's gstreamer, removing the third candidate. Nothing provides `1.24.0.imx`, so the pin yields only a `NOTE: preferred version ... not available` and bitbake picks the highest PV - `1.28.2` beats `1.28.1.imx` on the final component. imx93-frdm therefore silently loses the NXP gstreamer stack, and gstreamer is genuinely in the build (`packagegroup-avocado-extra.bb:197-202` RDEPENDS on all six via `avocado-pkg-extra.bb:9`). `MACHINE_GSTREAMER_1_0_PLUGIN:mx9-nxp-bsp = "imx-gst1.0-plugin"` then builds NXP's plugin against an unpatched base: either it fails to compile, or it builds and the board loses hardware video acceleration with no error. This is specific to this PR - imx91-frdm and imx95-frdm stay on `nxp-frdm.yml`'s 6.6.36 meta-imx, which does ship `1.24.0.imx`. One-line fix: bump the pins to `1.28.1.imx`, or drop them if the NXP stack is not wanted here - but decide, rather than leaving an unsatisfiable pin.

What I checked and found correct, so it does not get relitigated:

- **Pins resolve and are not floating.** `52516b92a7` is on meta-imx `origin/wrynose-6.18.20-2.0.0`, `2e2fc69da5` on meta-freescale `origin/wrynose` (both by `merge-base --is-ancestor`); both fragments pin branch **and** commit.
- **`LAYERSERIES_COMPAT`** lists `wrynose` in meta-imx-bsp, meta-imx-v2x and meta-imx-ml.
- **Dropping `meta-imx-frdm` is safe** - `imx93-11x11-lpddr4x-frdm.conf` exists in meta-imx wrynose `meta-imx-bsp/conf/machine/`, satisfying `avocado-imx93-frdm.conf:13`.
- **kas mechanics are right.** kas 5.4 `repos.py:376` uses `"disabled"` as the canonical token, and `includehandler.py:353-380` merges nested Mappings recursively so `meta-imx-bsp` and `meta-imx-v2x` survive from `nxp.yml`. All four `local_conf_header` keys in the include chain are distinct, so nothing is clobbered.
- **Boot-critical DTB name is stable.** `stone-imx93-frdm.json:92` lists `imx93-11x11-frdm.dtb`, and meta-freescale wrynose `imx93-evk.inc:11-12` still emits `freescale/${KERNEL_DEVICETREE_BASENAME}.dtb`. Only optional overlays changed (`-dsi` gone; `-8mic-reve`, `-root`, `-waveshare-7inch-c-panel`, `imx93-jailhouse-inmate` added).
- **Partition layout and env offsets are unchanged and self-consistent** across `stone-imx93-frdm.json:109-117` (env at 4 MiB, redundant 4352 KiB, size 128 KiB), `avocado-imx93-frdm.conf:16` and the hardcoded offsets in `stone-provision-uuu-emmc.sh:51-52`.
- **No firmware/bootloader version skew.** Every locally-forked BSP recipe carries `COMPATIBLE_MACHINE = "avocado-imx95-frdm"`, so imx93-frdm gets meta-imx wrynose's matched set (firmware-imx 8.32, imx-atf 2.14, ele 2.0.6, system-manager 2026q2, u-boot-imx 2026.04). Same for the kernel - `linux-imx_%.bbappend:4-62` scopes its 6.6.52 `SRCREV` and patch removals to `avocado-imx95-frdm`, so nothing forces a 6.6 kernel onto this board.
- **`kernel-module-nxp-wlan` still has a provider** at the new pin, so `NXP_WIFI_DRIVER` is fine. All three added `BBMASK` regexes are plain substrings of real paths.

Three things need a build or the board, and I am not claiming them either way:

- Whether the `avocado-*.cfg` kernel fragments still apply against 6.18. `linux-imx_%.bbappend:95-97` does `cat ${UNPACKDIR}/*.cfg >> ${B}/.config` with no `olddefconfig`, so any symbol renamed or dependency-gated out in 6.18 is dropped silently. Those lines are untouched by this diff, so not a finding - but worth a `bitbake -c diffconfig virtual/kernel`.
- Whether `imx-boot` from the 6.18 set still fits the 3072 KiB partition at `stone-imx93-frdm.json:101-107`; it would overlap the env partition at 4 MiB if it grew.
- Field update: `imx_boot` and `uboot_env` are in `images` but not in `update.os_artifacts` (`stone-imx93-frdm.json:13-16`), so an OTA swaps kernel, DTB and rootfs and leaves the old bootloader and env. Fine for a Rev1.1 board already booting and irrelevant for the Rev1.2 silicon this targets, but confirming a 6.18 `Image`/DTB boots under a 6.6.36 u-boot needs hardware.

Separate bug found while reviewing, not this PR: `meta-avocado-nxp/recipes-bsp/u-boot/u-boot-imx_%.bbappend:16` reads `UBOOT_DEFCONFIG = "${@'${UBOOT_CONFIG}'.split((',', 1)[0])}"`. `(',', 1)[0]` is just `','`, so this is `.split(',')` returning a **list**, and `${@...}` stringifies it - with `UBOOT_CONFIG ??= "sd"` the value becomes the literal `['sd']`. Lines 19-20 then append `avocado.cfg` and `env-mmc.cfg` to `${S}/configs/['sd']`, a bracket glob matching nothing, so the shell creates that file and the two U-Boot config fragments never reach the real defconfig. Filing separately.

jetm:
Moving this to draft. The premise needs settling first: I have not established that Rev1.1 and Rev1.2 silicon actually differ here, and this PR moves the whole machine on the assumption that they do.

## What is actually measured

- An FRDM-IMX93 carrying **Rev1.2** silicon does not boot the 6.6.36 BSP. SPL brings up the PMIC, trains LPDDR4X and stages the Cortex-M33, then stops before selecting a boot device - identical from SD or USB serial download.
- NXP's own `LF_v6.6.36-2.1.0_images_FRDM_4.0_IMX93` reproduces it on that board, and its SPL is the same build this layer ships (`gde16f4f1722`, `Sep 02 2024`).
- The 6.18.20 BSP boots that same board to a login prompt, from both eMMC and SD.

## What is not measured

**Whether a Rev1.1 board boots 6.6.36 fine.** I have one board and it is Rev1.2, so I have never observed the working case. Everything above is consistent with "6.6.36 predates Rev1.2 support", and equally consistent with other explanations that have nothing to do with silicon revision.

That gap matters because of what this PR does: it moves `imx93-frdm` wholesale to the wrynose BSP. If Rev1.1 boards work today on 6.6.36, this changes the BSP under users who had no problem, to fix a board they do not have.

## What would settle it

Boot a Rev1.1 board on the current 6.6.36 image.

- Boots fine -> the revisions genuinely differ, and the right shape is a **distinction**, not a blanket move: the two revisions need separating, and likely their own images, rather than one machine silently meaning different hardware depending on what you bought.
- Fails the same way -> revision is not the variable, the 6.6.36 BSP is simply broken for FRDM, and this PR is right but its reasoning needs rewriting.

Either way the answer changes what lands, so it is worth having before this merges rather than after.

## What stands regardless

The other four commits are independent of the revision question and measured on hardware:

- `meta-avocado-nxp/kernel: enable the netfilter raw table` - without `CONFIG_IP_NF_RAW` no container starts on the default bridge on any Avocado board; Docker 28+ writes a raw-table PREROUTING rule per container and `docker run` dies at network setup.
- `meta-avocado-distro/systemd: stop waiting on unconnected ports` - stock `80-wired.network` matches every ether link, so `systemd-networkd-wait-online` waits for all of them, times out at 120s and fails. Measured on a board still missing this drop-in: it delayed `docker.service` from 34s to 154s of boot, and the workload with it. `--any` takes it to 0.074s.
- `meta-avocado/rust: give nativesdk-rust's llvm-config its libraries`
- `avocado-stone: give the uuu-emmc profile its fw_setenv`

If the BSP question takes a while, those are worth splitting out and landing on their own - particularly the netfilter one, which is the difference between containers working and not working on every board, and has nothing to do with i.MX 93.

meta-avocado-nxp/recipes-avocado/packagegroups/packagegroup-avocado-imx-extra.bb
jetm: Right that the linux-firmware substitution was empty, and it is gone. `cf5fd87` replaced it with `NXP_WIFI_FIRMWARE:avocado-imx93-frdm = ""` rather than re-adding 8997 or reaching for the shims, and `f1c8a65` fixes the one claim in that comment's rationale that did not hold up.

On the 8997 vs AW693 question: 8997 is not restorable at this pin. meta-imx's `meta-imx-bsp/recipes-bsp/firmware-imx/firmware-nxp-wifi_%.bbappend` does `PACKAGES:remove = " ${PN}-nxp8997-sdio "`, so meta-freescale's only surviving 8997 package (`FILES` = `mrvl/sdiouart8997_combo_v4.bin`, at `firmware-nxp-wifi_1.1.bb:94-95`) is not produced in this layer set. And AW693 needs nothing from us: the same bbappend packages `${PN}-nxpaw693-sdio` (`sdiw693_wlan_v1.bin.se`, `sduartiw693_combo_v1.bin.se`, `uartiw693_bt_v1.bin.se`), `meta-imx-bsp/conf/layer.conf:116` puts it in `MACHINE_FIRMWARE` whenever the `nxpaw693-sdio` feature is set, `:277` sets that feature for every `mx93-nxp-bsp`, and `imx-base.inc:453` folds `MACHINE_FIRMWARE` into `MACHINE_EXTRA_RRECOMMENDS`. So the AW693 firmware already reaches an imx93-frdm image through the BSP; naming it in this packagegroup would only duplicate that.

The old comment said the opposite - that the AW693 feature "resolves to nothing either" because meta-freescale packages `-nxpaw693-pcie` and not `-nxpaw693-sdio`. True of meta-freescale alone, false once meta-imx-bsp is in the set, which it is. `f1c8a65` rewrites that paragraph; the override itself is unchanged.

The hardware question stays answered the way it was: this unit's M.2 slot is empty, and the FRDM-IMX93's supported module is an 88W8987, so there is nothing to target until a module is fitted.

Verified by parsing the recipe standalone with bitbake's own parser: `NXP_WIFI_FIRMWARE` resolves to `firmware-nxp-wifi-nxp8997-sdio` by default and to `''` under `avocado-imx93-frdm`, unchanged across `f1c8a65` (the diff is comment text only).

meta-avocado-nxp/recipes-avocado/packagegroups/packagegroup-avocado-imx-extra.bb
jetm: Fixed the default rather than the comment. `cf5fd87` drops `firmware-nxp-wifi-nxp8997-pcie` from `NXP_WIFI_FIRMWARE`, so no machine reaches `Nothing RPROVIDES` any more - `grep -rn nxp8997-pcie` over wrynose meta-freescale returns nothing, and its `PACKAGES =+` block ships `-nxp8997-sdio` and `-nxpaw693-pcie` only.

The "imx93-evk is still on 6.6.36" claim is gone with it. The replacement comment says imx93-evk is the only machine that would have hit that error, and that imx8mp-evk, imx91-frdm and imx95-frdm stop earlier on layer compat: their meta-imx pin is `scarthgap-6.6.36-2.1.0`, whose `meta-imx-bsp/conf/layer.conf:11` declares `LAYERSERIES_COMPAT_fsl-bsp-release = "mickledore nanbield scarthgap"` - no wrynose. `avocado-ucm-imx8m-plus.conf` drops this packagegroup outright.

meta-avocado-distro/recipes-avocado/distro/avocado-stone.bb
jetm: Agreed, and it landed in the bbappend. `e8a824c` puts it in `meta-avocado-nxp/recipes-avocado/distro/avocado-stone.bbappend:19` as `DEPENDS:append:stone-uuu-emmc = " libubootenv-native"`, right under the `jq-native mkfat-native fwup-native` line you pointed at - both scopings you named, in one line. `meta-avocado-distro/recipes-avocado/distro/avocado-stone.bb` carries no `libubootenv` reference; its `DEPENDS` is still `stone-native avocado-img-var avocado-img-rootfs`.

Verified by parsing the bbappend with bitbake's parser: `DEPENDS` under `OVERRIDES=stone-uuu-emmc` is `jq-native mkfat-native fwup-native libubootenv-native`, and under `stone-img` it is `jq-native mkfat-native fwup-native`. No non-uuu-emmc machine builds `libubootenv-native` for this.

meta-avocado/recipes-avocado/images/avocado-image-rootfs.bb
```diff
@@ -30,5 +30,42 @@ create_dirs() {
...
+# below skipped, and the volatile values were written afterwards - a silent
+# no-op that looked like a working patch. IMAGE_PREPROCESS_COMMAND runs after
+# do_rootfs finishes and before the erofs image is made, so the file is present
+# and nothing appends after us.
+persist_ssh_host_keys () {
+    if [ -f ${IMAGE_ROOTFS}${sysconfdir}/default/ssh ]; then
+        printf 'SYSCONFDIR=/var/lib/ssh\nSSHD_OPTS=\n' \
+            >> ${IMAGE_ROOTFS}${sysconfdir}/default/ssh
+    fi
+}
```
nicksinas: **#1 — High. `SYSCONFDIR` doesn't control where the keys go; this points sshd at the read-only `/etc/ssh` and likely leaves the device with no host key at all.**

In [`sshd_check_keys`](https://github.com/openembedded/openembedded-core/blob/wrynose/meta/recipes-connectivity/openssh/openssh/sshd_check_keys), `SYSCONFDIR` is used for exactly one thing — `mkdir -p $SYSCONFDIR` at line 40. The key paths come from a different source entirely:

```sh
HOST_KEYS=$(sshd -G -f "${sshd_config}" | grep -i '^hostkey ' | cut -f2 -d' ')
```

By clearing `SSHD_OPTS` we drop `-f /etc/ssh/sshd_config_readonly`, so `sshd_config` falls back to `/etc/ssh/sshd_config` — whose `HostKey` lines the openssh recipe writes from `OPENSSH_HOST_KEY_DIR ?= "${sysconfdir}/ssh"` (`openssh_10.3p1.bb:110`, used at `:126-132`). Nothing in this repo overrides that (`grep -rn OPENSSH_HOST_KEY_DIR` → no hits), so `HOST_KEYS` resolves to `/etc/ssh/ssh_host_{rsa,ecdsa,ed25519}_key`.

`/etc` is on the erofs rootfs: `base-files/files/fstab:3-7` mounts only `/var` and `/run` writable, and confext overlays `/etc` read-only. So `generate_key` runs `ssh-keygen -f /etc/ssh/ssh_host_rsa_key.tmp`, fails with EROFS, and `sshd.service` — which tolerates failure via `ExecStart=-@SBINDIR@/sshd` — comes up with no host keys and exits.

That trades "new key every boot" for "no SSH at all", which is worse for `avocado container dev up` than the bug being fixed. Worth confirming with `ssh-keyscan` across two reboots; a serial login prompt wouldn't catch this.

**Suggested fix**, which is one variable and does exactly what this function is reaching for:

```
OPENSSH_HOST_KEY_DIR_READONLY_CONFIG = "/var/lib/ssh"
```

in the distro conf, and delete this hook. oe-core's own `read_only_rootfs_hook` then writes `SSHD_OPTS='-f /etc/ssh/sshd_config_readonly'`, that config carries `HostKey /var/lib/ssh/…` (`openssh_10.3p1.bb:109`, `:137-143`), `generate_key` does the `mkdir -p`, and `/var` is persistent. No image post-processing, and it survives an oe-core bump.

**Secondary:** this appends unconditionally, so re-running `do_image` without `do_rootfs` duplicates the two lines in the file each time.

meta-avocado-distro/recipes-devtools/cmake/cmake-native_%.bbappend
```diff
@@ -0,0 +1,33 @@
...
+#     option() at :1404 never touches the cache (CMP0077).
+#   * -DHAVE_LIBIDN2=0 -DHAVE_IDN2_H=0 is inert for a subtler reason: :1411
+#     and :1412 set both as normal variables whenever find_package succeeds,
+#     overwriting whatever the cache holds.
+#
+# The only state that decides the outcome is LIBIDN2_FOUND, so disable the
+# find itself. CMake's own CMAKE_DISABLE_FIND_PACKAGE_<pkg> forces the
+# find_package(Libidn2 QUIET) at :1408 to fail, leaving HAVE_IDN2_H and
+# HAVE_LIBIDN2 at their OFF defaults from :1405 and :1406.
+EXTRA_OECMAKE += "-DCMAKE_DISABLE_FIND_PACKAGE_Libidn2=1"
```
nicksinas: **#2 — Medium. This flag is already set by the recipe you're appending to.**

At the oe-core commit `kas/vendor/oe.yml` pins (`0c69588b`), `cmake-native_4.3.1.bb:34` reads:

```
    -DCMAKE_DISABLE_FIND_PACKAGE_Libidn2=ON \
```

…inside the same `EXTRA_OECMAKE` block whose e2fsprogs entry the comment above quotes as precedent. So one of two things is true: this line is dead code carrying 16 lines of authoritative-sounding but stale rationale, or the `idn2.h: No such file or directory` failure you hit has a different cause and is still unfixed.

The `NGHTTP2` line above is genuinely new and correct — `CMAKE_USE_SYSTEM_LIBRARY_NGHTTP2` has no entry in the recipe's seven opt-outs. Worth splitting them so the real fix isn't obscured by the one that isn't, and both look like good oe-core submissions alongside the existing opt-outs.

meta-avocado/recipes-core/glib-2.0/glib-2.0_%.bbappend
```diff
@@ -0,0 +1,3 @@
+# Enable static library builds for qemu-user-static
+# glib-2.0 uses Meson, so we set default_library=both to build .a and .so
+EXTRA_OEMESON:append = " -Ddefault_library=both"
```
nicksinas: **#3 — Medium. This duplicates a bbappend already on the base branch, with a strictly worse comment.**

`origin/wrynose` already carries `meta-avocado-distro/recipes-core/glib-2.0/glib-2.0_%.bbappend` with the identical `EXTRA_OEMESON:append = " -Ddefault_library=both"`, and both layers are in this machine's layer set (`meta-avocado` from `base.yml:44`, `meta-avocado-distro` from `kas/target/distro.yml:8`). Net effect is the flag landing twice and two competing sources of truth for the same decision.

The pre-existing copy also explains *why*, which this one drops: "qemu-user-static links `--static` and would otherwise fail with `ld: cannot find -lglib-2.0`. (Ported from scarthgap; dropped during the wrynose migration when qemu-user-static was reauthored for QEMU 10.2.0.)"

Same applies to `meta-avocado/recipes-support/libpcre/libpcre2_%.bbappend` — `meta-avocado-distro/recipes-support/libpcre/libpcre2_%.bbappend` is already there with the same `DISABLE_STATIC = ""`.

Suggest dropping both new files and keeping the `meta-avocado-distro` ones. `meta-avocado/recipes-core/zlib/zlib_%.bbappend` is the only genuinely new one of the three — worth putting next to its siblings in `meta-avocado-distro` and giving it the same rationale, since all three exist for one reason.

nicksinas (commented):
Reviewed against the pinned upstream trees (oe-core `0c69588b`, meta-imx `52516b92`, wrynose meta-freescale, current U-Boot `env/Kconfig`) rather than from the diff alone. The BSP move itself (`a29165c`…`eeb8515`, plus the u-boot env and gstreamer commits) is well argued and I'd take it largely as-is. Most of my findings are in the front-ported build/image fixes, including one I think is a functional regression.

12 findings inline and below, most severe first. **#1 is the one that needs an answer before merge.**

---

### 8. PR description — "Known upstream gap, not fixed here" — *Low*

**This gap doesn't exist, and this PR's own code comment says so.**

meta-imx at the pinned `52516b92` ships [`meta-imx-bsp/recipes-bsp/firmware-imx/firmware-nxp-wifi_%.bbappend`](https://github.com/avocado-linux/vendor-meta-imx/blob/52516b92a70669eb58b805810f2a27a07c874771/meta-imx-bsp/recipes-bsp/firmware-imx/firmware-nxp-wifi_%25.bbappend#L15-L25), which does:

```
PACKAGES =+ " \
    ${PN}-nxpaw693-sdio \
"
FILES:${PN}-nxpaw693-sdio += " \
    ${nonarch_base_libdir}/firmware/nxp/sdiw693_wlan_v1.bin.se \
    ${nonarch_base_libdir}/firmware/nxp/sduartiw693_combo_v1.bin.se \
    ${nonarch_base_libdir}/firmware/nxp/uartiw693_bt_v1.bin.se \
"
```

That is exactly what `packagegroup-avocado-imx-extra.bb:42-48` in this PR correctly describes. meta-freescale not packaging `-nxpaw693-sdio` is fine, because meta-imx's own bbappend supplies it — which is precisely why `layer.conf:116` can put it in `MACHINE_FIRMWARE`. Suggest dropping this section rather than filing it upstream; the packagegroup comment is the accurate one.

---

### 9. `meta-avocado-nxp/dynamic-layers/compulab-bsp-imx8mp/recipes-bsp/u-boot/u-boot-compulab/env-mmc.cfg:9` — *Low*

*(Not in this diff, so noting it here.)* **Fourth copy of this fragment, still pre-rename only.**

The reasoning for carrying both spellings in `u-boot-imx/env-mmc.cfg:16-17` applies verbatim to `avocado-ucm-imx8m-plus`, whose copy still sets only `CONFIG_SYS_REDUNDAND_ENVIRONMENT=y`. The qemu and raspberrypi copies have already moved to `CONFIG_ENV_REDUNDANT` alone, so this is the only one left on the dead spelling.

Since the failure mode is silent (`bad CRC, using default environment` → NXP defaults → `mmcpart=1` names the raw `imx-boot` partition), whoever moves CompuLab to a post-rename U-Boot pays the same multi-day debug this series just paid. Worth adding the second spelling here too — and given this class of silent Kconfig drop has now bitten twice (migration doc row 45, and this PR), a `do_configure[postfuncs]` that greps the built `.config` for `ENV_REDUNDANT` would turn the next one into a build error instead of a boot failure.

---

### Verified as correct — not findings

- `stone-uuu-emmc` is a real override (`avocado.inc:6` synthesises it from `STONE_PROVISIONING`), and `do_stone_provision` is chained after `do_configure`, so `DEPENDS:append:stone-uuu-emmc = " libubootenv-native"` does put `fw_setenv` on `PATH`. Correctly scoped.
- `CONFIG_ENV_REDUNDANT` is the current U-Boot spelling (`env/Kconfig:499`), and `CONFIG_ENV_OFFSET_REDUND` does `depends on ENV_REDUNDANT`. Diagnosis confirmed.
- `ConditionPathExists=!/etc/initrd-release` is exactly complementary to `avocado-extension-initrd.service:4`. Clean fix.
- `:avocado-imx93-frdm` does beat `:mx9-nxp-bsp` — every family override reaches `MACHINEOVERRIDES` via `=.` prepend, so `${MACHINE}` stays highest priority. Empty `PREFERRED_VERSION_*` is falsy to `findPreferredProvider`, so it reads as unset. Both mechanisms are sound.
- `layers: meta-imx-ml: disabled` is valid kas (same form as `oe.yml:21`), all three meta-imx layers exist at `52516b92`, and `imx93-11x11-lpddr4x-frdm.conf` is indeed in `meta-imx-bsp` there — so dropping `meta-imx-frdm` is sound.
- The vim fix is idiomatic: `vim.inc:76-96` already steers configure through `ac_cv_*` in `EXTRA_OECONF` and uses `EXTRA_OECONF:append:class-native` at `:98`. `vim_9.2.bb:10` has `BBCLASSEXTEND = "native nativesdk"`, and nothing here builds `nativesdk-vim`, so `class-native` is sufficient.
- `${WORKDIR}/rust-snapshot` matches `rust_1.94.1.bb:51-59`, and the re-declared `patchelf-native` depends flag is a harmless (and acknowledged) duplicate of `:73`.

---

### Process note

15 commits across 21 files, and most of them — rust, cmake-native, vim, ganesha, glib/zlib/libpcre2, sshd host keys, `wait-online --any`, docker's raw table — have nothing to do with moving `imx93-frdm` to the wrynose BSP. Findings #1, #3 and #4 are all in that front-ported set, and #1 is the kind of thing that gets missed precisely because it's buried in a BSP PR. Suggest splitting the front-ported build/image fixes into their own PR.

meta-avocado-distro/recipes-core/systemd/systemd/wait-online-any.conf
```diff
@@ -0,0 +1,18 @@
...
+# Observed on FRDM-IMX93, where eth1 is unpopulated: `systemctl start docker`
+# blocked for just over two minutes before dockerd was even executed.
+#
+# --any reports online as soon as one managed link is up, which is what a device
+# with a single uplink actually means by "online". A product that genuinely
+# requires several links up before a service starts should express that with
+# RequiredForOnline= on the specific interfaces rather than by making every
+# board wait for a port that may not be connected.
+ExecStart=
+ExecStart=/usr/lib/systemd/systemd-networkd-wait-online --any
```
nicksinas: **#4 — Medium. Hardcoded libexec path, and a distro-wide policy change landing in an i.MX93 BSP PR.**

Two separate things:

**(a) The path.** `/usr/lib/systemd/systemd-networkd-wait-online` only resolves because `usrmerge` happens to be in `AVOCADO_DEFAULT_DISTRO_FEATURES` (`meta-avocado/conf/distro/avocado.conf:8`). Without it the binary is at `/lib/systemd/…`, the drop-in's `ExecStart` points at nothing, the unit fails, and `network-online.target` is never reached — strictly worse than the 120s delay this fixes, and with no build-time warning. Since the drop-in is installed from a `do_install:append` anyway, `sed` in `${rootlibexecdir}` rather than hardcoding, so the failure mode can't be reintroduced silently.

**(b) The policy.** `--any` is a defensible default and the rationale here is sound, but it changes `network-online.target` semantics for every Avocado device, including multi-NIC x86 targets where "all managed links up" may be what a product actually wants. That's not an i.MX93 change. Suggest landing it as its own PR so it gets reviewed as the distro-wide decision it is.

meta-avocado-nxp/recipes-kernel/linux/files/avocado-netfilter.cfg
```diff
@@ -11,6 +11,12 @@ CONFIG_IP_NF_TARGET_REJECT=m
 CONFIG_IP_NF_NAT=m
 CONFIG_IP_NF_TARGET_MASQUERADE=m
 CONFIG_IP_NF_MANGLE=m
+# Docker 28 added "direct access filtering", which writes a raw-table
+# PREROUTING DROP rule for every container it puts on the default bridge.
+# Without the raw table, docker 29 fails every `docker run` at network setup
+# with "can't initialize iptables table `raw'", leaving --network host as the
+# only usable mode.
+CONFIG_IP_NF_RAW=m
```
nicksinas, line 19: **#5 — Medium. The docker-28+ raw-table breakage isn't NXP-specific, but the fix is.**

This fragment only reaches `linux-imx`, `linux-fslc-imx` and `linux-compulab` (`grep -rn avocado-netfilter` → three bbappends, all under `meta-avocado-nxp`). Meanwhile:

- `meta-avocado-raspberrypi/recipes-kernel/linux/files/avocado-extra.cfg:48` enables `CONFIG_IP_NF_IPTABLES` and no `IP_NF_RAW`
- `meta-avocado-renesas/recipes-kernel/linux/files/avocado-extra.cfg:23` — same
- `meta-avocado-nvidia`'s fragments have no netfilter section at all

So `docker run` fails on those boards with the same `can't initialize iptables table 'raw'`, leaving `--network host` as the only usable mode. Either add the two symbols to those fragments here, or open a follow-up and link it — as written the commit message reads like the class of bug is fixed when it's fixed on one vendor.

kas/vendor/nxp-frdm-imx93.yml
```diff
@@ -0,0 +1,107 @@
...
+    BBMASK += "meta-avocado/meta-avocado-nxp/recipes-multimedia/gstreamer/gstreamer1.0-plugins-imx_%.bbappend"
+
+    # nnstreamer has no provider at all in this layer set. meta-imx-ml shipped
+    # nnstreamer_2.2.0.bb on 6.6.36 but is append-only at 6.18
+    # (nnstreamer_2.4.2.bbappend), expecting a recipe from a layer Avocado does
+    # not carry; meta-openembedded has never provided one on either series. The
+    # append's purpose was to drop the tvm PACKAGECONFIG to dodge a conflict
+    # between meta-imx-ml's and meta-ros's tvm_0.7.0, and neither layer is in
+    # this build.
+    BBMASK += "meta-avocado/meta-avocado-nxp/recipes-nnstreamer/nnstreamer/nnstreamer_%.bbappend"
```
nicksinas: **#6 — Medium-low. These two `BBMASK`s are machine-scoped fixes for machine-agnostic breakage, by their own argument.**

The gstreamer comment says the mask/append pairing "is not new — the mask and the append have coexisted since the FRDM boards were added — so this append has been dangling for any machine requiring that include". The nnstreamer comment says meta-openembedded "has never provided one on either series".

If both are true, `imx91-frdm`, `imx93-evk`, `imx95-frdm` and `imx8mp-evk` hit the same aborts the moment their BSP moves, and they won't pick these up from an imx93-only kas file — so the next BSP move rediscovers them.

Suggest moving these two next to the existing `BBMASK +=` lines in `fsl-imx-preferred-env.inc:57-66`, where the layer already keeps its mask policy and where the machines that share the problem will see them. Keep only the `udev-rules-imx` mask here — that rename really is wrynose-specific, so machine scope is right for it.

meta-avocado/recipes-devtools/rust/rust_1.94.1.bbappend
```diff
@@ -0,0 +1,28 @@
...
+# simply has no path to it, which is the same shape as the libz gap in
+# rust-native's snapshot. Point the loader at the sysroot's libdir for the task
+# that runs the binary rather than patching a RUNPATH into a sysroot artifact,
+# because the binary is correct as shipped - it is only ever meant to run from
+# an installed SDK, where its libraries are on the default path.
+#
+# Scoped to class-nativesdk: the native and target variants never hit this,
+# since the native llvm-config has its libraries beside it and the target one is
+# not executed.
+do_install:prepend:class-nativesdk() {
```
nicksinas: **#7 — Medium-low. Version-pinned append sitting next to a `%` append that argues against version-pinning.**

`rust_%.bbappend:22-26` reasons explicitly that "nothing pins `PREFERRED_VERSION_rust`, and `kas/base.yml` contemplates dropping the mixins layer outright, so re-declare it here rather than inherit it by luck". The mixins layer is in fact already commented out (`base.yml:36-42`), so rust now comes from oe-core and moves with every oe-core bump.

Which means the next bump silently drops this `LD_LIBRARY_PATH` fix, and `nativesdk-rust` starts dying at `llvm-config: libzstd.so.1` again with nothing pointing back here.

Suggest folding it into `rust_%.bbappend` — the `class-nativesdk` guard already scopes it correctly and the `do_install` shell body it prepends to is stable across versions. If it genuinely has to be version-pinned, worth a line saying why this one must be when the sibling must not.

meta-avocado-shared/recipes-connectivity/ganesha/ganesha_6.5.bb
```diff
@@ -42,6 +47,7 @@ EXTRA_OECMAKE = "\
     -DUSE_SYSTEM_NTIRPC=ON \
     -DUSE_GSS=OFF \
     -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
+    -DUSE_FSAL_GLUSTER=OFF \
```
nicksinas: **#10 — Low. Undocumented and unscoped, in a PR where every other line has a paragraph.**

`-DUSE_FSAL_GLUSTER=OFF` lands in the shared `EXTRA_OECMAKE`, so it changes the target and native builds too — not just the `nativesdk` case the `PACKAGECONFIG:remove:class-nativesdk` five lines up addresses.

If this is another host-header leak (gluster headers present on the build host), say so and scope it the same way. If it's a deliberate feature decision, it belongs as a `PACKAGECONFIG[gluster]` entry alongside the eight already at `:26-33` rather than a bare `OFF` in the shared block.

meta-avocado-nxp/recipes-avocado/packagegroups/packagegroup-avocado-imx-extra.bb
```diff
@@ -11,11 +11,55 @@ NXP_WIFI_DRIVER = " \
...
+# the only reachable content behind them is -nxp8997-common's two Bluetooth
+# files (uartuart8997_bt_v4.bin, helper_uart_3000000.bin). The blob the driver
+# loads, mrvl/sdiouart8997_combo_v4.bin, is packaged only by meta-freescale, in
+# the package meta-imx removes above - so pulling the shims in would install
+# two Bluetooth files and read like WiFi coverage that is not there.
+#
+# Which part to target is moot until a module is fitted: this unit's M.2 slot
+# is empty, and the FRDM-IMX93's supported module is an 88W8987, a third answer
+# again.
+NXP_WIFI_FIRMWARE:avocado-imx93-frdm = ""
```
nicksinas, line 68: **#11 — Low. `NXP_WIFI_DRIVER` is left in place for this machine while all its firmware is cleared.**

33 lines explain why no firmware is named here, but `kernel-module-nxp-wlan` still reaches the board through `RDEPENDS:${PN}` at `:63-67`, and meta-imx already recommends it for both AW693 features anyway (`layer.conf:109-110`). So `imx93-frdm` ends up with the driver declared twice and no firmware at all.

Worth one line either way: is the driver intentionally kept for a future AW693/8987 module, or should it be cleared alongside the firmware for the same reason?

meta-avocado-nxp/conf/machine/include/fsl-imx-preferred-env.inc
```diff
@@ -27,6 +27,25 @@ PREFERRED_VERSION_gstreamer1.0-plugins-ugly:mx9-nxp-bsp = "1.24.0.imx"
...
+PREFERRED_VERSION_gstreamer1.0:avocado-imx93-frdm              = ""
+PREFERRED_VERSION_gstreamer1.0-plugins-base:avocado-imx93-frdm = ""
+PREFERRED_VERSION_gstreamer1.0-plugins-good:avocado-imx93-frdm = ""
+PREFERRED_VERSION_gstreamer1.0-plugins-bad:avocado-imx93-frdm  = ""
+PREFERRED_VERSION_gstreamer1.0-plugins-ugly:avocado-imx93-frdm = ""
+PREFERRED_VERSION_gstreamer1.0-libav:avocado-imx93-frdm        = ""
+PREFERRED_VERSION_gstreamer1.0-rtsp-server:avocado-imx93-frdm  = ""
+
 PREFERRED_VERSION_xen-tools = "4.18.imx+stable"
 PREFERRED_VERSION_xen = "4.18.imx+stable"
```
nicksinas, line 50: **#12 — Low. The stale-pin cleanup stops one section short.**

The rationale for the block just above is that a pin which matches nothing "reads as authoritative while deciding nothing, and the next person to open this file would believe `1.24.0.imx` is what ships". That argument applies unchanged to `PREFERRED_VERSION_xen{,-tools} = "4.18.imx+stable"` here and `QEMUVERSION:mx93-nxp-bsp = "8.2.2.imx"` at `:54`, which almost certainly name versions the wrynose meta-imx no longer ships.

Either clear them for `avocado-imx93-frdm` in the same block, or note in the new comment that they were checked and are still live at 6.18 — otherwise the file is half-cleaned and the next reader can't tell which half.

meta-avocado/recipes-avocado/images/avocado-image-rootfs.bb
jetm: Confirmed, and worse than described: clearing `SSHD_OPTS` also drops the `-f sshd_config_readonly` flag oe-core's own hook had set, so sshd falls all the way back to `/etc/ssh/sshd_config` on an erofs rootfs. Verified against the pinned oe-core commit's `openssh_10.3p1.bb` and `read_only_rootfs_hook`. Replaced the hook with `OPENSSH_HOST_KEY_DIR_READONLY_CONFIG = "/var/lib/ssh"` in the distro conf, exactly as suggested, and deleted the broken `persist_ssh_host_keys` function.

meta-avocado-distro/recipes-devtools/cmake/cmake-native_%.bbappend
jetm: Confirmed at the pinned oe-core commit (`cmake-native_4.3.1.bb:34` carries `-DCMAKE_DISABLE_FIND_PACKAGE_Libidn2=ON`). Dropped the duplicate line and its comment, kept the nghttp2 opt-out, which has no upstream counterpart.

meta-avocado/recipes-core/glib-2.0/glib-2.0_%.bbappend
jetm: Confirmed for all three: `meta-avocado-distro` already carries glib-2.0 and libpcre2 bbappends with the identical flags and both layers land in this machine's build. Dropped the `meta-avocado` duplicates and moved zlib's (the one genuinely new file) next to its siblings in `meta-avocado-distro` with matching rationale.

meta-avocado-distro/recipes-core/systemd/systemd/wait-online-any.conf
jetm: Fixed (a): the drop-in now installs a placeholder and `do_install:append` substitutes it with `${rootlibexecdir}` via `sed`, so the usrmerge answer is derived rather than baked in. On (b), the distro-wide `--any` policy scope: leaving as landed here rather than splitting it out - that's a call for a human, raised separately.

meta-avocado-nxp/recipes-kernel/linux/files/avocado-netfilter.cfg
jetm: Confirmed: raspberrypi and renesas kernel fragments enable `IP_NF_IPTABLES` but not `IP_NF_RAW`. Added `CONFIG_IP_NF_RAW=m` to both, matching the NXP fragment. Left `IP6_NF_RAW` out since neither fragment has any IPv6 netfilter config to pair it with. nvidia's fragments have no netfilter section at all, so nothing to add there.

kas/vendor/nxp-frdm-imx93.yml
jetm: Agreed and moved: the gstreamer and nnstreamer masks now live in `fsl-imx-preferred-env.inc`, verified via git blame that the gstreamer mask predates FRDM board support entirely, and that the nnstreamer PACKAGECONFIG dodge is scoped to `mx8mp-nxp-bsp` (a no-op everywhere `fsl-imx-preferred-env.inc` reaches). Kept the udev-rules-imx mask in the imx93 kas file, which really is wrynose-specific.

meta-avocado/recipes-devtools/rust/rust_1.94.1.bbappend
jetm: Agreed. Folded the `do_install:prepend:class-nativesdk` llvm-config fix into `rust_%.bbappend` and deleted the version-pinned file, so it now tracks whatever rust version is current the same way the sibling libz.so.1 fix does.

meta-avocado-shared/recipes-connectivity/ganesha/ganesha_6.5.bb
jetm: Traced it: Avocado carries no glusterfs-api recipe anywhere, so the FSAL's own default-ON probe would just fail cleanly via pkg-config (sandboxed, confirmed against upstream's CMakeLists.txt) - not a host-header leak like the dbus/pam cases nearby. Forcing it off skips a doomed probe, which is a real if minor reason to keep it off. Reworked as `PACKAGECONFIG[gluster]`, off by default, matching this recipe's other FSAL toggles, so it's documented and scoped correctly rather than a bare flag in the shared block.

meta-avocado-nxp/recipes-avocado/packagegroups/packagegroup-avocado-imx-extra.bb
jetm: Kept intentionally, documented now: `kernel-module-nxp-wlan` is the same module for every NXP WiFi chip meta-imx-bsp knows, and meta-imx-bsp's own `layer.conf` already soft-recommends it once `nxpaw693-sdio` lands in `MACHINE_FEATURES` for this SoC. Keeping the explicit RDEPENDS makes it a firm guarantee rather than relying on that recommendation, for the day a module is fitted to this unit's empty M.2 slot.

meta-avocado-nxp/conf/machine/include/fsl-imx-preferred-env.inc
jetm: Partially confirmed. `QEMUVERSION:mx93-nxp-bsp = "8.2.2.imx"` is the same shape as the gstreamer pins already cleared - wrynose's meta-imx ships `qemu_10.2.5.imx.bb`, not `8.2.2.imx` - so it's cleared the same way, scoped to the machine. `PREFERRED_VERSION_xen{,-tools}`, though, isn't 6.18-specific staleness at all: the only xen recipes in the entire layer set live under `meta-imx-sdk/dynamic-layers/virtualization-layer/`, and `meta-imx-sdk` has never been part of any Avocado NXP kas config, at any BSP. That pin has been fully inert since it was introduced, independent of this PR's BSP bump, so a machine-scoped clear here would misstate the reason. Left untouched pending that being confirmed dead code rather than folded into this cleanup.

meta-avocado-distro/recipes-core/systemd/systemd/wait-online-any.conf
jetm: Confirmed with Javier: leaving the `--any` policy change in this PR rather than splitting it out. Consistent with the other repo-wide fixes already landed here during imx93 bring-up; `RequiredForOnline=` on specific interfaces remains the escape hatch for a board that needs different semantics later.

meta-avocado-nxp/conf/machine/include/fsl-imx-preferred-env.inc
jetm: Correction to my last reply here: the reason I gave was wrong on a checkable
point, even though the conclusion stands.

I said `meta-imx-sdk` was the only layer providing a xen recipe, and that it
isn't in any Avocado NXP kas config. The second half is true, but the first
half isn't: `kas/vendor/nxp.yml` includes `kas/feature/virtualization.yml`
unconditionally, which pulls in `meta-virtualization` - and that layer does
ship `xen_{4.19,4.20,4.21,git}.bb` and matching `xen-tools` recipes, in every
Avocado NXP build including this one.

The pin is still dead, for a different reason: nothing in this tree sets
`DISTRO_FEATURES` or `PACKAGE_INSTALL` to pull in xen, and meta-virtualization's
xen recipes all carry `DEFAULT_PREFERENCE ??= "-1"` with no
`REQUIRED_DISTRO_FEATURES` gate, so they're never selected regardless of what
`PREFERRED_VERSION_xen` says. The variable only matters once something makes
xen a build candidate, and nothing here does.

Worth noting since it changes the risk if that ever stops being true: wrynose's
meta-imx ships `xen_4.21.imx.bb`, not the pinned `4.18.imx+stable` - so if xen
were ever enabled on a FRDM board, this line would select a version present in
neither meta-imx-sdk (not in the layer set) nor meta-virtualization (doesn't
ship 4.18), and the build would fail on a missing recipe rather than silently
picking the wrong one. That's a real argument for clearing it at some point,
just still not this PR's reason to touch it - the pin predates this BSP bump
(`7deb37dd`, May 2025, copied verbatim from NXP's own preferred-versions file
alongside a layer this tree never enabled) and has matched nothing since it was
added.

mobileoverlord (approved):
Reviewed in full. The BSP move is the only viable path — scarthgap's SPL stalls pre-`Normal Boot` on Rev1.2 silicon and NXP's own image reproduces it — and the pin split (own vendor file, meta-imx-ml disabled with both parse failures documented, dangling appends masked with the two-line rules delta accounted for) is scoped so imx91/95 stay untouched on 6.6.36. Verified `stone-uuu-emmc` is a real synthesized override (avocado.inc's STONE_PROVISIONING → MACHINEOVERRIDES), so the libubootenv-native dep is live. The distro-wide pieces (extension-service initrd race fix, wait-online --any, host-contamination fixes, docker raw-table configs, /var/lib/ssh host keys) all hold up and are the kind of thing every target wants.

Housekeeping done on the branch: rebased onto current wrynose. The rust_%.bbappend add/add conflict with #300 (which landed an hour ago) resolved to this branch's version — it's the variant written for wrynose's rust 1.94.1 and additionally carries the nativesdk llvm-config fix, so it supersedes the ported copy cleanly.

PR #262, opened 2026-08-05
Stop cmake-native and rust-native picking up host libraries

jetm:
> **This blocks any cold build, on any machine.** It reads like cleanup and is not.
> A fresh `avocado-qemuarm64` build failed on `rust-native do_install` at
> 6,887/12,092 tasks on 2026-08-06 with exactly the `libz.so.1` error below.
> Detail in "Why this is blocking now".

## Problem

Two native recipes pick up libraries from the build host instead of the native
sysroot, and both fail late and opaquely on a host that has those libraries
installed.

**cmake-native** - the bundled cmcurl probes for libidn2 with `check_library_exists`,
which searches the default linker path rather than `recipe-sysroot-native`. On a host
that ships libidn2 and its header in the base system the probe succeeds and `-lidn2`
reaches five link lines including cmake's own, but nothing stages libidn2 into the
native sysroot, so RUNPATH never covers it:

    bin/cmake: error while loading shared libraries: libidn2.so.0

**rust-native** - `do_rust_setup_snapshot` repoints the prebuilt snapshot's three
binaries at the uninative loader but sets no rpath, and never touches the libraries
beside them. Those carry `RUNPATH=$ORIGIN`, and libLLVM needs `libz.so.1`, which the
uninative tarball does not ship (`uninative-tarball.bb` packages glibc, patchelf,
libxcrypt, libnss-nis and libgcc, nothing more):

    rust-snapshot/bin/rustc: error while loading shared libraries: libz.so.1

Both surface as `oe_runmake failed` or a bare exit 127, with the real error dozens of
lines above. Whether either recipe links the host library depends on what the builder
happens to have installed, so these are reproducibility problems as much as build
failures.

## Solution

Carry both as bbappends in `meta-avocado` rather than patching the vendored recipes -
the vendor forks exist to pin hashes, so layer-level fixes belong here. This also
matches the neighbouring `qemu-system-native_%.bbappend`, which contains the same class
of host auto-detection for the same reproducibility reason.

**cmake-native** - seed the `HAVE_` probe results. `CMAKE_DISABLE_FIND_PACKAGE_Libidn2`,
oe-core's fix for 4.3.x, is deliberately not reused: it disables a `find_package(Libidn2)`
call the bundled cmcurl only grew in the 3.28.x -> 4.3.x upgrade, so against the pinned
3.28.3 it has nothing to disable. `USE_LIBIDN2=0` alone is also insufficient, because
`check_library_exists` runs independently of the feature switch and still leaves
`HAVE_LIBIDN2=1` in the cache.

**rust-native** - append `STAGING_LIBDIR_NATIVE` after the existing `$ORIGIN` entries so
a library the snapshot ships still wins over a staged one of the same name. This is the
rpath the recipe already hands `RUST_ALTERNATE_EXE_PATH`; the snapshot was the one
consumer left without it. `zlib-native` is already staged, so nothing needs building -
the loader simply had no path to it.

## Key changes

- `meta-avocado/recipes-devtools/cmake/cmake-native_%.bbappend`: add
  `-DUSE_LIBIDN2=0 -DHAVE_LIBIDN2=0 -DHAVE_IDN2_H=0` to `CMAKE_EXTRACONF`
- `meta-avocado/recipes-devtools/rust/rust_%.bbappend`: rpath the snapshot binaries and
  libraries at the native sysroot via `do_rust_setup_snapshot:append`
- Probe each snapshot library with `patchelf --print-rpath` before patching, since
  `lib/libLLVM-*.so` is a 42-byte ld script rather than an ELF and patchelf exits
  non-zero on it

## Reviewer notes

Verified on qemuarm64 with both appends and no other overrides in play:

- `cmake-native` rebuilds from `cleansstate`; `readelf -d bin/cmake | grep -ci idn2`
  goes 1 -> 0, and `HAVE_LIBIDN2:UNINITIALIZED=0` confirms the probe is skipped rather
  than merely overridden
- `rust-native` rebuilds from `cleansstate` (416/416 tasks, 0 errors) where it
  previously failed in `do_install`; `rustc` and `libLLVM` both come out with
  `$ORIGIN...:.../recipe-sysroot-native/usr/lib`

Each changes its recipe's task signature, so both rebuild once.

The rust gap is latent rather than newly introduced - it fires only when a snapshot
links a library uninative does not carry, which the 1.91.0 snapshot does via libLLVM.
oe-core master has the same gap. Upstream fixed the cmake one for 4.3.x and later only,
leaving the 3.28.x LTS branch exposed; the cmake append is annotated to be dropped on
the 4.3.x bump, which carries the upstream fix and renames `CMAKE_EXTRACONF` to
`EXTRA_OECMAKE`.

## Why this is blocking now

"Latent" turned out to mean "masked by sstate", not "rare".

A fresh `avocado-qemuarm64` build without this PR failed on 2026-08-06 at
6,887/12,092 tasks:

    rust-snapshot/bin/rustc: error while loading shared libraries: libz.so.1
    ERROR: rust-native-1.94.1-r0 do_install: ... failed with exit code 1

The reason no one has hit it is that existing `rust-native` sstate artifacts were
built during a window when the local `yocto-uninative-tarball` package still shipped
`libz` - added in `af47fc2` ("Ship libz for rust-native's bootstrap rustc"), removed
again in `726428e` ("Drop libz again, it is OE's to stage"). Removing it was correct:
shipping the host's zlib (`zlib-ng-compat` on Arch) would substitute it for the zlib
OE built, for every native binary without a RUNPATH, which is the contamination
uninative exists to prevent. But it left this PR as the only remaining fix, and any
build that has to rebuild `rust-native` rather than restore it from sstate now fails.

A build at 98% sstate does not notice. A build at 24% does.

### The dependency chain, for the record

`libz` is not in `rustc`'s own `DT_NEEDED`; it is reached transitively, which is why
patching only the binaries would not have been enough:

    rustc -> librustc_driver-*.so -> libLLVM.so.21.1-* -> libz.so.1

All three carry `RUNPATH=$ORIGIN/../lib`, which contains no libz, while `zlib-native`
is staged in `recipe-sysroot-native/usr/lib` one directory away. The append's loop over
`rust-snapshot/lib/*.so*` is what closes it - the binary-only half would not have.

### Verified today

With this branch's `rust_%.bbappend` applied to an otherwise unmodified tree, and after
removing a hand-made symlink used during diagnosis so the append had to do the work
itself:

    RUNPATH: [$ORIGIN/../lib:.../rust-native/1.94.1/recipe-sysroot-native/usr/lib]
    rustc 1.93.0 (254b59607 2026-01-19)
    libz.so.1 => .../recipe-sysroot-native/usr/lib/libz.so.1

The resolved path is OE's staged zlib inside the build tree, not the host's - which is
the property that makes this the right fix rather than re-adding libz to uninative.


meta-avocado/recipes-devtools/cmake/cmake-native_%.bbappend
```diff
@@ -0,0 +1,24 @@
...
+# Seed the HAVE_ results rather than using CMAKE_DISABLE_FIND_PACKAGE_Libidn2
+# (oe-core's fix for 4.3.x): that knob disables a find_package(Libidn2) call the
+# bundled cmcurl only grew in the 3.28.x -> 4.3.x upgrade, so on 3.28.x it has
+# nothing to disable. USE_LIBIDN2 alone is not enough either - check_library_exists
+# runs whether or not the feature switch is set, leaving HAVE_LIBIDN2=1 in the
+# cache and -lidn2 on the link line.
+#
+# Drop this append when cmake moves to 4.3.x or later: that version carries the
+# upstream fix, and renamed CMAKE_EXTRACONF to EXTRA_OECMAKE.
+CMAKE_EXTRACONF += "-DUSE_LIBIDN2=0 -DHAVE_LIBIDN2=0 -DHAVE_IDN2_H=0"
```
jetm: **Append silently no-ops when cmake reaches 4.3.x, and one seed is dead**

Two problems on this line.

**`CMAKE_EXTRACONF` disappears at 4.3.x while `_%` keeps matching.** oe-core's cmake-native 4.3.1 passes `-- ${EXTRA_OECMAKE}` to bootstrap (`cmake-native_4.3.1.bb:45`); `CMAKE_EXTRACONF` appears nowhere in `meta/recipes-devtools/cmake/` on wrynose `06dd66e622`. So after the bump this append parses cleanly, sets a variable nobody reads, emits no warning, and on any builder with libidn2 in the base system `-lidn2` returns to the link line - the exact failure this fixes, now with a file in the tree that looks like it is preventing it. Pinning the filename to `cmake-native_3.28.%.bbappend` turns that into a hard dangling-append error at the upgrade instead of silence.

**`-DUSE_LIBIDN2=0` cannot take effect.** cmcurl does `set(USE_LIBIDN2 ON)` as a normal (non-cache) variable, which shadows anything passed with `-D`. That is precisely why oe-core carries `0001-CMakeLists.txt-disable-USE_NGHTTP2.patch` for the adjacent `set(USE_NGHTTP2 ON)`. The comment at :18-20 reaches the right conclusion - "USE_LIBIDN2 alone is not enough" - for the wrong reason, and then keeps the inert flag. `HAVE_LIBIDN2=0` is the load-bearing seed; `HAVE_IDN2_H=0` is defensible for cache consistency.

meta-avocado/recipes-devtools/cmake/cmake-native_%.bbappend
```diff
@@ -0,0 +1,24 @@
...
+# result the same across build machines.
+#
+# Seed the HAVE_ results rather than using CMAKE_DISABLE_FIND_PACKAGE_Libidn2
+# (oe-core's fix for 4.3.x): that knob disables a find_package(Libidn2) call the
+# bundled cmcurl only grew in the 3.28.x -> 4.3.x upgrade, so on 3.28.x it has
+# nothing to disable. USE_LIBIDN2 alone is not enough either - check_library_exists
+# runs whether or not the feature switch is set, leaving HAVE_LIBIDN2=1 in the
+# cache and -lidn2 on the link line.
+#
+# Drop this append when cmake moves to 4.3.x or later: that version carries the
```
jetm: **"Drop at 4.3.x, oe-core has the fix" is unsupported**

This tells the next maintainer to delete the append at cmake 4.3.x because oe-core carries a `CMAKE_DISABLE_FIND_PACKAGE_Libidn2` fix there. `grep -rn -i idn2 meta/` on oe-core wrynose returns no cmake-related hit at all.

`CMAKE_DISABLE_FIND_PACKAGE_<pkg>` is also a stock CMake variable, not an oe-core knob, so the parenthetical attribution is wrong regardless of the version question.

Failure: whoever bumps cmake follows this line and deletes the append. If 4.3.x's bundled cmcurl still probes with `check_library_exists`, host libidn2 is back - and the deletion looks sanctioned by the file itself.

Settling the rest needs cmake 4.3.x's `Utilities/cmcurl/CMakeLists.txt`, which is not in either local clone. Either read it and cite it, or reword to "verify before dropping".

meta-avocado/recipes-devtools/rust/rust_%.bbappend
```diff
@@ -0,0 +1,33 @@
...
+        for bin in cargo rustc rustdoc; do
+            patchelf ${WORKDIR}/rust-snapshot/bin/$bin \
+                --set-rpath \$ORIGIN/../lib:${STAGING_LIBDIR_NATIVE}
+        done
+        for lib in ${WORKDIR}/rust-snapshot/lib/*.so*; do
+            # lib/libLLVM-*.so is a 42-byte ld script rather than an ELF, and
+            # patchelf exits non-zero on it. Use patchelf as its own ELF probe so
+            # such stubs are skipped without masking a real failure below.
+            patchelf "$lib" --print-rpath >/dev/null 2>&1 || continue
+            patchelf "$lib" --set-rpath \$ORIGIN:${STAGING_LIBDIR_NATIVE}
```
jetm: **RUNPATH is overwritten, not appended as the comment claims**

Line 29 reads the existing RUNPATH and throws it away - `>/dev/null`, used only as an ELF probe - and line 30 then asserts a hardcoded `\$ORIGIN:${STAGING_LIBDIR_NATIVE}`. So the "append after the existing `$ORIGIN` entries" behaviour promised by the comment at :12-13 and by the commit body is not what the code does.

Failure: any `rust-snapshot/lib/*.so*` whose RUNPATH is not exactly `$ORIGIN` - `$ORIGIN/../lib`, or a two-entry runpath - loses those entries, and a sibling library it used to resolve becomes unreachable. That produces a second `error while loading shared libraries` one layer deeper than the one being fixed.

Related coverage gap in the same loop: the glob only reaches the top-level `lib/`, not `lib/rustlib/*/lib/*.so*` or `libexec/`.

Needs `readelf -d` on the actual 1.92 snapshot to say which libraries are affected - could not settle it statically.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
```diff
@@ -0,0 +1,33 @@
...
+# after the existing $ORIGIN entries so a library the snapshot ships still wins
+# over a staged one of the same name. This is the rpath the recipe already hands
+# RUST_ALTERNATE_EXE_PATH; the snapshot was the one consumer left without it.
+#
+# Guarded on UNINATIVE_LOADER like the task it extends: without uninative the
+# host loader resolves the library and there is nothing to fix.
+do_rust_setup_snapshot:append () {
+    if [ ! -z "${UNINATIVE_LOADER}" -a -e "${UNINATIVE_LOADER}" ]; then
+        for bin in cargo rustc rustdoc; do
+            patchelf ${WORKDIR}/rust-snapshot/bin/$bin \
```
jetm: **Uses `patchelf` without declaring `patchelf-native`**

This calls bare `patchelf`, which is available only because the *recipe* in another layer declares `do_rust_setup_snapshot[depends] += "patchelf-native:do_populate_sysroot"`. The bbappend declares nothing itself.

`patchelf` is not in `HOSTTOOLS` or `HOSTTOOLS_NONFATAL` (oe-core scarthgap `meta/conf/bitbake.conf:516-535`), so it resolves only from `recipe-sysroot-native`. The flag exists at `meta-lts-mixins/recipes-devtools/rust/rust_1.92.0.bb:70` and oe-core wrynose `rust_1.94.1.bb:73` - but **not** in oe-core scarthgap `rust_1.75.0.bb`, whose task uses `patchelf-uninative` instead. `rust_%.bbappend` matches that recipe too, and there is no `PREFERRED_VERSION_rust` pin anywhere in meta-avocado.

Failure: drop the mixins layer while still on scarthgap oe-core - which `kas/base.yml` explicitly contemplates ("likely delete outright") - and `do_rust_setup_snapshot` dies with `patchelf: command not found`, exit 127, in the same opaque shape as the bug being fixed.

One line re-declaring the `[depends]` flag in the bbappend removes the cross-layer coupling.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
```diff
@@ -0,0 +1,33 @@
...
+#
+# zlib-native is already staged in recipe-sysroot-native, so nothing needs
+# building - the loader simply has no path to it. Append STAGING_LIBDIR_NATIVE
+# after the existing $ORIGIN entries so a library the snapshot ships still wins
+# over a staged one of the same name. This is the rpath the recipe already hands
+# RUST_ALTERNATE_EXE_PATH; the snapshot was the one consumer left without it.
+#
+# Guarded on UNINATIVE_LOADER like the task it extends: without uninative the
+# host loader resolves the library and there is nothing to fix.
+do_rust_setup_snapshot:append () {
```
jetm, line 65: **cargo's snapshot has the same exposure and is untouched**

`do_cargo_setup_snapshot` repoints the interpreter and sets no rpath - exactly like the rust task before this fix - and this PR only covers `rust_%`.

`meta-lts-mixins/recipes-devtools/rust/cargo_1.92.0.bb:29` patches only `bin/cargo`'s interpreter under the same `UNINATIVE_LOADER` guard, with no rpath and, unlike rust, no `patchelf-native` depends. If that snapshot binary links anything the uninative tarball does not ship, cargo-native fails identically.

Narrower than rust's case - the cargo snapshot installs `bin/cargo` with no `lib/*.so*` beside it - so this may be inert in practice. A `readelf -d` on the snapshot cargo decides it. Worth deciding rather than leaving as an implicit assumption, since the sibling was the whole reason to look.

jetm (commented):
Second read of my own change, cold. The mechanism is right and I want to say why before the findings, because it is the part worth keeping: seeding `HAVE_*` cache results is oe-core's own pattern in this very recipe - scarthgap's `cmake-native_3.28.3.bb` ships `-DENABLE_ACL=0 -DHAVE_ACL_LIBACL_H=0 -DHAVE_SYS_ACL_H=0`, and wrynose adds `-DHAVE_EXT2FS_EXT2_FS_H=0 -DHAVE_WORKING_EXT2_IOC_GETFLAGS=0` with the comment "Ensure e2fsprogs isn't found on the host". `-DCMAKE_LIBRARY_PATH` cannot substitute, because `check_library_exists` adds `-L` without restricting the linker's default path. And `STAGING_LIBDIR_NATIVE` is the right variable on the `-native` side, already referenced by the surrounding recipe, and `STAGING_DIR_HOST`/`TMPDIR`/`WORKDIR` are in `BB_HASHEXCLUDE_COMMON`, so no sstate-hash pollution.

Five findings inline. The two that matter:

- The cmake append will stop working silently at the wrynose bump. `CMAKE_EXTRACONF` does not exist in cmake-native 4.3.1, but `_%` keeps matching it, so the file sits in the tree looking like it prevents the bug while `-lidn2` comes back. Pinning the filename to `3.28.%` makes that a hard error instead.
- The rust append calls bare `patchelf` and declares no dependency on it - that only works because a recipe in *another* layer declares the `[depends]` flag, and oe-core scarthgap's `rust_1.75.0.bb` does not.

One convention point that did not fit an anchor: both files are the first oe-core bbappends in `meta-avocado`. Every existing oe-core or vendor override in this repo lives in `meta-avocado-distro` (~33 bbappends, `BBFILE_PRIORITY 10`, plus the `BBFILES_DYNAMIC` machinery); `meta-avocado` carries the distro conf, images, packagegroups and classes and had zero bbappends before this. Functionally harmless - both match `recipes-*/*/*.bbappend` in the layer conf and append priority does not affect whether an append applies - but the next person hunting a toolchain override greps `meta-avocado-distro` and misses these. `meta-avocado-distro/recipes-devtools/` is the sibling home.

Three things need a real build to settle and I am not claiming them either way: which snapshot libraries actually carry a non-`$ORIGIN` RUNPATH, whether the cargo snapshot binary needs a library uninative omits, and what cmake 4.3.x's bundled cmcurl does about libidn2.

meta-avocado/recipes-devtools/cmake/cmake-native_%.bbappend
jetm: Both confirmed, and the 4.3.x half turned out to be stronger than I could show when I raised it. `cmake-native/4.3.1` is unpacked in a wrynose build tree here, so I read the source instead of inferring it.

`Utilities/cmcurl/CMakeLists.txt` in 4.3.1 detects with `find_package(Libidn2 QUIET)` at :1408, and immediately before that sets both `HAVE_IDN2_H` and `HAVE_LIBIDN2` to `OFF` as normal variables (:1405-1406). So at 4.3.x the seeds are shadowed the same way `-DUSE_LIBIDN2=0` is shadowed here by `set(USE_LIBIDN2 ON)` at :101. The unpinned append would not merely stop being read, it could not work even if it were.

Renamed to `cmake-native_3.28.%.bbappend` so the bump is a dangling-append error, and dropped `-DUSE_LIBIDN2=0`, keeping `HAVE_LIBIDN2=0 HAVE_IDN2_H=0`. Rewrote the :18-20 comment to give the shadowing as the reason rather than the check-runs-anyway argument, which is true but is not why the flag is inert.

cde39ad.

meta-avocado/recipes-devtools/cmake/cmake-native_%.bbappend
jetm: Right on the attribution, and I could settle the version question after all: `cmake-native/4.3.1/sources/cmake-4.3.1/Utilities/cmcurl/CMakeLists.txt` is unpacked in the wrynose build tree here.

What it actually says: 4.3.1 replaced the `check_library_exists` probe with `find_package(Libidn2 QUIET)` (:1408). That is the substantive difference, and it is not about a fix oe-core carries. `check_library_exists` bypasses cmake's find-root confinement by asking the linker directly, which is what let the host copy in; `find_package` honours it. So the leak does close at 4.3.x, but for a reason the old comment did not give, and `CMAKE_DISABLE_FIND_PACKAGE_Libidn2` is a stock cmake variable rather than an oe-core knob, exactly as you said.

Rewrote the comment to state that, cite the line numbers, and drop the oe-core attribution. The "delete this at 4.3.x" instruction is gone: the filename is now pinned to `3.28.%`, so the bump raises a dangling append and whoever hits it decides with the source in front of them rather than following a line in a comment.

cde39ad.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
jetm: Confirmed with `readelf -d` on the 1.94.1 snapshot, and the measurement changes the severity in both directions.

Every object carries `$ORIGIN/../lib`, not `$ORIGIN`: `librustc_driver-*.so`, `libLLVM.so.21.1-*`, `lib/rustlib/*/lib/libstd-*.so`, `libexec/rust-analyzer-proc-macro-srv` and all three `bin/` binaries. So line 30 was overwriting rather than appending, as you said. But for a file that lives in `lib/`, `$ORIGIN` and `$ORIGIN/../lib` resolve to the same directory, so nothing is actually unreachable today. The bug is latent, not live, and I would rather say that than claim a breakage I could not produce.

Fixed it anyway, because the property holding is about where these files happen to sit rather than about the code. The loop now reads the entry back and appends to it:

    existing=$(patchelf "$lib" --print-rpath 2>/dev/null) || continue

Verified against real snapshot libraries rather than by reading: copied `librustc_driver` and `libLLVM.so.21.1` out and ran the loop over them. New behaviour gives `[$ORIGIN/../lib:/FAKE/staging/usr/lib]`; the old line on the same input gives `[$ORIGIN:/FAKE/staging/usr/lib]`, so the entry is demonstrably preserved now. Also re-ran the loop under `sh -e` with the 42-byte `libLLVM-21-*.so` ld script in place, since bitbake runs tasks with `set -e` and the skip depends on `|| continue` surviving that. It completes and skips the stub.

On the coverage gap: left as-is, deliberately, with the reasoning now recorded in the file rather than in my head. `readelf -d` puts `lib/rustlib/*/lib/libstd-*.so` NEEDED entirely inside what uninative ships, and `libexec/rust-analyzer-proc-macro-srv` resolves `librustc_driver` through its own `$ORIGIN/../lib` into `lib/`, reaching the libz edge via libLLVM's RUNPATH, which this loop fixes. Widening the glob would have added churn without adding reach.

b8ca892.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
jetm: Agreed, and the scarthgap half is the part that makes it worth a line rather than a note. `rust_%` matches oe-core scarthgap `rust_1.75.0.bb` too, that recipe patches with `patchelf-uninative` and declares no `patchelf-native` depends, and nothing here pins `PREFERRED_VERSION_rust`. So the coupling is not just untidy, it is load-bearing on a layer `kas/base.yml` says it expects to drop.

Added the flag to the bbappend:

    do_rust_setup_snapshot[depends] += "patchelf-native:do_populate_sysroot"

`+=` rather than an assignment so it composes with the recipe-side flag where one already exists (meta-lts-mixins `rust_1.92.0.bb:70`, oe-core wrynose `rust_1.94.1.bb:73`) instead of racing it.

Not verified by a build: this needs `bitbake -c cleansstate rust-native && bitbake rust-native` against a scarthgap tree with the mixins layer removed, which is the configuration that does not exist here to build in. A compile in the current configuration would pass either way and prove nothing, so I am calling this reasoned from `bitbake.conf:516-535` and the three recipes, not measured.

b8ca892.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
jetm: Settled with `readelf -d`, and it comes out inert: no change needed for cargo.

The cargo-native 1.94.1 workdir ships the snapshot as `cargo-1.93.0-x86_64-unknown-linux-gnu/bin/cargo` with no `*.so*` anywhere in the tree beside it. Its NEEDED set is `libdl.so.2`, `libgcc_s.so.1`, `librt.so.1`, `libpthread.so.0`, `libm.so.6`, `libc.so.6`, `ld-linux-x86-64.so.2`, all of which the uninative tarball ships (glibc plus libgcc). RUNPATH is `$ORIGIN/../lib`. So there is nothing for an rpath to reach that the loader cannot already find, which is why `do_cargo_setup_snapshot` gets away with repointing the interpreter and stopping there.

Your instinct that it was narrower than rust's case was right, and the reason is the one you guessed: rust's exposure comes entirely from `libLLVM` needing `libz.so.1`, and the cargo snapshot has no libLLVM.

Recorded it in the bbappend rather than closing it silently, so the next person reading the loop sees cargo was checked and excluded rather than missed.

b8ca892.

meta-avocado/recipes-devtools/cmake/cmake-native_3.28.%.bbappend
```diff
@@ -0,0 +1,34 @@
...
+# unconditionally sets HAVE_IDN2_H and HAVE_LIBIDN2 to OFF as normal variables
+# first (:1405-1406), so these seeds are shadowed there exactly as -DUSE_LIBIDN2
+# is here. find_package also honours cmake's find-root confinement, which
+# check_library_exists bypasses by asking the linker directly - that is the
+# difference that closes the host leak at 4.3.x, and CMAKE_DISABLE_FIND_PACKAGE_Libidn2
+# (a stock cmake variable, not an oe-core knob) is the lever if it ever reopens.
+# A _% filename would keep matching after the bump and leave an append that
+# parses, sets nothing readable, warns about nothing, and looks like it is still
+# preventing the bug. Pinned, the same bump is a dangling-append error instead.
+CMAKE_EXTRACONF += "-DHAVE_LIBIDN2=0 -DHAVE_IDN2_H=0"
```
jetm, line 44: **The version pin excludes the release that actually needs this, and becomes a parse error on wrynose**

Two problems from one decision, and the second is the worse one.

**The pin misses 4.3.x.** The `3.28.%` scope rests on 4.3.x being immune. Upstream's own `854e940580d cmake-native: prevent host libidn2 contamination` proves it is not, and that commit is **not** an ancestor of the wrynose oe-core pin this repo already uses: `origin/wrynose` pins `vendor-openembedded-core@709c4fd69a3`, whose `cmake-native_4.3.1.bb` carries no `-DCMAKE_DISABLE_FIND_PACKAGE_Libidn2=ON` - I read the file. So a builder with libidn2 headers present builds any wrynose-pinned config and cmake-native fails with upstream's reported `fatal error: idn2.h: No such file or directory`.

**And the file becomes fatal there.** Merging `scarthgap` into `feat/wrynose-migration` - `LAYERSERIES_COMPAT_meta-avocado = "scarthgap wrynose"`, same 4.3.1 pin - makes this a dangling bbappend, which is `bb.fatal` at parse (`bitbake/lib/bb/cooker.py:957`; nothing here sets `BB_DANGLINGAPPENDS_WARNONLY`). The obvious conflict resolution is to delete the file, which leaves wrynose with no guard at all - so the failure mode is a parse error that gets 'fixed' by removing the protection.

Widening to `cmake-native_%.bbappend` and seeding both releases addresses both at once.

meta-avocado/recipes-devtools/cmake/cmake-native_3.28.%.bbappend
```diff
@@ -0,0 +1,34 @@
...
+# 0001-CMakeLists.txt-disable-USE_NGHTTP2.patch for on the adjacent
+# set(USE_NGHTTP2 ON). check_library_exists also runs whether or not the feature
+# switch is set, so even a switch that took would leave HAVE_LIBIDN2=1 in the
+# cache and -lidn2 on the link line.
+#
+# Filename pinned to 3.28.% on purpose. cmake 4.3.1 detects with
+# find_package(Libidn2 QUIET) instead (Utilities/cmcurl/CMakeLists.txt:1408) and
+# unconditionally sets HAVE_IDN2_H and HAVE_LIBIDN2 to OFF as normal variables
+# first (:1405-1406), so these seeds are shadowed there exactly as -DUSE_LIBIDN2
+# is here. find_package also honours cmake's find-root confinement, which
```
jetm: **The load-bearing premise here is contradicted verbatim by the upstream commit**

This comment argues that `find_package` honours cmake's find-root confinement where `check_library_exists` bypasses it, that this is what closes the leak at 4.3.x, and that `CMAKE_DISABLE_FIND_PACKAGE_Libidn2` is the lever if it ever reopens.

Upstream `854e940580d` says the opposite in as many words: cmcurl "calls `find_package(Libidn2)`. On hosts with libidn2-devel installed ... this **detects the host library during configure**" - and that knob is the fix they already merged, not a hypothetical lever.

Why it matters beyond being wrong: a maintainer generalises "`find_package` in a native recipe is host-safe", skips the audit at the next cmake bump, and drops this append believing the leak is structurally closed. There is nothing to confine anything - `cmake-native` does not `inherit cmake`, and `bootstrap` gets no toolchain file, so there is no `CMAKE_SYSROOT` or `CMAKE_FIND_ROOT_PATH` in play at all.

Worth reconsidering the whole shape while rewriting this. The layer now carries two one-off appends for one systemic cause - the sibling is `recipes-devtools/qemu/qemu-system-native_%.bbappend` (`DEPENDS += "keyutils-native"`), identical shape, identical trigger - and the real precondition is recorded nowhere: `avocado.inc:46` sets `SANITY_TESTED_DISTROS ?= "ubuntu-22.04 ubuntu-24.04 fedora-40"`, where `idn2.h` and the `libidn2.so` link symlink live in a `-dev`/`-devel` package absent by default, so this only fires on a build distro shipping dev headers in the base package. Two durable alternatives neither of us wrote down: `DEPENDS += "libidn2-native"` (`libidn2_2.3.7.bb` has `BBCLASSEXTEND = "native nativesdk"`, and `BUILD_LDFLAGS` already carries `-Wl,-rpath,${STAGING_LIBDIR_NATIVE}`, so a staged lib is reachable - the qemu precedent exactly, and it keeps IDN support), or naming the real build distro in `SANITY_TESTED_DISTROS` / building under buildtools so probes never see host headers.

meta-avocado/recipes-devtools/cmake/cmake-native_3.28.%.bbappend
```diff
@@ -0,0 +1,34 @@
...
+#
+#   bin/cmake: error while loading shared libraries: libidn2.so.0
+#
+# do_install then dies with exit 127, reported only as "oe_runmake failed" some
+# forty lines below the real error. Whether cmake links libidn2 at all depends on
+# what the builder happens to have installed, so seeding the probe keeps the
+# result the same across build machines.
+#
+# Seeding the HAVE_ results is what actually works here. USE_LIBIDN2 is not a
+# lever at all: cmcurl sets it as a normal variable (CMakeLists.txt:101), which
```
jetm: **Wrong line citation, and it points at the wrong file**

`set(USE_LIBIDN2 ON)` is at `Utilities/cmcurl/CMakeLists.txt:90`, not `CMakeLists.txt:101`. Proven from the hunk header of oe-core's own `0001-CMakeLists.txt-disable-USE_NGHTTP2.patch` (`@@ -88,7 +88,7 @@`): 88 `PICKY_COMPILER`, 89 `SHARE_LIB_OBJECT`, 90 `USE_LIBIDN2`, 91 `USE_NGHTTP2`. The patch is 1-for-1, so the numbering is identical before and after it applies.

The citation also drops the `Utilities/cmcurl/` prefix that the 4.3.1 citations two lines below do carry, so it reads as cmake's top-level `CMakeLists.txt`.

Concrete cost: someone opening `:101` to check the "USE_LIBIDN2 is not a lever" premise lands on unrelated code and either declares the comment stale or accepts the premise unverified - and per the thread above, that premise is the one that needs checking.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
```diff
@@ -0,0 +1,64 @@
...
+            patchelf ${WORKDIR}/rust-snapshot/bin/$bin \
+                --set-rpath \$ORIGIN/../lib:${STAGING_LIBDIR_NATIVE}
+        done
+        for lib in ${WORKDIR}/rust-snapshot/lib/*.so*; do
+            # lib/libLLVM-*.so is a 42-byte ld script rather than an ELF, and
+            # patchelf exits non-zero on it. Use patchelf as its own ELF probe so
+            # such stubs are skipped without masking a real failure below, and
+            # keep what it prints: these objects carry $ORIGIN/../lib, so
+            # asserting a bare $ORIGIN would drop an entry rather than add one.
+            existing=$(patchelf "$lib" --print-rpath 2>/dev/null) || continue
```
jetm: **`|| continue` turns the whole lib loop into a silent no-op on any failure**

`existing=$(patchelf "$lib" --print-rpath 2>/dev/null) || continue` uses patchelf's exit status as an ELF probe, so *any* failure is swallowed: patchelf absent, an unreadable file, or an unexpanded glob. That is the opposite of what the comment at 36-40 claims ("without masking a real failure below").

The reachable path: a future snapshot ships no top-level `lib/*.so*`. That is not hypothetical - mixins already carries `revert-link-std-statically-in-rustc_driver-feature.patch`, so upstream has a mode that empties that directory. POSIX sh then leaves `$lib` as the literal pattern, patchelf fails on ENOENT, `|| continue` hides it, the bin loop still succeeds, and the task exits 0. `do_configure` then fails with exactly the `rust-snapshot/bin/rustc: error while loading shared libraries: libz.so.1` this file exists to prevent, with nothing in the log pointing back here.

Probe for ELF-ness explicitly, or guard the glob with a `[ -e "$lib" ] || continue` and let a real patchelf failure be fatal.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
```diff
@@ -0,0 +1,64 @@
...
+            if [ -n "$existing" ]; then
+                patchelf "$lib" --set-rpath "$existing:${STAGING_LIBDIR_NATIVE}"
```
jetm: **The lib-loop append is not idempotent across a forced re-run**

It appends `:${STAGING_LIBDIR_NATIVE}` to whatever it reads, and the task has `[dirs]` rather than `[cleandirs]`, so nothing guarantees a clean `${WORKDIR}/rust-snapshot`.

`bitbake -f -c rust_setup_snapshot rust-native` over a surviving WORKDIR reads `$ORIGIN/../lib:<staging>` and writes `$ORIGIN/../lib:<staging>:<staging>` - one more copy per forced run, growing the dynstr and making a `-f` re-run non-equivalent to a clean one.

Unverified without a build: whether rust's `install.sh --prefix=` overwrites the patched libraries first, which would make it moot. A `case "$existing" in *"${STAGING_LIBDIR_NATIVE}"*) continue;; esac` settles it either way for one line.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
```diff
@@ -0,0 +1,64 @@
...
+# oe-core scarthgap rust_1.75.0, which patches with patchelf-uninative instead.
+# Nothing pins PREFERRED_VERSION_rust, and kas/base.yml contemplates dropping the
+# mixins layer outright, so re-declare it here rather than inherit it by luck.
+do_rust_setup_snapshot[depends] += "patchelf-native:do_populate_sysroot"
+
+do_rust_setup_snapshot:append () {
+    if [ ! -z "${UNINATIVE_LOADER}" -a -e "${UNINATIVE_LOADER}" ]; then
+        for bin in cargo rustc rustdoc; do
+            patchelf ${WORKDIR}/rust-snapshot/bin/$bin \
+                --set-rpath \$ORIGIN/../lib:${STAGING_LIBDIR_NATIVE}
```
jetm: **The bin loop asserts a fixed rpath while the lib loop reads-then-appends**

The two loops use different shapes for the same job, and the commit body argues the read-then-append one is correct precisely because a fixed value "is a property of where these files happen to sit, not of the code". The bin loop then sets a fixed `\$ORIGIN/../lib:${STAGING_LIBDIR_NATIVE}` anyway.

Failure: a future snapshot's `bin/rustc` carries a second RUNPATH entry - say `$ORIGIN/../lib/rustlib/x86_64-unknown-linux-gnu/lib` - this line overwrites it, that entry is lost, and rustc fails to start with the same opaque 127 the append exists to prevent.

Lower confidence: confirming it needs `readelf -d` on the fetched snapshot, which I did not fetch. But making both loops read-then-append costs nothing and removes the question.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
```diff
@@ -0,0 +1,64 @@
...
+# Guarded on UNINATIVE_LOADER like the task it extends: without uninative the
+# host loader resolves the library and there is nothing to fix.
+#
+# patchelf is in neither HOSTTOOLS nor HOSTTOOLS_NONFATAL, so it resolves only
+# out of recipe-sysroot-native. The recipe this appends to declares that depends
+# flag on meta-lts-mixins rust_1.92.0 and oe-core wrynose rust_1.94.1, but not on
+# oe-core scarthgap rust_1.75.0, which patches with patchelf-uninative instead.
+# Nothing pins PREFERRED_VERSION_rust, and kas/base.yml contemplates dropping the
+# mixins layer outright, so re-declare it here rather than inherit it by luck.
+do_rust_setup_snapshot[depends] += "patchelf-native:do_populate_sysroot"
```
jetm: **This `[depends]` is unnecessary, and the task-name binding it relies on can silently stop matching**

Two things about this line.

**It is not needed.** Inside this exact guard (`-e "${UNINATIVE_LOADER}"`), `uninative.bbclass:enable_uninative()` has already prepended the uninative tarball's bindir to `PATH`, and `uninative-tarball.bb:64` renames `usr/bin/patchelf` to `patchelf-uninative` there. So `patchelf-uninative` is guaranteed present with zero recipe dependency - both oe-core `rust_1.75.0.bb` and mixins `cargo_1.92.0.bb` already call it that way. As written this puts `patchelf-native:do_populate_sysroot` on rust's critical path for a tool the build already has. Switching the four call sites deletes this line and the comment block at 21-26. Honest tradeoff: it diverges from the mixins base task, which calls bare `patchelf`.

**And it fails quietly if it stops matching.** The cmake half argues for loud failure and earns it through the version pin. This half binds by task name under `rust_%`: if `do_rust_setup_snapshot` is ever renamed or dropped, BitBake creates a variable that is never in `__BBTASKS`, never runs, and never warns - and a `[depends]` varflag on a non-existent task is never read. The three candidate recipes already disagree about this task's internals (which patchelf binary, which installer directory, whether the depends is declared), so that churn is demonstrated rather than imagined. After a bump that renames it, the original opaque `rustc -vV` exit 127 comes back with an append sitting beside it that looks like it prevents exactly that.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
```diff
@@ -0,0 +1,64 @@
...
+# without it.
+#
+# Guarded on UNINATIVE_LOADER like the task it extends: without uninative the
+# host loader resolves the library and there is nothing to fix.
+#
+# patchelf is in neither HOSTTOOLS nor HOSTTOOLS_NONFATAL, so it resolves only
+# out of recipe-sysroot-native. The recipe this appends to declares that depends
+# flag on meta-lts-mixins rust_1.92.0 and oe-core wrynose rust_1.94.1, but not on
+# oe-core scarthgap rust_1.75.0, which patches with patchelf-uninative instead.
+# Nothing pins PREFERRED_VERSION_rust, and kas/base.yml contemplates dropping the
```
jetm: **This cites `kas/base.yml` for a claim base.yml does not make**

"kas/base.yml contemplates dropping the mixins layer outright" is not in that file. The `meta-lts-mixins` entry is a plain pinned repo with no comment. base.yml *does* annotate keep-rationale for meta-atmel, meta-erlang, meta-fwup, meta-swupdate and openjdk - which is probably where the impression came from.

This matters because of what it is holding up: line 27 duplicates a `[depends]` the in-use recipe already declares (mixins `rust_1.92.0.bb:70`), so this hedge is its only justification. A reviewer who checks base.yml finds nothing, and under the delete-and-test rule in `coding-discipline.md` the line then reads as unjustified and gets removed - silently reintroducing the exit-127 on the oe-core-only fallback.

Given the thread on line 27, the cleaner resolution is to drop the depends for the right reason rather than keep it for a wrong one.

jetm (commented):
Independent second read on my own PR, cold. Eight findings inline, and the first is the one that changes what this PR should look like.

**The `3.28.%` pin is scoped away from the release that needs it, and turns fatal on wrynose.** Upstream's `854e940580d` proves cmake 4.3.x has the identical libidn2 leak, and that commit is not an ancestor of the wrynose oe-core pin this repo already uses - I read `cmake-native_4.3.1.bb` at `vendor-openembedded-core@709c4fd69a3` and it carries no guard. Then merging `scarthgap` into `feat/wrynose-migration` makes this a dangling bbappend, which is `bb.fatal` at parse - and the natural way to clear that parse error is to delete the file, removing the protection entirely. Widening to `cmake-native_%.bbappend` and seeding both releases fixes both halves.

The comment block justifying that pin rests on a premise the same upstream commit contradicts word for word, so it needs rewriting rather than re-scoping - and while it is being rewritten, the shape is worth reconsidering. The layer now carries two one-off appends for one systemic cause (the sibling is `qemu-system-native_%.bbappend`, `DEPENDS += "keyutils-native"`, identical trigger), and the real precondition - a build distro that ships dev headers in its base package, which none of the three in `SANITY_TESTED_DISTROS` does - is written down nowhere. `DEPENDS += "libidn2-native"` works (checked: `BBCLASSEXTEND` covers native, and `BUILD_LDFLAGS` already carries the staging rpath) and keeps IDN support; naming the real build distro or building under buildtools is the other durable option. Neither is mentioned in the commit.

On the rust half, the `|| continue` is the one I would fix regardless of the rest: it uses patchelf's exit status as an ELF probe, so a missing `lib/*.so*` directory - which upstream already has a mode for - makes the loop silently patch nothing and the task exit 0, surfacing later as the exact `libz.so.1` failure the file exists to prevent.

Three of the remaining findings are self-inflicted accuracy problems worth owning: a wrong line citation that points at the wrong file, a comment citing `kas/base.yml` for a claim base.yml does not make, and a `[depends]` that is unnecessary because `patchelf-uninative` is already on `PATH` inside the very guard that wraps it.

**One thing I have not reviewed.** The head moved from `b8ca892b` to `1357106a` while this review was running, and the push added `meta-avocado-shared/recipes-connectivity/ganesha/ganesha_6.5.bb` - a new recipe unrelated to host-library contamination. The two bbappends are byte-identical between the two heads, so everything above still applies as written, but nothing here covers the ganesha recipe. It also raises a scope question: a new connectivity recipe does not obviously belong in a change about `cmake-native` and `rust-native` picking up host libraries, and splitting it out would make both halves easier to review and to revert.

For completeness, things I checked and am not raising: `UNPACKDIR`/`${WORKDIR}/rust-snapshot` is the right prefix on all three candidate recipes; `STAGING_LIBDIR_NATIVE` resolves through `WORKDIR`, which is in `BB_HASHEXCLUDE_COMMON`, so no TMPDIR path leaks into the task hash; the override syntax, `[depends] +=`, `CMAKE_EXTRACONF +=` and `\$ORIGIN` escaping are all correct for BitBake 2.8; file placement and the version-pinned append filename both match layer convention; and seeding both `HAVE_LIBIDN2` and `HAVE_IDN2_H` does reproduce a clean host byte-for-byte, with `do_configure[cleandirs]` preventing a stale cache.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
jetm: Fixed in `fa77b15`, and one of the three examples here turned out not to hold.

The core is real and now reproduces: with an empty `lib/`, the glob stays unexpanded, `patchelf` fails on ENOENT, `|| continue` swallows it and the task exits 0 with nothing in the log. I built a harness that runs the extracted task body against a stub `patchelf` and it failed exactly that way before the change.

The fix separates the two meanings that line was carrying. `command -v patchelf` runs once before either loop, so the per-object skip inside `rust_snapshot_ensure_rpath` now means "patchelf cannot read this object" and nothing else. The lib loop guards the glob with `[ -e "$lib" ] || continue` and tracks whether anything matched, warning when `lib/` holds no shared libraries rather than passing silently - a snapshot with none needs the re-audit, since the libz edge is reached through libLLVM's RUNPATH.

Where this comment was wrong: "patchelf absent" was not silently swallowed. BitBake runs shell tasks under `set -e` (`bitbake/lib/bb/build.py:377,381`), so the bin loop's unguarded call aborted the task with `patchelf: command not found` and exit 127 - loud, just with a poor diagnostic. I only found that after making the harness match bitbake's shell semantics; measuring without `set -e` had credited the loop with swallowing a failure the real runner surfaces. It now exits via `bbfatal` with a message naming why patchelf resolves only out of recipe-sysroot-native.

An unreadable file is still skipped rather than fatal, which is deliberate: that is the ld-script case.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
jetm: Confirmed and fixed in `fa77b15`. It reproduced before the change: two runs over a surviving WORKDIR left `$ORIGIN/../lib:<staging>:<staging>`, one more copy per forced run.

`rust_snapshot_ensure_rpath` now checks whether each entry is already present before appending, and returns without calling `patchelf` at all when the rpath would not change. So `bitbake -f -c rust_setup_snapshot` over a surviving `${WORKDIR}/rust-snapshot` is equivalent to a clean one.

I did not settle the question this comment raised about `install.sh --prefix=` overwriting the patched libraries first, which would have made it moot - the guard is one `case` and does not depend on the answer, so I took it rather than spending a build finding out.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
jetm: Fixed in `fa77b15`. Both loops now call `rust_snapshot_ensure_rpath`, which reads what the object carries and appends only what is missing, so neither can assert over an entry it did not read.

I did not fetch the snapshot to check for a second RUNPATH entry either, and it does not matter now: ensure-semantics cannot drop one whether or not today's snapshot has it. The harness pins it - a `rustc` seeded with `$ORIGIN/../lib:$ORIGIN/../lib/rustlib/x86_64-unknown-linux-gnu/lib` keeps the `rustlib` entry, where the old fixed assertion dropped it.

Worth naming the regression this could have caused, since it is the reason the helper takes the entries as arguments rather than just appending the staging dir: the bins carry no rpath at all before this task runs, so a plain read-then-append would have set only `<staging>` and left them unable to find their sibling libraries in `../lib`. There is a harness case asserting a bare bin ends up with `$ORIGIN/../lib:<staging>`, exactly what the old line produced.

meta-avocado/recipes-devtools/cmake/cmake-native_3.28.%.bbappend
jetm: Half of this holds and half of it does not. The comment is rewritten in `d8ff0fa`; the pin stays.

**The dangling-append argument does not apply at this commit.** It rests on `LAYERSERIES_COMPAT_meta-avocado = "scarthgap wrynose"`, and at `1357106a` that line reads `"scarthgap"` alone (`meta-avocado/conf/layer.conf:13`). The layer never parses against a wrynose oe-core, so `3.28.%` cannot strand a build that is already happening and there is no `bb.fatal` to trip. It becomes a live concern the moment compat widens, which is why the rewritten comment says so in as many words rather than leaving it implicit.

**"The pin misses 4.3.x" is right, and it does not argue for widening.** `CMAKE_EXTRACONF` does not exist in `cmake-native_4.3.1.bb` - it passes `-- ${EXTRA_OECMAKE}` to bootstrap (`:46`). So a `cmake-native_%.bbappend` would match after the bump and set a variable no recipe reads: precisely the append that "parses, sets nothing readable, warns about nothing" that the pin exists to avoid, just reached from the variable rather than from a claim about `find_package`. Widening the glob without also switching variable would be worse than the pin, not better.

And 4.3.x needs nothing from this layer regardless: `854e940580d` puts `-DCMAKE_DISABLE_FIND_PACKAGE_Libidn2=ON` in that recipe's own `EXTRA_OECMAKE`, and it is on the wrynose branch. The rewritten comment records that, plus the one thing to actually verify at bump time - that the pinned oe-core carries the commit. I could not check that here: the pin is a vendor fork revision that is not an object in upstream oe-core, so it is a bump-time check rather than something I could settle now.

You were right that the old text was load-bearing in the wrong direction; see the thread on line 27 for the part of this that was outright wrong.

meta-avocado/recipes-devtools/cmake/cmake-native_3.28.%.bbappend
jetm: Confirmed and rewritten in `d8ff0fa`. I read `854e940580d` and it says what you quote: cmcurl "calls `find_package(Libidn2)`. On hosts with libidn2-devel installed ... this detects the host library during configure", failing with `fatal error: idn2.h: No such file or directory`. So the old text had the mechanism backwards and named the already-merged fix as a hypothetical lever.

The comment now states that 4.3.x has the same leak by a different route, that oe-core fixes it with `-DCMAKE_DISABLE_FIND_PACKAGE_Libidn2=ON` in `cmake-native_4.3.1.bb`'s `EXTRA_OECMAKE`, and that the bump's real obligation is checking the pinned oe-core carries that commit. Nothing in it now claims `find_package` is host-safe.

On reconsidering the shape: I kept the seeding rather than switching to `DEPENDS += "libidn2-native"`. Both work - `libidn2` does class-extend to native, and `BUILD_LDFLAGS` carries `-Wl,-rpath,${STAGING_LIBDIR_NATIVE}`, so the qemu-system-native precedent transfers cleanly. What decided it is that upstream looked at the same choice and disabled IDN instead ("cmake-native does not need IDN support in its internal curl usage"), so staging the library would keep a capability oe-core deliberately dropped and leave us maintaining the divergence. Converging on upstream also means this append shrinks to a scarthgap-only concern and disappears at the bump, where `DEPENDS` would persist.

The `SANITY_TESTED_DISTROS` observation is a good one and I did not act on it - naming the real build distro is a change with a much wider blast radius than this PR, and it belongs on its own.

meta-avocado/recipes-devtools/cmake/cmake-native_3.28.%.bbappend
jetm: Confirmed and fixed in `d8ff0fa`. `set(USE_LIBIDN2 ON)` is at `Utilities/cmcurl/CMakeLists.txt:90`, and the citation now carries the `Utilities/cmcurl/` prefix so it no longer reads as cmake's top-level file.

I derived it the way you described, from the hunk header of oe-core's `0001-CMakeLists.txt-disable-USE_NGHTTP2.patch` on scarthgap. Worth writing down that the header's trailing text is the section heading rather than the hunk's first line: `@@ -88,7 +88,7 @@ set(HTTP_ONLY OFF ...)` means the hunk body starts at 88 with `set(PICKY_COMPILER OFF ...)`, giving 89 `SHARE_LIB_OBJECT`, 90 `USE_LIBIDN2`, 91 `USE_NGHTTP2` - and the seven-line count checks out. Reading `HTTP_ONLY` as line 88 would shift everything by one and put `USE_LIBIDN2` at 91.

The premise you flagged as needing checking was indeed the wrong one - see the thread on line 27.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
jetm: Both halves accepted. The `[depends]` line is gone in `40d2856` and the call sites now use `patchelf-uninative`.

Verified before making the change: `uninative.bbclass:151` prepends the uninative bindir to `PATH`, and `uninative-tarball.bb:63` moves that tarball's patchelf to `patchelf-uninative`. Both are unconditional on the uninative path, so the tool exists under exactly the condition the guard tests - which makes calling it by that name a removal of the dependency rather than a way of satisfying it.

I took the divergence-from-mixins tradeoff you named. Calling it by the uninative name matches oe-core scarthgap `rust_1.75.0`, and the mixins base task calling bare `patchelf` is not a constraint on an append that only runs inside the loader guard.

The task-name binding argument stands on its own and is the better half of this: a varflag on a task that is never in `__BBTASKS` is set, never read, and never warned about, so the failure would have been the original opaque `rustc -vV` 127 with an append sitting beside it that looked like it prevented exactly that. Removing the flag removes that mode rather than documenting it.

One thing the change adds rather than just deletes: a `command -v patchelf-uninative` check before either loop, which exists because the per-object skip inside `rust_snapshot_ensure_rpath` would otherwise be able to stand in for an absent tool. It is a guard against the uninative layout changing, not against a missing dependency.

meta-avocado/recipes-devtools/rust/rust_%.bbappend
jetm: Confirmed, and resolved the way you suggested: the hedge is gone because the line it was propping up is gone (`40d2856`).

I checked `kas/base.yml` at this PR's head and you are right - the `meta-lts-mixins` entry is a plain pinned repo with no annotation. There is an annotated, commented-out version of that entry, but it lives on `wrynose-frontport-fixes`, not here, so reading it out of a checked-out clone rather than at the head SHA is what would make this claim look supported. I nearly made that mistake checking it.

So the resolution is the one you pointed at: drop the `[depends]` for the right reason - `patchelf-uninative` is already on `PATH` inside the guard - rather than keep it for a wrong one. The keep-rationale annotations you spotted for meta-atmel, meta-erlang and the rest are presumably where the impression came from.

mobileoverlord (approved):
Host leakage in three recipes, fixed at the layer level where the vendor forks make it belong. CI green, all 13 self-review threads resolved with fix commits behind them, and the cold-build failure this unblocks is documented with cleansstate evidence rather than asserted.

Two notes, neither blocking: the ganesha hunk (1357106a) is the same class of bug but is missing from the body's "Key changes" list, and the 3.28.% pin is the right call precisely because it fails loudly as a dangling append at the wrynose bump instead of quietly setting a variable nothing reads.

PR #264, opened 2026-08-06
scripts: add avocado-flash, a backend-dispatching image writer

jetm:
## Problem

Writing a built image to a board is a hand-assembled command per medium today:
`fwup` against the stone archive for a card, `uuu` against `imx-boot` for serial
download, each with its own argument shape and its own way of locating the
artifact. The knowledge lives in whoever last did it.

The destructive half has no guard. `fwup` rewrites the partition table from
sector 0, and a development host keeps multi-terabyte disks on the same
`/dev/sd*` namespace as the card reader, so one mistyped letter destroys the
wrong disk with no prompt and no undo.

## Solution

Dispatch the backend from the medium rather than hardcoding one:

| Medium | Default backend | Alternative |
|---|---|---|
| `sd` (removable block device) | `fwup` | `bmaptool` |
| `emmc` (USB serial download) | `uuu` | - |

Adding a board that provisions over a different transport becomes a
`resolve_*` / `flash_*` pair, not a second script. The safety guards are
written once instead of per caller.

`fwup` stays the default for a card rather than a generic raw writer because
it is the only backend that reads the manifest, so partitions marked `expand`
grow to fill the medium. Measured on an i.MX93 FRDM with a 32GB card: the same
build yields a 29G `/var` written this way, and 112M written through the uuu
path, which lays down the fixed-size image.

`bmaptool` is offered for plain `.wic` images, where its per-mapped-region
SHA256 and its refusal to write a mounted device are worth more than manifest
awareness.

## Key changes

- New `scripts/avocado-flash`, backend selected from the medium argument.
- Block-device guards: removable-only, whole-disk-only, no mounted
  partitions, and a retype-the-device confirmation. `--assume-yes` exists for
  a board farm where the device comes from an exporter.
- Backends resolved from the build's native sysroot before `PATH`, because
  `fwup` is not packaged for most distributions and its own component
  directory does not resolve its shared libraries - they live in sibling
  component trees, while a recipe sysroot resolves them.
- Machine and deploy directory autodetected, overridable with `--machine` /
  `--deploy`.
- `--dry-run` prints the exact command and touches nothing.

## Reviewer notes

- Formatted to the repo's enforced `shfmt -i 2 -ci -bn`; `shellcheck` clean.
  Existing `scripts/dev-*.sh` use 4-space indent and predate that hook, which
  only runs on changed files.
- Uses `set -euo pipefail` rather than the bare `set -e` its siblings carry.
  This one writes raw block devices, where a variable expanding to empty
  selects the wrong disk silently. The deviation is deliberate.
- The archive mtime is printed because the fwup archive is produced by
  `stone provision`, not `stone bundle`. A bundle-only rebuild refreshes
  `os-bundle.aos` and leaves the archive untouched, so its age is the only
  visible signal that the bootloader about to be written is stale.
- Guards exercised against a real host: a mounted 9.1T data disk, a partition
  rather than a whole disk, the root NVMe, a missing device argument, an
  unknown medium, and a backend that cannot write a block device are each
  refused with a distinct message.
- The `sd`/`fwup` path is verified end to end on hardware: autodetect from the
  workspace root, confirmation, archive resolution, and the real
  `sudo env LD_LIBRARY_PATH=... fwup -a -i ... -d /dev/sde -t complete` write
  to a 32GB card in 26.9s, followed by a successful boot of the resulting card
  on an i.MX93 FRDM to a login prompt with `/` on `mmcblk1p6` and `/var`
  expanded to 29G on `mmcblk1p8`.
- Deploy-directory autodetect verified from three working directories: the
  workspace root and `meta-avocado/scripts` both resolve, an unrelated
  directory fails with a message naming both escape hatches.
- Not yet exercised: the `bmaptool` and `uuu` paths, which need a `.wic`
  image and a board in serial-download mode respectively.

## Follow-up worth considering

Bootlin's [snagboot](https://github.com/bootlin/snagboot) (GPL-2.0, Python) is
a vendor-agnostic replacement for `uuu`, STM32CubeProgrammer, SAM-BA and
`sunxi-fel`, and its supported list already includes i.MX91 and i.MX93. It
would slot in as one more backend. Whether `snagflash` can drive the full
i.MX93 eMMC flow including the boot-partition write and `mmc partconf`, or
only the stage after U-Boot is running, is unverified - worth an afternoon on
a board before committing to it.

### DIFF
--- a/scripts/avocado-flash
+++ b/scripts/avocado-flash
@@ -137,17 +137,38 @@ confirm_destructive() {
   [ "$reply" = "$dev" ] || die "mismatch, aborting"
 }
 
+# Look next to the script, then walk up from the working directory. The second
+# half matters because the script is not always run from the checkout it
+# belongs to - a worktree, a copy on a board farm, or a symlink into PATH all
+# leave SCRIPT_DIR pointing somewhere with no build beside it.
 resolve_deploy_dir() {
   [ -n "$DEPLOY_DIR" ] && return 0
 
-  local candidate
-  for candidate in "$SCRIPT_DIR"/../../build*/build/tmp/deploy "$SCRIPT_DIR"/../../build/tmp/deploy; do
-    if [ -d "$candidate" ]; then
-      DEPLOY_DIR=$(cd "$candidate" && pwd)
-      return 0
-    fi
+  # Inside a bitbake or kas shell this is already the answer.
+  if [ -n "${BUILDDIR:-}" ] && [ -d "$BUILDDIR/tmp/deploy" ]; then
+    DEPLOY_DIR=$(cd "$BUILDDIR/tmp/deploy" && pwd)
+    return 0
+  fi
+
+  local root candidate
+  for root in "$SCRIPT_DIR/../.." $(ancestors_of "$PWD"); do
+    for candidate in "$root"/build*/build/tmp/deploy "$root"/build/tmp/deploy "$root"/tmp/deploy; do
+      if [ -d "$candidate" ]; then
+        DEPLOY_DIR=$(cd "$candidate" && pwd)
+        return 0
+      fi
+    done
   done
-  die "could not find a deploy directory; pass --deploy"
+
+  die "could not find a deploy directory; pass --deploy or set BUILDDIR"
+}
+
+ancestors_of() {
+  local dir="$1"
+  while [ "$dir" != "/" ] && [ -n "$dir" ]; do
+    printf '%s\n' "$dir"
+    dir=$(dirname "$dir")
+  done
 }
 
 resolve_machine() {
--- a/scripts/avocado-flash
+++ b/scripts/avocado-flash
@@ -113,22 +113,39 @@ run_native_tool() {
 # hosts routinely carry multi-terabyte disks on the same /dev/sd* namespace as
 # the card reader.
 assert_safe_block_device() {
-  local dev="$1" base mounted
+  local dev="$1" resolved base mounts
 
   [ -b "$dev" ] || die "not a block device: $dev"
 
-  base=$(basename "$dev")
+  # Resolve before deriving the sysfs name. /dev/disk/by-id and /dev/disk/by-path
+  # entries are symlinks, so they satisfy -b while /sys/block/<link name> does
+  # not exist - which made the whole-disk check below reject them as partitions.
+  # Those are the names a board farm hands out, and the -y path they are used on
+  # is the one with no prompt left to catch a mistake, so refusing them forced
+  # the caller down to a re-enumerable /dev/sdX letter instead.
+  resolved=$(readlink -f "$dev")
+  base=$(basename "$resolved")
   [ -d "/sys/block/$base" ] \
     || die "refusing a partition: pass the whole disk (e.g. /dev/sdb, not $dev)"
 
   [ "$(cat "/sys/block/$base/removable" 2>/dev/null)" = "1" ] \
     || die "refusing $dev: the kernel does not report it removable"
 
-  mounted=$(lsblk -nro MOUNTPOINTS "$dev" | grep -c . || true)
-  [ "$mounted" -eq 0 ] || {
-    lsblk -o NAME,SIZE,TYPE,RM,MOUNTPOINTS "$dev" >&2
+  # Read lsblk's own exit status rather than the pipeline's. `grep -c .` exits 1
+  # on zero lines, which is the legitimate nothing-is-mounted answer, so the old
+  # `|| true` was needed - and it swallowed a failing lsblk too, leaving
+  # mounted=0 and passing the guard. MOUNTPOINT is singular on purpose: the
+  # plural column arrived in util-linux 2.37, and scarthgap's own
+  # SANITY_TESTED_DISTROS still lists hosts on 2.32-2.36, where lsblk exits 1
+  # with `unknown column: MOUNTPOINTS`. On those the guard passed and fwup
+  # rewrote the partition table from sector 0 under a live mount.
+  if ! mounts=$(lsblk -nro MOUNTPOINT "$dev" 2>/dev/null); then
+    die "refusing $dev: cannot read its mount state (lsblk failed)"
+  fi
+  if printf '%s\n' "$mounts" | grep -q .; then
+    lsblk -o NAME,SIZE,TYPE,RM,MOUNTPOINT "$dev" >&2
     die "refusing $dev: it has mounted partitions"
-  }
+  fi
 }
 
 confirm_destructive() {
@@ -298,6 +315,24 @@ execute() {
   "${RUN_CMD[@]}"
 }
 
+# Close out a write, or say plainly that none happened. Neither `sync` nor the
+# completion message used to be guarded, so `-n -y sd /dev/sdb` printed `done.
+# set the board's boot switches...` and exited 0 having written nothing: an
+# operator sanity-checking a job definition powers the board, sees the previous
+# image, and concludes the flash failed rather than that it never ran. `sync` is
+# also a real syscall against every mounted filesystem, so running it broke the
+# --dry-run promise to touch nothing.
+finish() {
+  local next_step="$1" target="$2"
+
+  if [ "$DRY_RUN" = "1" ]; then
+    echo "dry run: nothing was written to $target."
+    return 0
+  fi
+  sync
+  echo "done. $next_step"
+}
+
 MACHINE_NAME=""
 DEPLOY_DIR=""
 BACKEND=""
@@ -349,6 +384,15 @@ done
   exit 1
 }
 
+# `*) break` above stops the option loop at the first non-option, and nothing
+# read past $2, so every option written after the medium was discarded without
+# a word - including the ones that decide what gets written. `sd /dev/sdb
+# --machine avocado-imx93-frdm` fell through to autodetect and wrote whatever it
+# guessed; a trailing `--backend bmaptool` wrote with fwup instead; a trailing
+# `-y` left the board-farm path prompting at EOF. Refusing the surplus closes
+# all of them at once, and says where the option belongs.
+[ $# -le 2 ] || die "unexpected argument: $3 (options go before the medium; see --help)"
+
 MEDIUM="$1"
 DEVICE="${2:-}"
 
@@ -377,15 +421,14 @@ case "$MEDIUM" in
       bmaptool) flash_bmaptool "$DEVICE" ;;
       *) die "backend ${BACKEND} cannot write a block device" ;;
     esac
-    sync
-    echo "done. set the board's boot switches to the SD position and power on."
+    finish "set the board's boot switches to the SD position and power on." "$DEVICE"
     ;;
   emmc)
     case "${BACKEND:-uuu}" in
       uuu) flash_uuu ;;
       *) die "backend ${BACKEND} cannot write over serial download" ;;
     esac
-    echo "done. set the board's boot switches to the eMMC position and power on."
+    finish "set the board's boot switches to the eMMC position and power on." "the board"
     ;;
   *)
     die "unknown medium: $MEDIUM (see --help)"
--- a/scripts/avocado-flash
+++ b/scripts/avocado-flash
@@ -36,6 +36,15 @@ Options:
                         when exactly one image directory exists)
     -d, --deploy DIR    Deploy directory (default: autodetect from the build)
     -b, --backend NAME  Override the backend: fwup, bmaptool, uuu
+    -t, --task NAME     fwup task (default: complete). 'complete' writes the
+                        whole medium including var, so it provisions a blank
+                        card and discards any state already on one. 'upgrade'
+                        writes only the inactive A/B boot+rootfs slots and
+                        leaves var untouched, which is how you put a new OS
+                        onto a device whose extensions and data must survive.
+                        fwup selects upgrade.a or upgrade.b from the current
+                        slot recorded in uboot-env, so the running slot is
+                        never the one overwritten. fwup backend only.
     -n, --dry-run       Print what would run; touch nothing
     -y, --assume-yes    Skip the interactive confirmation. Intended for a
                         board farm where the device comes from an exporter,
@@ -227,7 +236,8 @@ flash_fwup() {
   archive=$(resolve_fwup_archive)
 
   echo "  archive: $archive ($(date -r "$archive" '+%Y-%m-%d %H:%M'))"
-  run_native_tool fwup -a -i "$archive" -d "$dev" -t complete
+  echo "  task:    $FWUP_TASK"
+  run_native_tool fwup -a -i "$archive" -d "$dev" -t "$FWUP_TASK"
   execute
 }
 
@@ -293,6 +303,11 @@ DEPLOY_DIR=""
 BACKEND=""
 DRY_RUN=0
 ASSUME_YES=0
+# fwup's task, passed through as -t. `complete` writes the whole medium
+# including var; `upgrade` writes only the inactive A/B boot+rootfs slots and
+# leaves var alone, which is the difference between provisioning a blank card
+# and updating a device that already carries state.
+FWUP_TASK="complete"
 
 while [ $# -gt 0 ]; do
   case "$1" in
@@ -306,6 +321,10 @@ while [ $# -gt 0 ]; do
       ;;
     -b | --backend)
       BACKEND="$2"
+      shift 2
+      ;;
+    -t | --task)
+      FWUP_TASK="$2"
       shift 2
       ;;
     -n | --dry-run)
@@ -338,6 +357,15 @@ resolve_machine
 
 echo "machine: $MACHINE_NAME"
 echo "deploy:  $DEPLOY_DIR"
+
+# Only fwup evaluates the manifest's tasks. Refusing here rather than ignoring
+# the flag matters because the two tasks differ in whether /var survives: a
+# --task upgrade silently downgraded to a whole-medium write destroys the state
+# the caller asked to keep, and the write looks successful either way.
+if [ "$FWUP_TASK" != "complete" ] && [ "${BACKEND:-}" != "fwup" ] \
+  && { [ "$MEDIUM" != "sd" ] || [ -n "$BACKEND" ]; }; then
+  die "--task applies to the fwup backend only (got backend '${BACKEND:-default for $MEDIUM}')"
+fi
 
 case "$MEDIUM" in
   sd)
--- a/scripts/avocado-flash
+++ b/scripts/avocado-flash
@@ -177,19 +177,28 @@ assert_safe_block_device() {
   fi
 }
 
+# Ask before a destructive write, showing what is about to be written where.
+# `payload` is the artifact already resolved by the caller: the prompt used to
+# come before any of that, so the type-the-device gate - the only thing between a
+# typo and a wiped disk - was cleared as a reflex before the script had
+# established it had anything to write at all.
+#
+# `target` is a block device on the sd path and a USB path on the emmc one, so
+# the lsblk detail is shown only when it applies rather than assumed.
 confirm_destructive() {
-  local dev="$1" reply
+  local target="$1" payload="${2:-}" reply
 
   # A dry run writes nothing, so requiring confirmation would only make it
   # hang when called non-interactively.
   [ "$DRY_RUN" = "1" ] && return 0
   [ "$ASSUME_YES" = "1" ] && return 0
 
-  echo "=== writing the whole OS to $dev (destructive) ==="
-  lsblk -o NAME,SIZE,TYPE,RM,MODEL "$dev"
+  echo "=== writing the whole OS to $target (destructive) ==="
+  [ -n "$payload" ] && echo "  payload: $payload"
+  [ -b "$target" ] && lsblk -o NAME,SIZE,TYPE,RM,MODEL "$target"
   echo
-  read -r -p "type the device again to confirm: " reply
-  [ "$reply" = "$dev" ] || die "mismatch, aborting"
+  read -r -p "type the target again to confirm: " reply
+  [ "$reply" = "$target" ] || die "mismatch, aborting"
 }
 
 # Look next to the script, then walk up from the working directory. The second
@@ -205,15 +214,45 @@ resolve_deploy_dir() {
     return 0
   fi
 
-  local root candidate
-  for root in "$SCRIPT_DIR/../.." $(ancestors_of "$PWD"); do
+  # Collect every candidate at the nearest level that has any, then refuse if
+  # there is more than one. Returning the first match made the choice
+  # alphabetically: a workspace holding both build-imx93-frdm and
+  # build-qemuarm64 resolved the i.MX93 tree for a caller who meant qemuarm64
+  # and wrote that machine's fwup archive. resolve_machine's ambiguity guard
+  # never fired, because the directory it landed on held exactly one machine -
+  # so the guard existed and was unreachable in the very layout it was written
+  # for.
+  #
+  # The roots are fed in through a loop rather than word-split from
+  # $(ancestors_of "$PWD"). Unquoted, a path containing a space skipped the real
+  # build tree and a glob metacharacter - proj[1] - resolved to a sibling.
+  # The linter is silent on it too: SC2046 is not raised for a substitution in a
+  # for-list.
+  local root candidate found=()
+  while IFS= read -r root; do
+    found=()
     for candidate in "$root"/build*/build/tmp/deploy "$root"/build/tmp/deploy "$root"/tmp/deploy; do
-      if [ -d "$candidate" ]; then
-        DEPLOY_DIR=$(cd "$candidate" && pwd)
-        return 0
-      fi
+      [ -d "$candidate" ] && found+=("$(cd "$candidate" && pwd -P)")
     done
-  done
+    # `pwd -P` and then de-duplicate, because a `build` symlink pointing at the
+    # active build-<machine> directory is a common layout and both names match the
+    # globs above. Resolved logically they look like two trees and would trip the
+    # ambiguity refusal below over what is really one.
+    if [ "${#found[@]}" -gt 1 ]; then
+      mapfile -t found < <(printf '%s\n' "${found[@]}" | sort -u)
+    fi
+    case "${#found[@]}" in
+      0) continue ;;
+      1)
+        DEPLOY_DIR="${found[0]}"
+        return 0
+        ;;
+      *) die "several build trees under $root (${found[*]}); pass --deploy to choose one" ;;
+    esac
+  done < <(
+    printf '%s\n' "$SCRIPT_DIR/../.."
+    ancestors_of "$PWD"
+  )
 
   die "could not find a deploy directory; pass --deploy or set BUILDDIR"
 }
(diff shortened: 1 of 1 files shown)


scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+resolve_fwup_archive() {
+  local archive="$DEPLOY_DIR/stone/_build/${MACHINE_NAME#avocado-}-rootdisk.zip"
+  [ -f "$archive" ] || archive="$DEPLOY_DIR/stone/_build/$MACHINE_NAME-rootdisk.zip"
+  [ -f "$archive" ] || die "no fwup archive under $DEPLOY_DIR/stone/_build (run stone_provision)"
+  printf '%s' "$archive"
+}
+
+resolve_raw_image() {
+  local image
+  for image in "$DEPLOY_DIR/images/$MACHINE_NAME"/*.wic "$DEPLOY_DIR/stone/_build"/*-rootdisk.img; do
```
jetm: **This glob can return another machine's image, and the error below claims it cannot**

The second glob searches the flat, shared `stone/_build` directory with no `$MACHINE_NAME` filter, while line 213's error message claims per-machine scoping.

In a tree where only qemuarm64 was provisioned, `avocado-flash --machine avocado-imx93-frdm --backend bmaptool sd /dev/sdb` misses `images/avocado-imx93-frdm/*.wic`, then matches `stone/_build/avocado-qemuarm64-rootdisk.img`, and writes a qemuarm64 image to the i.MX93 card. The only thing that reveals it is the `image:` echo, which a `-y` caller never reads.

`resolve_machine` (174-180) also never checks that the named machine exists in this tree, so `--machine` is not a gate on anything - it is a hint that gets silently discarded when it does not match.

Scope the second glob to `$MACHINE_NAME` and fail rather than fall through.

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+  [ -n "$DEPLOY_DIR" ] && return 0
+
+  # Inside a bitbake or kas shell this is already the answer.
+  if [ -n "${BUILDDIR:-}" ] && [ -d "$BUILDDIR/tmp/deploy" ]; then
+    DEPLOY_DIR=$(cd "$BUILDDIR/tmp/deploy" && pwd)
+    return 0
+  fi
+
+  local root candidate
+  for root in "$SCRIPT_DIR/../.." $(ancestors_of "$PWD"); do
```
jetm: **Autodetect silently picks the first of several build trees - reproduced in this workspace**

The ancestor walk accepts the first `"$root"/build*/build/tmp/deploy` match, and `resolve_machine`'s ambiguity guard at 188-192 never fires, because the dir it lands on holds exactly one machine.

This workspace has both `build-imx93-frdm/build/tmp/deploy` and `build-qemuarm64/build/tmp/deploy`. Running `avocado-flash sd /dev/sdb` from anywhere under `~/repos/work/peridio` intending qemuarm64 resolves `build-imx93-frdm` - alphabetically first - and writes the i.MX93 fwup archive. So the ambiguity guard exists but is unreachable in exactly the layout it was written for.

The unquoted `$(ancestors_of "$PWD")` compounds it: a path containing a space skips the real build tree, and a glob metacharacter (`proj[1]`) resolves to a sibling. shellcheck does not catch this - SC2046 is not raised inside `for`.

Collect all matches and fail on more than one, the way `resolve_machine` already intends to.

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+flash_uuu() {
+  local boot="$DEPLOY_DIR/images/$MACHINE_NAME/imx-boot"
+  [ -f "$boot" ] || die "no imx-boot for $MACHINE_NAME"
+
+  if ! lsusb -d 1fc9: >/dev/null 2>&1 && ! lsusb -d 15a2: >/dev/null 2>&1; then
+    die "no NXP device in serial download mode (check the boot switches)"
+  fi
+
+  echo "  bootloader: $boot"
+  RUN_CMD=(sudo uuu -b emmc "$boot")
```
jetm: **`uuu -b emmc` writes only the bootloader, then the script prints success**

Verified locally with uuu 1.5.243: `uuu -bshow emmc` ends at `FB: flash bootloader _image`. The script that carries `FB: flash -raw2sparse all _image` is `emmc_all`.

So `avocado-flash emmc` on a factory-blank i.MX93 programs `imx-boot` only, exits 0, and prints `done. set the board's boot switches to the eMMC position and power on.` (335) - while line 3 promises to "Write a built Avocado OS image to a board". The board stops at the U-Boot prompt and reads as bricked to whoever picks it up.

The layer already does this correctly: `meta-avocado-nxp/stone/avocado-imx/stone-provision-uuu-emmc.sh` expands the archive, patches `devnum`/`mmcblk`, and emits `flash -raw2sparse all rootdisk.img`. Worth reusing that rather than re-deriving the uuu invocation here - which is also the answer to the duplication point I raise at the bottom of the review.

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+  [ -b "$dev" ] || die "not a block device: $dev"
+
+  base=$(basename "$dev")
+  [ -d "/sys/block/$base" ] \
+    || die "refusing a partition: pass the whole disk (e.g. /dev/sdb, not $dev)"
+
+  [ "$(cat "/sys/block/$base/removable" 2>/dev/null)" = "1" ] \
+    || die "refusing $dev: the kernel does not report it removable"
+
+  mounted=$(lsblk -nro MOUNTPOINTS "$dev" | grep -c . || true)
```
jetm: **`|| true` turns the mounted-partition guard into an unconditional pass**

`grep -c .` legitimately exits 1 on zero lines, so something is needed here - but this swallows `lsblk`'s status too, and a failed `lsblk` prints nothing, so `mounted=0` and the guard passes.

The version trigger is concrete: `MOUNTPOINTS` (plural) landed in util-linux 2.37. Scarthgap's own `SANITY_TESTED_DISTROS` includes Ubuntu 20.04 (2.34), Debian 11 (2.36) and CentOS/Rocky 8 (2.32). On any of them `lsblk` exits 1 with `unknown column: MOUNTPOINTS`, the guard passes, the user confirms the device they did mean, and fwup rewrites the partition table from sector 0 under a live mount.

It is version-independent too: `lsblk` on a device removed between 109 and 118 returns rc=32 and also passes.

Note lines 14-17 advertise `pipefail` as the defence against exactly this class of thing, and this is the one line that opts out of it. Capture the status and the output separately, and treat a non-zero `lsblk` as fatal rather than as "nothing mounted".

(The happy path does work - `MOUNTPOINTS` on a whole disk does report child partitions. I checked.)

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+    confirm_destructive "$DEVICE"
+    case "${BACKEND:-fwup}" in
+      fwup) flash_fwup "$DEVICE" ;;
+      bmaptool) flash_bmaptool "$DEVICE" ;;
+      *) die "backend ${BACKEND} cannot write a block device" ;;
+    esac
+    sync
+    echo "done. set the board's boot switches to the SD position and power on."
+    ;;
+  emmc)
```
jetm, line 577: **The emmc branch reaches a destructive write with no confirmation and no target pinning**

`confirm_destructive` is called only at 321, inside the `sd` arm. `-y` is irrelevant on this path because nothing prompts.

Line 248 matches a vendor ID only (`lsusb -d 1fc9:`), and uuu without `-m <usbpath>` takes whichever known device enumerates first. With two i.MX boards on the bench - one in SDP mode for an unrelated bisect - `avocado-flash emmc` overwrites that board's eMMC boot area with no prompt at all.

The header defence at 241-243 says "the transport itself selects the target", and that does not hold: a vendor-ID match is not a target selection. Either pin `-m` to a specific USB path and show it in a confirmation, or refuse when more than one `1fc9:` device is present.

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+    -y | --assume-yes)
+      ASSUME_YES=1
+      shift
+      ;;
+    -h | --help)
+      usage
+      exit 0
+      ;;
+    -*) die "unknown option: $1 (see --help)" ;;
+    *) break ;;
```
jetm, line 515: **Surplus arguments are silently discarded, including the ones that change what gets written**

`*) break ;;` exits the option loop at the first non-option, and nothing after `DEVICE="${2:-}"` inspects `$3` or rejects extra words.

So `avocado-flash sd /dev/sdb --machine avocado-imx93-frdm` drops `--machine` entirely - `resolve_machine` falls through to `$MACHINE` or single-dir autodetect and writes whatever it guessed, with no gate on image identity anywhere. `--backend bmaptool` trailing writes with fwup instead. `sd /dev/sdb --dry-run` performs a real write (the confirm prompt does stop that one, so it is the least bad of the set).

Trailing `-y` silently breaks the board-farm path documented at 40-42: the write still prompts, so a non-interactive caller aborts at EOF.

A `[ $# -gt 2 ] && die` after the loop closes all of these at once.

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+    0) die "no images found under $DEPLOY_DIR/images; pass --machine" ;;
+    *) die "several machines built (${dirs[*]}); pass --machine" ;;
+  esac
+}
+
+# The fwup archive is produced by stone provision, not by stone bundle. A
+# bundle-only rebuild refreshes os-bundle.aos and leaves this stale, so its
+# age is worth reporting rather than silently writing yesterday's bootloader.
+resolve_fwup_archive() {
+  local archive="$DEPLOY_DIR/stone/_build/${MACHINE_NAME#avocado-}-rootdisk.zip"
```
jetm: **The fwup archive name is guessed, and the guess fails on most machines**

I enumerated all 30 stone manifests. Every one names it `avocado-<short>-rootdisk.*`, i.e. `$MACHINE_NAME-rootdisk.zip` - so line 199's `${MACHINE_NAME#avocado-}` probe never matches anything, and only line 200's fallback does any work. The two are in the wrong order.

Worse, neither probe matches 17 of the 30: qemu is `avocado-os-qemux86-64.zip`, renesas and qcom are `.tar.gz`, nvidia and x86-64 are `avocado-image-jetson-*` with no extension, and alif/rockchip/stm are `.img`. On those machines `flash_fwup` - the **default** `sd` backend - dies with `no fwup archive under .../stone/_build (run stone_provision)` after `do_stone_provision` already succeeded, sending the user to re-run a build that ran.

The manifest is deployed alongside the image and already carries the answer: `meta-avocado-nxp/stone/avocado-imx/stone-provision-sd.sh:19` reads it with `jq -r .storage_devices.rootdisk.out`. Read the name instead of guessing it.

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+
+  echo "  archive: $archive ($(date -r "$archive" '+%Y-%m-%d %H:%M'))"
+  run_native_tool fwup -a -i "$archive" -d "$dev" -t complete
+  execute
+}
+
+flash_bmaptool() {
+  local dev="$1" image bmap
+  image=$(resolve_raw_image)
+  bmap="${image%.*}.bmap"
```
jetm: **This bmap filename is one Yocto never produces, so bmaptool always full-copies**

oe-core `meta/classes-recipe/image_types.bbclass:357` emits `${IMAGE_NAME}.${type}.bmap`, i.e. `<image>.wic.bmap`. Stripping `.wic` first yields a name that never exists. The layer already knows the real name - `meta-avocado-rockchip/.../avocado-rockchip.inc:22` does `IMAGE_FSTYPES:remove = "wic.bmap"`.

So `--backend bmaptool sd /dev/sdb` always takes the `--nobmap` branch, and the cost is worse than losing sparse skipping in two ways. bmaptool's own `find_and_open_bmap()` tries `<image>.wic.bmap` first and would have found the file unaided, so passing `--nobmap` actively disables discovery that works. And with no bmap, `BmapCopy` yields `chksum = None`, which forfeits the per-region SHA256 verification this script advertises at 50-53 - on every run.

The message it prints reads like a broken build rather than a bug on this line, which is why it has gone unnoticed.

jetm (commented):
Independent second read on my own PR, cold. Twelve findings inline. This is a tool that writes to block devices, so I went looking hardest at the paths where it can write the wrong thing to the right device, or the right thing to the wrong device - and both exist.

Three I would not merge without:

1. **`uuu -b emmc` writes only the bootloader and then the script prints success.** Verified against uuu 1.5.243: `emmc` ends at `flash bootloader`; `emmc_all` is the one that writes the rootfs. A factory-blank i.MX93 gets `imx-boot` only, exit 0, and "done. set the board's boot switches to the eMMC position and power on." The board stops at the U-Boot prompt and reads as bricked.
2. **The `|| true` on the mounted-partition check makes it an unconditional pass.** On any util-linux older than 2.37 - which includes three distros in scarthgap's own `SANITY_TESTED_DISTROS` - `lsblk` exits 1 on the `MOUNTPOINTS` column, the guard passes, and fwup rewrites the partition table from sector 0 under a live mount.
3. **The emmc path reaches a destructive write with no confirmation at all**, and selects its target by vendor ID, so with two i.MX boards on the bench it can flash the wrong one silently.

Then two that make it write the wrong image, both of which I reproduced rather than reasoned about: the bmaptool glob is not machine-scoped, so it can pick another machine's `.img` out of the shared `stone/_build`; and deploy-dir autodetect takes the alphabetically first `build*/` tree, which **I reproduced in this workspace** - it has both `build-imx93-frdm` and `build-qemuarm64`, and a bare `avocado-flash sd /dev/sdb` intending qemuarm64 resolves the i.MX93 archive. The ambiguity guard at 188-192 exists but is unreachable in exactly that layout.

The rest are a mix of "the default path does not work" (the fwup archive name never matches on 17 of 30 machines; the bmap filename is one Yocto never emits, which also silently forfeits the SHA256 verification the header advertises), argument handling that drops flags that change what gets written, a dry run that calls `sync` and claims success, and two device-name predicates that are simultaneously too strict and too loose - `removable=1` refuses `/dev/mmcblk0`, which is the documented primary use case, while by-id/by-path names die telling you to pass the whole disk you just passed.

Two things worth saying that are not inline comments.

**Nothing references this script.** No recipe, README, `docs/`, kas yaml, workflow, or sibling script mentions `avocado-flash`, while every sibling script is documented. For a destructive tool that is the gap I would close first - `docs/` has no build-host or flashing section for it to live in.

**It duplicates an extension point the layer already has.** `STONE_PROVISIONING`, the stone manifest profiles, and `AVOCADO_PROVISION_PROFILE` on `do_stone_provision` are the existing per-machine provisioning registry, with a contract written down at `docs/adding-a-machine-target.md:337-357`. This adds a second parallel registry that no build exercises - only the manifest side is - which is why findings 4 and 7 both come down to re-deriving something the manifest already states. Reading the manifest instead of guessing filenames would collapse several of these at once.

Checked and clean, so it does not need re-litigating: the recipe side is a non-issue rather than a gap (`scripts/` at the layer root is developer-side and never packaged - the deployed provisioning scripts live in each layer's `stone/` dir), `set -euo pipefail` does what it should everywhere except line 118 (every command-substitution assignment that can fail is deliberately split from its `local`, so `die` inside `$( )` does propagate), both backend `case` statements have explicit `*) die` arms so there is no silent fallthrough, a hard failure in the `sd` path does propagate before the success message, and a `trap` genuinely is not needed - the only mutated state is the block device, which a trap cannot roll back. shellcheck 0.11.0, `shfmt -i 2 -ci -bn` and `bash -n` all pass, so none of the above is CI-catchable as it stands.

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+  sd)
+    [ -n "$DEVICE" ] || die "'sd' needs a device (see --help)"
+    assert_safe_block_device "$DEVICE"
+    confirm_destructive "$DEVICE"
+    case "${BACKEND:-fwup}" in
+      fwup) flash_fwup "$DEVICE" ;;
+      bmaptool) flash_bmaptool "$DEVICE" ;;
+      *) die "backend ${BACKEND} cannot write a block device" ;;
+    esac
+    sync
```
jetm: **A dry run calls `sync` and prints the success message**

Neither `sync` nor the "done." lines are `DRY_RUN`-guarded; `execute` returns 0 and control falls straight through.

A farm operator runs `avocado-flash -n -y sd /dev/sdb` to sanity-check a job definition, sees `done. set the board's boot switches to the SD position and power on.` and exit 0, powers the board, and it boots the previous image - concluding the flash silently failed rather than that it never ran.

Line 39 also promises `--dry-run ... touch nothing`, and `sync` is a real syscall against every mounted filesystem, so that promise is not quite kept either.

Guard both the `sync` and the completion message on `DRY_RUN`, and print something that says plainly nothing was written.

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+resolve_machine
+
+echo "machine: $MACHINE_NAME"
+echo "deploy:  $DEPLOY_DIR"
+
+case "$MEDIUM" in
+  sd)
+    [ -n "$DEVICE" ] || die "'sd' needs a device (see --help)"
+    assert_safe_block_device "$DEVICE"
+    confirm_destructive "$DEVICE"
```
jetm: **The destructive confirmation fires before the payload is resolved**

Ordering is: guard (320), human retypes the device (136), and only then `resolve_fwup_archive` (218) / `resolve_raw_image` (227) / the `imx-boot` check (246) / the sysroot lookup.

After a `stone bundle`-only rebuild, `avocado-flash sd /dev/sdb` shows the destructive banner, the user types `/dev/sdb`, and *then* gets `no fwup archive`.

The cost is not the wasted keystrokes - it is that the type-the-device-name gate is the only thing standing between a typo and a wiped disk, and this trains people to clear it as a reflex before the script has established it has anything to write at all. Resolve the artifact first, show it in the confirmation, then ask.

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+assert_safe_block_device() {
+  local dev="$1" base mounted
+
+  [ -b "$dev" ] || die "not a block device: $dev"
+
+  base=$(basename "$dev")
+  [ -d "/sys/block/$base" ] \
+    || die "refusing a partition: pass the whole disk (e.g. /dev/sdb, not $dev)"
+
+  [ "$(cat "/sys/block/$base/removable" 2>/dev/null)" = "1" ] \
```
jetm: **The `removable` check refuses the primary documented use case**

`drivers/mmc/core/block.c` deliberately never sets `GENHD_FL_REMOVABLE` ("MMC block devices clearly fall under the second case"), so an SD card in any native SDHCI/MMC reader appears as `/dev/mmcblk0` with `removable=0`, and line 116 refuses it - for exactly the case line 29 documents as "sd - removable block device (SD card, USB stick)".

That is the failure that matters, because the fallback is a hand-rolled `dd` with none of the guards in this file.

It over-accepts in the other direction too: `scsi_scan.c:928` copies the SCSI RMB bit verbatim with no size or bus check, so a multi-TB USB enclosure asserting RMB passes. (That half is unverified - no USB storage attached here to demonstrate it.)

`ID_BUS`/`ID_PATH` from udev, or a size ceiling, is the durable predicate. The current one is both too strict and too loose at once.

scripts/avocado-flash
```diff
@@ -0,0 +1,340 @@
...
+# bmaptool both rewrite the partition table from sector 0, so a mistyped
+# letter destroys whatever it names with no prompt and no undo. Development
+# hosts routinely carry multi-terabyte disks on the same /dev/sd* namespace as
+# the card reader.
+assert_safe_block_device() {
+  local dev="$1" base mounted
+
+  [ -b "$dev" ] || die "not a block device: $dev"
+
+  base=$(basename "$dev")
```
jetm: **Stable device names die with a message telling you to do what you just did**

Verified on this host: `/dev/disk/by-id/ata-TOSHIBA_...` and `/dev/disk/by-path/pci-...` both pass `[ -b ]` but have no `/sys/block/<basename>`, so line 113 dies with `refusing a partition: pass the whole disk`.

This bites hardest on the `-y` board-farm path documented at 40-42, because farm exporters hand out by-id/by-path names. It forces the caller to resolve to a re-enumerable `/dev/sdX` letter - the least stable identifier available - on the one path where there is no prompt left to catch a mistake.

`readlink -f` before `basename` fixes it, and would let the farm path use the stable name it was given.

scripts/avocado-flash
jetm: Fixed in `a13b934`, and the failure reproduces exactly as described.

I extracted the old guard and ran it against a stub `lsblk` that exits 1 with `unknown column: MOUNTPOINTS`: it printed `OLD GUARD PASSED (mounted=0)` and returned 0. So the path to a partition-table rewrite under a live mount was open, not theoretical.

The fix separates the two things that one line was conflating. `lsblk`'s own exit status is now read on its own, and a failure is fatal (`refusing <dev>: cannot read its mount state`), so nothing about the mount state is inferred from an absent answer. The `grep -q .` that follows lives in an `if`, so its legitimate exit 1 on zero lines no longer needs suppressing.

I also switched to the singular `MOUNTPOINT`, which every util-linux carries, rather than keeping `MOUNTPOINTS` and relying on the new fatal path to catch old hosts. Failing loudly on Ubuntu 20.04 would still have been a regression for anyone building there; asking for a column that exists means those hosts work. Verified all three cases against stubs - lsblk failing dies, nothing mounted passes, a mounted `/boot` refuses - and that `lsblk -nro MOUNTPOINT` exits 0 on this host (util-linux 2.42.2).

The rc=32 variant you mention is covered by the same change, since it is also a non-zero exit rather than empty output.

scripts/avocado-flash
jetm: Fixed in `a13b934` with the `readlink -f` you suggested, and the failure reproduces on this host:

```
link name:               ata-Samsung_SSD_870_EVO_2TB_...
/sys/block/<link name>   absent   -> died "refusing a partition"
readlink -f              resolves to a real disk whose /sys/block entry exists -> passes
```

The whole-disk check still does its job: a real partition resolves to a partition name, which has no `/sys/block` entry either, so it is refused exactly as before. Only the symlink case changes.

Worth noting the write itself was never the problem - fwup, bmaptool and raw copies all follow the symlink - so this was purely the validator rejecting a name the rest of the script would have handled. Which is why it bit the `-y` farm path specifically, as you said.

scripts/avocado-flash
jetm: Fixed in `a13b934` with the length check you proposed.

Confirmed by running it: passing `--machine avocado-imx93-frdm` after the medium and device now dies with `unexpected argument: --machine (options go before the medium; see --help)` before `resolve_deploy_dir`, so nothing is read or written first.

I phrased the message in terms of where the option belongs rather than only refusing it, since all four cases you listed are someone writing a real option in a reasonable-looking place. The `--dry-run` case you called the least bad is closed by the same line, so it no longer depends on the confirm prompt happening to catch it.

scripts/avocado-flash
jetm: Fixed in `a13b934`. Both arms now go through one `finish` helper that either syncs and prints the next step, or says plainly `dry run: nothing was written to <target>.`

Verified both ways: with `DRY_RUN=1` only the dry-run line appears, with `DRY_RUN=0` it syncs and prints `done. set the board's boot switches...`.

You were right that the `sync` broke the `--dry-run` promise on its own terms, independent of the misleading message. It is guarded now, so a dry run makes no syscall against any mounted filesystem.

scripts/avocado-flash
jetm: Replaced with the udev predicate in `8dcb84f`, and one correction to the suggestion: `ID_PATH`, not `ID_BUS`.

`ID_BUS` reads like the obvious property but udev only sets it for ata, scsi and usb. It is empty on nvme on this host, which means it is not something every disk carries - so keying on it would have refused an mmc device for want of a value, the same failure as the removable flag by a different route. `ID_PATH` comes from the `path_id` builtin, which names the controller on every bus: `-usb-` for a reader or stick, `mmc` for a native slot, `-ata-` and `-nvme-` for the host's own disks.

What I verified, and what I could not:

```
/dev/sdc      ID_PATH=pci-0000:14:00.0-ata-4.0    REFUSED   (this host's SATA SSD)
/dev/nvme0n1  ID_PATH=pci-0000:04:00.0-nvme-1     REFUSED
udevadm absent (real, PATH without it)             REFUSED   - not a fallback
ID_PATH empty                                      REFUSED
injected  ...-usb-0:3:1.0-scsi-0:0:0:0             ACCEPTED
injected  platform-fe340000.mmc                    ACCEPTED  <- the case the old flag rejected
```

The two accepts are against injected `ID_PATH` values, not attached hardware - there is no SD card or USB stick on this machine. So the pattern match is proven and the real-world property values are taken from udev's documented `path_id` output rather than observed. The two refusals that matter most are against real devices.

An absent `udevadm` refuses rather than falling back. It is in neither `HOSTTOOLS` nor `HOSTTOOLS_NONFATAL`, so its absence is reachable, and a safety guard that quietly degrades to a weaker predicate is exactly the bug being fixed here.

The over-accepting half you flagged as unverified is still open, and I put that in the function header rather than leaving it implicit: a USB enclosure holding a multi-terabyte disk is on a USB bus and passes. The retype-the-device prompt stands behind it, and `-y` removes that, so a size ceiling is the complement. Say the word if you want it in this PR rather than noted.

scripts/avocado-flash
jetm: Fixed in `b97bc79`, and the old behaviour reproduces exactly as described.

I ran the old resolver against a synthetic tree with only qemuarm64 provisioned and `MACHINE_NAME=avocado-imx93-frdm`: it returned `<deploy>/stone/_build/avocado-qemuarm64-rootdisk.img`. So the qemuarm64 image really was the next thing to reach the i.MX93 card, with the `image:` echo as the only signal.

Both globs are scoped to `$MACHINE_NAME` now, so the fall-through has nowhere to land: the second one is `stone/_build/$MACHINE_NAME-rootdisk.img`. Verified both directions - another machine's image is no longer offered, and the machine's own still resolves.

On `resolve_machine` not checking that the named machine exists: it still does not, and after this change it does not need to for the raw path, because a `--machine` that does not match resolves no image and dies rather than falling through. The fwup path is the same shape after the archive fix on the thread at line 208. So `--machine` is now a gate on the artifact rather than on a directory listing - which I think is the property you were after, though by a different route than checking the tree up front.

scripts/avocado-flash
jetm: Fixed in `b97bc79`. Both halves, and the first one reproduces.

The old resolver, run in a workspace holding `build-imx93-frdm` and `build-qemuarm64`, returned `<ws>/build-imx93-frdm/build/tmp/deploy` with no mention that a second tree existed - alphabetically first, exactly as you said, and `resolve_machine`'s guard silent because that directory held one machine.

It now collects every candidate at the nearest level that has any and refuses when there is more than one, naming both. A single tree still resolves. The guard lives here rather than in `resolve_machine` because this is where the ambiguity actually is - the layout you described defeats the downstream guard by construction.

The unquoted expansion is gone too: the roots arrive through a `while read` loop, so a path with a space resolves (verified) and a metacharacter cannot glob into a sibling. You were right that shellcheck says nothing - SC2046 is not raised for a substitution in a for-list.

One thing I added that you did not ask for, because the new refusal would otherwise misfire: paths resolve with `pwd -P` and de-duplicate. A `build` symlink pointing at the active `build-<machine>` directory is a common layout, and both names match the globs, so resolved logically it would look like two trees and trip the refusal over what is really one. Verified that case resolves to the single tree.

scripts/avocado-flash
jetm: Already fixed - `emmc_all`, not `emmc`, is what the current head runs, with both the bootloader and the image passed:

```
RUN_CMD=(sudo uuu -m "$usb_path" -b emmc_all "$boot" "$image")
```

so this thread is addressed rather than outstanding. Your `uuu -bshow emmc` evidence is what the comment above that line now records, including the detail that both built-ins default the image parameter to the bootloader when it is omitted, which is why the old form wrote imx-boot and nothing else while exiting 0.

On reusing `stone-provision-uuu-emmc.sh` instead of re-deriving the invocation: I did not, and I want to be straight that this is a judgement rather than an oversight. That script expands the archive and patches `devnum`/`mmcblk` for a provisioning flow that owns the whole sequence; this helper's job is to hand uuu a bootloader and an image it already resolved. Sharing it would mean either calling out to a script that does more than this needs, or factoring the uuu invocation out of it - the second is the right shape, and it is a change to a shipped provisioning path, so it belongs in its own PR rather than folded in here.

The duplication you raised on the other thread is real and I am not claiming otherwise, just declining to resolve it by importing the larger script's assumptions.

scripts/avocado-flash
jetm: Fixed in `b97bc79` - both halves, and I took the refuse-when-ambiguous option rather than adding a flag.

`flash_uuu` now enumerates the NXP devices in SDP mode, and the three cases behave as they should: none dies with the boot-switch hint, one pins `-m` to that USB path and prints it, and two or more refuses with both listed rather than letting uuu take whichever enumerated first. That closes the second-board-on-the-bench case directly.

I did not add a `--usb-path` option to name a target when several are present. Drafting it, I wrote that suggestion into the error message and then removed it, because a flag mentioned in an error and not implemented is worse than no flag - refusing until one board is left is honest and needs no new surface. Easy to add later if the bench case turns out to be routine.

The missing confirmation is fixed too: this path now calls `confirm_destructive` with the pinned USB path and the resolved image, so `-y` means something here for the first time. That required teaching the prompt not to assume its target is a block device - it runs `lsblk` only when the target is one, since a USB path is not.

And you were right about the header. It claimed the transport selects the target; a vendor-ID match does not, so that sentence is gone rather than reworded.

scripts/avocado-flash
jetm: Fixed in `b97bc79`, with one correction to the proposed mechanism.

The ordering bug is real and gone: `${MACHINE_NAME#avocado-}-rootdisk.zip` matched nothing, so only the fallback did any work. It now globs the machine's own name, and the hardcoded `.zip` is gone too, so a machine shipping `.fw` resolves instead of dying. Verified: `.zip` resolves, `.fw` resolves, two present refuses rather than picking, and none present says what it looked for.

**The manifest does not carry the answer.** I checked `meta-avocado-nxp/stone/avocado-imx/rootdisk.conf` - it is the fwup partition layout (`AVOCADO_IMAGE_ROOTFS`, per-partition `name = ${...}` blocks), not an index naming the output artifact. So reading it would not have told us the filename, and I did not build that.

What that leaves: for machines whose artifact genuinely is not a fwup archive - your `.tar.gz`, extensionless and `.img` cases - `flash_fwup` still cannot write them, because fwup takes a zip-based archive and those are different artifacts, not differently-named ones. `.img` in particular is the unpacked raw image, which is what the bmaptool and raw paths take, so accepting it here would fail further in and less legibly. The error now says so and points at `--backend bmaptool` rather than sending the user to re-run a build that already succeeded, which was the part that actually wasted people's time.

If some of those 17 do produce a real fwup archive under a name the glob still misses, that is a live gap and worth its own thread with the filename in it - I could not confirm one from the manifests.

scripts/avocado-flash
jetm: Fixed in `b97bc79`. `bmap="$image.bmap"` rather than stripping the extension first, so it looks for `<image>.wic.bmap` - the name `image_types.bbclass:357` actually emits.

I also dropped `--nobmap` from the fallback, which is the half that was doing active harm. As you note, bmaptool's own `find_and_open_bmap()` looks for exactly that name, so passing `--nobmap` disabled discovery that works; without it, a bmap this function did not predict still gets used, and a genuinely absent one still degrades to a full copy.

The verification point is the one that made this worth fixing over losing sparse skipping: with no bmap, `BmapCopy` yields `chksum = None`, so the per-region SHA256 the script advertises was forfeited on every single run. The success path now says `bmap: <path> (per-region checksums verified)` so the difference is visible rather than inferred.

The `IMAGE_FSTYPES:remove = "wic.bmap"` line in `meta-avocado-rockchip` you pointed at is what settles the naming, and it also means rockchip deliberately has no bmap - which the fallback message now states plainly instead of reading as a broken build.

scripts/avocado-flash
jetm: Fixed in `b97bc79`. The payload resolves first, appears in the prompt, and only then does it ask.

The `sd` arm now resolves the artifact per backend before `confirm_destructive`, and passes it in, so the prompt reads:

```
=== writing the whole OS to /dev/sdb (destructive) ===
  payload: .../stone/_build/avocado-imx93-frdm-rootdisk.zip
NAME   SIZE TYPE RM MODEL
...
type the target again to confirm:
```

A bundle-only rebuild now fails with `no fwup archive` before the banner, not after the operator has retyped the device.

Your framing of the cost is what decided how to fix it rather than just reordering two lines: since the retype gate is the only thing between a typo and a wiped disk, it now shows what is about to be written as well as where. Clearing it should require reading something, and previously there was nothing there to read that the script had actually established.

`flash_fwup` and `flash_bmaptool` still resolve for themselves when called without a payload, so neither function became dependent on the dispatch order to work.

mobileoverlord (approved):

PR #238, opened 2026-07-08
secure boot: LUKS2-encrypted /var sealed to the TPM, ported to wrynose

yoctopidg3:
## Target branch

Open this against **[`avocado-linux/meta-avocado:wrynose`](https://github.com/avocado-linux/meta-avocado/tree/wrynose)**.

Unlike the scarthgap PR (which stacks on Javier's `imx93-secure-boot-and-var-encryption` base), there is no equivalent base branch on wrynose yet, so **this PR carries the whole feature**: Javier Tia's foundational secure-boot / var-encryption work (27 commits, ported forward) plus the enablement, QEMU bring-up, and fixes on top (22 commits).

- base: [`avocado-linux/meta-avocado`](https://github.com/avocado-linux/meta-avocado) @ `wrynose`
- head: [`yoctopidg3/meta-avocado`](https://github.com/yoctopidg3/meta-avocado) @ `pidge/wrynose-secureboot`
- [compare / open PR](https://github.com/avocado-linux/meta-avocado/compare/wrynose...yoctopidg3:pidge/wrynose-secureboot)

## What this adds

1. **Encrypted `/var`** as a LUKS2 container, formatted first boot and unlocked in the initramfs (`cryptsetup-var.service`), across `qemux86-64`, `qemuarm64`, `intel-x86-64`, and `imx93-frdm`.
2. **TPM2 seal to PCR 7** (Secure Boot state) via `systemd-cryptenroll`, with an Argon2id recovery keyslot retired only after a verified TPM2 unseal.
3. A **boot-method abstraction (`AVOCADO_BOOTLOADER`)** so boot-artifact build deps are selected by boot method (`uboot` / `uefi`), not machine name.
4. An **OP-TEE firmware-TPM (fTPM)** on `qemuarm64`, which has no discrete TPM.
5. The **wrynose port fixes**, including the one that makes measured boot actually work (see below).

## Why we're doing this

**Encrypt `/var`, not the whole disk.** The rootfs is read-only and verified; the mutable state that needs confidentiality at rest lives in `/var`.

**Seal the disk key to TPM2 PCR 7.** PCR 7 records the platform's Secure Boot state. Sealing `/var`'s key to it means the volume only auto-unlocks when the machine boots in the same trusted Secure Boot configuration it was enrolled in; change that policy and the PCR differs, the unseal fails, and the volume will not silently open. The key is never on disk in the clear.

[...]

PR #265, opened 2026-08-11
Port the secure-boot and /var-encryption stack onto wrynose

jetm:
## Problem

PRs #232 and #236 target `scarthgap`, and `scarthgap` cannot boot the FRDM board the work is named for. Its 6.6.36 SPL stalls before `Normal Boot` on Rev1.2 silicon regardless of boot medium, so the hardware validation those PRs are waiting on is not possible on their own base. That validation has never run, which is also why a published field note claimed verified hardware secure boot that had never happened (corrected in peridio/docs#477).

## Solution

Carry the stack onto the wrynose BSP the board actually runs, so the hardware pass can happen while #232 and #236 stay where they are. Stacked on #259, the same shape as #236 on #232.

`scarthgap` and `wrynose` have separate roots rather than a common ancestor, so this is one net-diff port rather than a replay of the 56 commits. 28 files are new and land untouched; of the 31 the stack modifies, only 11 had also moved on wrynose, and 9 of those are qemu and x86 targets the imx93 kas config never pulls in.

## Key changes

- Port the stack onto wrynose, resolving 5 conflicts (2 qemu-only, taken to the wrynose side since that layer is not in the imx93 build)
- Fix stale `WORKDIR` references. The stack predates Yocto's split of unpacked sources into `UNPACKDIR`, so three recipes would have applied cleanly and then failed `do_install`, and the kernel bbappend would have globbed nothing and produced a kernel with no dm-crypt and no error to read. `sb-keys` keeps `WORKDIR` for its generated key directory, which is output rather than unpacked source
- Set `AVOCADO_VAR_PART_DEV` for imx93. Both qemu machines point `/var` at `/dev/mapper/var` when `encrypted-var` is on; imx93 had no such line and fell through to the base-files raw-PARTUUID default, so the container would have been created and opened in the initrd while fstab mounted `/var` from the raw partition underneath it. imx93 was the only `encrypted-var` target missing this
- Make dm-crypt reachable. The fragment asked for `CONFIG_DM_CRYPT=y` and silently got `=m`: kconfig caps a tristate at the weakest symbol it depends on, and both `BLK_DEV_DM` and `TRUSTED_KEYS` were `=m` in the imx defconfig. The initramfs packagegroup ships `cryptsetup` with no kernel module and nothing that modprobes one, so the unlock had no dm-crypt to call
- Commit two build fixes that existed only as uncommitted working-tree changes in one checkout: the rust-native snapshot `libz.so.1` rpath, and ganesha's nativesdk D-Bus removal. Both block a fresh clone of this branch from building; the ganesha one had already been rediscovered once in a previous session
- Guard `block_size` in the qemux86-64 provisioning script, which read it from the manifest and ignored it, matching what the rockchip sibling already does

## Reviewer notes

Draft: the hardware pass has not run yet. `avocado-core` builds clean (1038/1038 tasks, 0 errors) and the kernel config was verified against the produced `.config` rather than the fragment text, but nothing has been flashed to a board.

The two imx93 defects above are not port artifacts. They are present on #232, #236 and the July verification branch alike, and belong upstream in those PRs once the board pass confirms them. Both were invisible on the qemuarm64 proxy, which has the machine-conf line imx93 lacks.

Neither had a falsifier that could catch it. Task 2.1 greps the fragment file for `CONFIG_DM_CRYPT=y`, so it passes on the text while the built kernel disagrees.

Two known issues reported by another engineer on 2026-08-04 are not addressed here and will hit the boot pass first: `var-key.sh` reads its secret from the volume it is unlocking, and `var.img` provisioning fails on the merged branch. Those go before my own findings.

`avocado-distro` does not build on this branch, blocked by a pre-existing `wlroots 0.20` / `libdrm 2.4.127` version skew unrelated to secure boot. `avocado-core` is unaffected and is the right target for a boot image.


jetm:
Unified with #238 into #271. The two were the same port done independently, each carrying fixes the other lacked. #271 has both histories rebased onto #259 with every fix folded in, and takes #238's `AVOCADO_BOOTLOADER` gating over the version committed here.

Closing as superseded by #271.

PR #266, opened 2026-08-11
Keep the SSH host identity, and stop the extension-merge units racing

jetm:
## Problem

Two defects that only show up when you boot a locally built image on real hardware, both found bringing up an i.MX93 FRDM.

**The device hands out a new SSH host key on every boot.** `read_only_rootfs_hook` finds no pre-generated key in `/etc/ssh` and concludes the device cannot keep one, so it writes `/etc/default/ssh` pointing sshd at `sshd_config_readonly` with keys under `/var/run/ssh`, on tmpfs. Every client's `known_hosts` breaks on every reboot, and `avocado container dev up` bootstraps the device over SSH, so a session cannot reconnect to a device that has rebooted. The inference is right for a stateless image and wrong for this one: the rootfs is read-only but `/var` is writable and persistent, which is a third case oe-core does not model.

**Both extension-merge units run at once in the initrd and race.** `avocado-extension.service` and `avocado-extension-initrd.service` are the same unit apart from a description and one condition. The initrd variant is gated on `ConditionPathExists=/etc/initrd-release`; the general one had no complementary gate, so nothing kept it out of the initrd. Observed 14 ms apart:

```
Starting Avocado merge extensions (initrd)...
Starting Avocado merge extensions...
```

Both then set up loop devices for every extension, and whichever reaches `systemd-confext` second exits 1 with `Hierarchy '/etc' is already merged`.

## Solution

The shipped `sshd_config` already points `HostKey` at `/var/lib/ssh`, so only the override needs undoing: append the persistent values to `/etc/default/ssh`. `sshd_check_keys` sources it as shell so the last assignment wins, and its `generate_key` does `mkdir -p` on the key directory. Keys are generated once on first boot and reused after.

For the race, give the general unit the complementary gate its sibling already has.

## Key changes

- `/etc/default/ssh` gets `SYSCONFDIR=/var/lib/ssh` and an empty `SSHD_OPTS`, appended from `IMAGE_PREPROCESS_COMMAND`
- `avocado-extension.service` gains `ConditionPathExists=!/etc/initrd-release`

## Reviewer notes

The image hook runs from `IMAGE_PREPROCESS_COMMAND`, not `ROOTFS_POSTPROCESS_COMMAND`, and the distinction is load-bearing. A recipe-level append to the latter does not order after the entry `rootfs-postcommands.bbclass` contributes: measured on this image, `read_only_rootfs_hook` runs last of the three. Placed there the function ran before oe-core had created the file at all, its guard skipped, and the volatile values were written afterwards - a silent no-op that inspected as a working patch.

The failed merge unit is the visible half and the less interesting one. Two processes racing to publish the same hierarchies means the set of extensions in place when `sysinit.target` is reached is not the same on every boot, and both units are ordered before `systemd-userdb-load-credentials.service`, so a user record backing sshd's `AuthorizedKeysCommand` can be present on one boot and absent on the next.

Both verified on hardware, on a wrynose-based build of the same sources:

- Host keys: identical `ssh_host_ed25519_key.pub` hash and an unchanged 11:13 mtime observed across a reboot **and** a full rootfs+initramfs replacement, read off the card offline. Before the fix the same board produced a new key every boot.
- Merge race: units now start and finish sequentially with no `already merged`, and the boot shows four loop devices (one per extension) rather than eight in doubled pairs.

Neither fix is board-specific; both files are shared across targets, which is why this targets `scarthgap` rather than the BSP branch the bugs were found on.

### DIFF
--- a/meta-avocado-distro/recipes-avocado/avocadoctl/avocadoctl/avocado-extension.service
+++ b/meta-avocado-distro/recipes-avocado/avocadoctl/avocadoctl/avocado-extension.service
@@ -1,6 +1,24 @@
 [Unit]
 Description=Avocado merge extensions
 
+# Not in the initrd: avocado-extension-initrd.service covers that, gated on the
+# complementary ConditionPathExists=/etc/initrd-release. Without this line the
+# two units are eligible at the same moment - identical Requires, identical
+# After=local-fs.target avocadoctl.socket - so systemd starts both inside the
+# initrd and they race. Each sets up loop devices for every extension (the
+# doubled loop0..loop7 at matching sizes in a boot log is the signature), and
+# whichever reaches systemd-confext second exits 1 with
+#
+#   Command 'systemd-confext' exited with error code Some(1):
+#   Hierarchy '/etc' is already merged.
+#
+# The loser's failure is not cosmetic. The two units race to publish the same
+# hierarchies, so which extensions are actually in place when sysinit.target is
+# reached varies between boots, and anything reading them that early - a user
+# record backing sshd's AuthorizedKeysCommand, for one - sees an inconsistent
+# view depending on who won.
+ConditionPathExists=!/etc/initrd-release
+
 ConditionDirectoryNotEmpty=|/var/lib/avocado
 ConditionCapability=CAP_SYS_ADMIN
 
--- a/meta-avocado/recipes-avocado/images/avocado-image-rootfs.bb
+++ b/meta-avocado/recipes-avocado/images/avocado-image-rootfs.bb
@@ -30,5 +30,42 @@ create_dirs() {
     mkdir -p ${IMAGE_ROOTFS}/opt
 }
 
+# Keep the SSH host identity across reboots.
+#
+# oe-core's read_only_rootfs_hook finds no pre-generated key in /etc/ssh and
+# concludes the device cannot keep one, so it writes /etc/default/ssh pointing
+# sshd at sshd_config_readonly with keys under /var/run/ssh - tmpfs. The device
+# then presents a NEW host key on every boot: every client's known_hosts breaks
+# on every reboot, and `avocado container dev up`, which bootstraps the device
+# over SSH, cannot reconnect to a device that has rebooted.
+#
+# The inference is sound for a stateless image and wrong for this one. The rootfs
+# is read-only but /var is writable and persistent, which is the third case
+# oe-core does not model - its choice is between shipping a key in the image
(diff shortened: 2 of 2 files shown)


jetm:
After talking more about this change @mobileoverlord we found out we are missing systemd changes from `scratchgap` branch, which these changes might not be required for.

jetm:
Closing this. The case for it is speculative right now, and the one branch where I could test it says the defect does not occur.

**On wrynose, neither symptom reproduces.** Measured on an i.MX93 FRDM booting a wrynose image, not inferred from the recipes:

```
# cat /etc/default/ssh
(no such file)

# sshd -T | grep hostkey
hostkey /var/lib/ssh/ssh_host_rsa_key
hostkey /var/lib/ssh/ssh_host_ecdsa_key
```

No `/etc/default/ssh` means nothing redirects sshd at `sshd_config_readonly`, and the host keys are already on the writable, persistent `/var` - which is the end state this PR exists to produce. The image half here is guarded by `if [ -f /etc/default/ssh ]`, so on that branch it cannot fire even when applied.

**Both fixes are already on wrynose** via #267, so nothing is lost by closing this: `persist_ssh_host_keys` is in `avocado-image-rootfs.bb` and `avocado-extension.service` carries its conditions there.

**On the systemd question** raised in the discussion with @mobileoverlord - the direction is the opposite of what we assumed. wrynose is not missing scarthgap's systemd work:

| branch | systemd |
| --- | --- |
| scarthgap | `systemd_258.1.bb`, `systemd-boot_258.1.bb`, … - a fork carried onto an older oe-core |
| wrynose | `systemd_%.bbappend` only - 259.x comes from Yocto 6.0 natively |

The fork exists to backport onto scarthgap. wrynose supersedes it rather than lacking it, so "port the systemd changes and this becomes unnecessary" does not apply there.

**What is still open, and why that argues for closing rather than merging:** whether scarthgap itself still needs this. That depends on whether a current scarthgap image build still produces `/etc/default/ssh` with the readonly override, and I have not rebuilt scarthgap to check. There is also a loose end on the wrynose side worth recording: oe-core's `read_only_rootfs_hook` writes that file whenever `/etc/ssh` is a directory, and it *is* one in the wrynose rootfs - so by the code the file should exist, and on the device it does not. I verified the outcome, not the mechanism.

If a scarthgap build turns up `/etc/default/ssh` carrying `SSHD_OPTS='-f /etc/ssh/sshd_config_readonly'`, this is worth reopening with that output attached - at that point the change is demonstrated rather than presumed.

PR #268, opened 2026-08-11
kas: make image content opt-in via feature groups

jetm:
Ports the scarthgap feature-groups design (#229) and minimal-by-default (#247) to `wrynose`.

## Why both, in one PR

On scarthgap these landed separately: #229 introduced the mechanism with boards still including the umbrella, and #247 removed it afterwards. Doing only the first here would move ~250 lines of recipe around and change nothing anyone builds, because `kas/target/distro.yml` includes `kas/extra.yml`, which reaches every board regardless of what the packagegroup says.

## What a default board builds now

| | before | after |
|---|---|---|
| Unconditional feed packages | 98 | **12** |
| Total incl. gated bundles | 181 | 12 + opted-in groups |
| Layers force-pulled by `distro.yml` | 7 | 0 |
| Boards pulling the full umbrella | 26 | 0 |

`kas build kas/machine/<board>.yml` is minimal. Append `:kas/feature/<group>.yml` per group, or `:kas/feature/complete.yml` for the full set.

## Deltas from the scarthgap original

A straight cherry-pick breaks wrynose: #229 moves five vendor layers into `base.yml` pinned to `scarthgap`, two of which declare a `LAYERSERIES_COMPAT` older than wrynose.

- **Dropped from `base.yml`:** meta-atmel (`nanbield`), meta-erlang (`whinlatter`). `system-base` therefore drops `cryptoauthlib` and `livebook`.
- **Repinned:** meta-fwup `wrynose-avocado@0998acc`, meta-swupdate `wrynose@6616d4e`, meta-openjdk-temurin `main@01b411f`, meta-wayland `wrynose@059d0a3`.
- **Not in `complete.yml`:** `cameras`, `cloud-aws`, `ros`. Fragments keep their scarthgap pins on purpose, so including one fails loudly at parse instead of silently shipping less.
- **`graphics`** ships without wpewebkit/wpebackend-fdo/cog (meta-webkit has no wrynose branch).
- **Collision fixed** that scarthgap does not have: `containers.yml` and this branch's existing `virtualization.yml` both pinned meta-virtualization at different commits. `containers.yml` now includes the latter rather than re-pinning.
- **Pre-existing breakage fixed:** `imx95-frdm.yml` included `kas/extra/deepx.yml`, which this change deletes; repointed at `kas/feature/ai.yml`.

## Verification

- All five pinned layers are checked out locally at exactly those commits, and each declares `wrynose` in `LAYERSERIES_COMPAT`.
- **Nothing lost:** all 181 packages from the old monolith are covered by the base plus the 12 groups. Zero dropped, verified by set comparison.
- Every include across `kas/**/*.yml` resolves; all files valid YAML, no empty `includes:` keys.
- Measured which boards gain new full-variant CI coverage rather than trusting the ported comment: 4 here (alif-e8-devkit, qemuarm64, qemux86-64, stm32mp25-dk), not scarthgap's 5 — `imx8mp-var-dart` does not exist on wrynose. Comment corrected.

A board build is running now; I will report the result here.

The second commit fixes two long-standing MD034 bare-URL failures in the migration log. Unrelated to the port, separated for that reason.


mobileoverlord (commented):
Reviewed against the branch tree rather than the diff. Re-derived the main claims independently and they hold:

- **"Nothing lost"** — set-compared the old monolith against the new base + 12 groups. Nothing dropped. The two tokens that look dropped, `basler-common` and `meta-aws`, are `BBFILE_COLLECTIONS` layer names from the old gating expressions, not packages.
- **Token ↔ packagegroup** — 12 tokens, 12 `packagegroup-avocado-feature-*.bb`, exact correspondence, no orphan in either direction. Fragments that add only a layer (`ros`, `clang`, `python-ai`) correctly emit no token.
- **Includes** — 163 includes across 79 kas files all resolve; no dangling reference survives the `kas/extra*` removal.
- **Base pins** — read `conf/layer.conf` at each pinned commit: meta-fwup `0998acc`, meta-swupdate `6616d4e`, meta-openjdk-temurin `01b411f` all declare `wrynose`.
- The `containers.yml` → `virtualization.yml` include instead of a second pin is the right call, and the historical-paths callout at the top of the migration doc already answers the broken `kas/extra/...` links.

**One correction.** `kas/base.yml` and the PR description say meta-atmel declares `nanbield` and meta-erlang `whinlatter`. At the commits scarthgap actually pins — meta-atmel `b24958f1`, meta-erlang `62acd049` — both declare `scarthgap`. At current HEAD they are `scarthgap` and `blacksail`. Dropping them is still right, since neither declares `wrynose`, but the comment ends with "Restore all four together once the forks carry a wrynose branch" and points whoever does that at codenames the forks never had. Looks like the values came from the 2026-04-28 survey table, which surveyed the upstream layers rather than the avocado forks.

**The thing actually blocking a merge is CI.** It is not a branch filter — `pr-parse-container-sdk.yml` does not exist on `wrynose` at all. The only workflow there is `pr-build-labeled.yml`, which needs a `build-<machine>` label plus the `build` trigger (this PR has no labels), and its last run was 2025-08-27. Repo-level self-hosted runners report `total_count: 0`.

Worth porting the parse job to `wrynose` before this lands: it runs on GitHub-hosted `ubuntu-latest` in a container, so it needs no molcajete, and `kas/sdk/container-x86_64.yml` enables `meta-avocado` — so `bitbake -p` would parse all 12 new packagegroups *and* run the layer-compatibility check over the layers this PR moves into `base.yml`, which is the main risk here. That turns "reviewed and reasoned about" into a signal.

Also flagging that you offered a board build result on 11 Aug and it never landed here — if you have it, that plus a ported parse job would be enough for me.

Minor, and not for this PR: `complete.yml` enables 9 of the 12 tokens. `cameras` and `cloud-aws` are deliberate and documented; `ai` is omitted on both branches, so "the full set" excludes AI on scarthgap too.

mobileoverlord:
I pushed the forward-ported parse workflow onto this branch (`a68a1a9a`) so it would have CI. Both jobs went red immediately — **but not because of anything in this PR.** Flagging that up front so the red badge doesn't read as your problem.

```
ERROR - Branch "wrynose-avocado" in repository "openembedded-core"
        does not contain commit "709c4fd69a3476e79d3c37c2fe504ddc7a7cacbb"
```

It fails during repo checkout, before a single recipe is parsed. The pin lives in `kas/vendor/oe.yml`, which this PR does not touch — `git diff origin/wrynose...HEAD -- kas/vendor/oe.yml` is empty. It is pre-existing on `wrynose`.

What happened to the vendor fork:

- `709c4fd6` still exists, dated 2026-04-04
- `wrynose-avocado` now tips at `0c69588b`, same date, **same commit subject**
- `compare/wrynose-avocado...709c4fd6` returns `status=diverged`, `ahead=2 behind=176`

So the branch was rewritten and the pinned sha orphaned. `scarthgap-avocado` is `status=identical` against its pin, so this is wrynose-only.

The practical consequence is that **`wrynose` is currently unbuildable at the base** — any kas invocation on that branch dies at checkout. That is also why "a board build is running now" from 11 Aug presumably never produced a result worth posting.

This wants its own one-line PR repinning `openembedded-core` — deciding the correct commit is a call for whoever rewrote the branch, and a 176-commit oe-core move has no business riding along on a feature-groups port. Once that lands, this branch gets a real parse result and I can finish the review properly.

Net: the workflow port did its job on the first run — it found that the branch it was added to cannot build. Everything I checked in the port itself still looks right.

mobileoverlord (approved):
Green on both arches at `6f18f941` — the first passing CI this branch has had, and it now exercises the real thing: kas checks out every layer, and `bitbake -p` parses the twelve new feature packagegroups and runs the layer-compatibility check over what this PR moves into `kas/base.yml`.

Re-verified after merging the repin: 40 pins audited, none diverged; 12 feature packagegroups, `complete.yml` and the workflow all intact through the merge.

Everything from my earlier review stands — nothing lost in the package set, exact 12↔12 token/packagegroup correspondence, all includes resolve, base pins declare `wrynose` at their pinned commits.

Two things carried forward rather than blocking on them:

- `kas/base.yml` still says meta-atmel declares `nanbield` and meta-erlang `whinlatter`; at the commits scarthgap pins, both say `scarthgap`. Dropping them is still correct — neither declares `wrynose` — but the restore note points at codenames the forks never had. Worth a one-line fix whenever you next touch the file.
- Parse is not build. This proves recipes parse and layers are compatible; it cannot catch a packagegroup that parses cleanly and RDEPENDs on something unbuildable. The label-driven board build is still the only thing that would, and it needs the self-hosted runner.

Thanks for the port, and for the measurements in the description — the "nothing lost" set comparison held up when I redid it independently.

PR #269, opened 2026-08-12
Front-port three scarthgap changes wrynose is missing

jetm:
## Problem

Three changes landed on scarthgap after the last front-port and never crossed to
wrynose. A file-level comparison of the two branches puts 169 files in scarthgap
that wrynose does not have, but most of that is expected divergence rather than
work that is missing: a systemd 258.1 fork that wrynose does not need (it carries
259.5 natively), a full chromium recipe copy that wrynose replaces with
`dynamic-layers` bbappends, and layers with no wrynose branch at all. These three
are the part that is genuinely absent.

One of them is a functional regression rather than a missing convenience. `#256`
enabled the BPF cgroup device controller on two targets; only the x86-64 half is
on wrynose, so a wrynose qemu target cannot start containers that its scarthgap
twin can - `runc create` fails with `bpf_prog_query(BPF_CGROUP_DEVICE) failed:
function not implemented`.

## Solution

Cherry-pick the three, unchanged, with `-x` so each keeps a reference to the
scarthgap commit it came from. Nothing here needed adapting for wrynose: all
three applied without conflict, and the kernel fragment adds no line that the
wrynose target already had.

## Key changes

- `scripts/avocado-flash` (#264): the backend-dispatching image writer.
- `avocado-ensure-extensions.service` (#263): mirror the merge unit's conditions
  so a device with no extensions skips this unit quietly instead of failing its
  start job on every boot.
- `avocado-extra.cfg` for qemu and x86-64 (#256): BPF cgroup device controller
  for cgroup-v2 container targets.

## Reviewer notes

Verified rather than assumed, since a clean cherry-pick only proves the patch
applied:

- `shellcheck -S error` is clean on `avocado-flash` and `--help` runs.
- No duplicate `CONFIG_` line in either kernel fragment after the port (23 and 20
  lines, all unique) - worth checking because wrynose already carried part of
  `#256`.
- The service conditions landed against the same `BindsTo=` block they were
  written for.

Not included, and deliberately: the Variscite DART and both Jetson Orin devkit
machines are also scarthgap-only, but each pulls vendor layers whose wrynose
branches are unconfirmed - the same wall that keeps meta-atmel, meta-erlang and
meta-webkit off wrynose today. Those belong in their own PRs, after someone
checks the layers. The executorch/ML stack and a 7-file meson fork are likewise
absent; the meson fork may be a scarthgap-only build fix that wrynose has no use
for, which needs a judgement call rather than a port.

### DIFF
--- a/meta-avocado-distro/recipes-avocado/avocadoctl/avocadoctl/avocado-ensure-extensions.service
+++ b/meta-avocado-distro/recipes-avocado/avocadoctl/avocadoctl/avocado-ensure-extensions.service
@@ -1,6 +1,17 @@
 [Unit]
 Description=Reload manager and retrigger activation targets after Avocado extensions are merged
 Documentation=https://github.com/flatcar/init/pull/65
+
+# Mirrored from avocado-extension.service. BindsTo= requires the bound unit to
+# be active, and a condition-skipped unit is inactive - so without these, a
+# device with no extensions installed fails this unit's start job with
+# "Dependency failed" on every boot rather than skipping quietly alongside the
+# merge it exists to follow up. systemd evaluates conditions before the BindsTo
+# check, so matching them here suppresses the job without weakening the binding
+# for the case the merge genuinely fails.
+ConditionDirectoryNotEmpty=|/var/lib/avocado
+ConditionCapability=CAP_SYS_ADMIN
+
 BindsTo=avocado-extension.service
 After=avocado-extension.service
 DefaultDependencies=no
--- a/meta-avocado-qemu/recipes-kernel/linux/files/avocado-extra.cfg
+++ b/meta-avocado-qemu/recipes-kernel/linux/files/avocado-extra.cfg
@@ -2,6 +2,18 @@
 CONFIG_BRIDGE=m
 CONFIG_BRIDGE_NETFILTER=m
 
+# BPF cgroup device controller: on cgroup v2, runc/docker gate per-container
+# device access with a BPF_CGROUP_DEVICE program, so without BPF_SYSCALL +
+# CGROUP_BPF container start fails at runc create with:
+#   bpf_prog_query(BPF_CGROUP_DEVICE) failed: function not implemented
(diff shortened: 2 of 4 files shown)


mobileoverlord (approved):
Approving. Verified rather than assumed, since a clean cherry-pick only proves the patch applied:

- All four ported files are **byte-identical to scarthgap** after the port.
- The one real adaptation risk checks out: the ported unit mirrors `avocado-extension.service`'s conditions, and wrynose's own copy of that sibling carries exactly `ConditionDirectoryNotEmpty=|/var/lib/avocado` + `ConditionCapability=CAP_SYS_ADMIN`. Mirroring scarthgap's conditions onto a branch whose sibling had drifted is how this class of cherry-pick goes wrong; it didn't.
- No duplicate `CONFIG_` lines, 23 and 20 as claimed.

Two notes, neither blocking the merge.

**The `#256` sentence is backwards.** "only the x86-64 half is on wrynose" - neither half was. Both `avocado-extra.cfg` files had zero BPF lines before this PR, which is why the diff correctly adds to both. The change is bigger than advertised, and the duplicate-line check was guarding a condition that did not exist.

**"These three are the part that is genuinely absent" does not hold.** `#247` ("Default boards to minimal, opt into the full feature set") is entirely missing from wrynose - 16 `kas/feature/*.yml` and 12 `packagegroup-avocado-feature-*.bb`. wrynose's `kas/feature/` holds only the five pre-`#247` files, and `git log --grep` finds no trace of it on the branch. It is also not in the "not included, and deliberately" list. That may be a conscious call - it is a structural change, not a bug fix - but it is worth naming either way, so the next front-port does not re-derive the same comparison.

The other large deltas are not gaps and I would not chase them: `nodejs_22.22.2.bb`, `qemu_9.2.0.bb` and friends are version-pinned recipe forks, the same class the description already explains away with the systemd 258.1 fork. One exception worth a glance - `qemu-system-native_%.bbappend` is a bbappend rather than a version fork, so that reasoning may not cover it.

Merging with rebase rather than squash so the three `cherry picked from commit` references survive.

PR #270, opened 2026-08-12
fluentbit: carry the 5.0.8 recipe in this layer

jetm:
## Problem

meta-openembedded's scarthgap branch still ships Fluent Bit **1.9.7** while
upstream master is on the **5.x** series. An image built from this distro gets a
release several years old, and misses the journald input plugin that the
persistent `/var/log/journal` rolling-window logging design depends on.

## Solution

Carry the 5.x recipe in this layer, pinned to **v5.0.8** (SRCREV `c2b1cfd`),
rather than rewriting the recipe inside the vendor fork of meta-openembedded.

That placement is the point of this PR. The vendor fork exists to track upstream
and bump versions; a rewritten recipe inside it turns every future rebase onto
upstream into a reconciliation of our own divergence, and hides a local decision
inside what should read as a mirror. Kept here, the divergence is ours, visible
in this layer's history, and deletable in one commit the day upstream ships 5.x.

No `PREFERRED_VERSION` is required - BitBake prefers the highest version it can
see, so 5.0.8 here wins over 1.9.7 in the vendor layer.

## Key changes

- `meta-avocado-distro/recipes-extended/fluentbit/fluentbit_5.0.8.bb`, ported
  from upstream meta-oe master: git fetcher on the `v${PV}` tag rather than a 1.9
  release tarball, and `-DFLB_IN_SYSTEMD=On` pinned so the journald input plugin
  is always built.
- Four patches alongside it: two refreshed carry-overs from the 1.9.x set and two
  new for 5.x. The twelve stale 1.9.x patches are simply not carried.
- A header comment on the recipe recording why it lives here and what makes it
  removable later.

## Reviewer notes

The recipe is self-contained - no `require` of a meta-oe include, and it inherits
only `cmake`, `systemd` and `pkgconfig` - so it needs nothing from the layer it
came from.

Every patch referenced by `SRC_URI` is present in the directory and none is left
over; I diffed the two lists rather than eyeballing them.

Not build-tested on a target yet. The recipe is a straight port of upstream
meta-oe master with no Avocado-specific edits, so the risk sits in the 1.9.7 ->
5.0.8 jump itself (service renamed `td-agent-bit.service` -> `fluent-bit.service`)
rather than in this move between layers.

This supersedes avocado-linux/vendor-meta-openembedded#1, which made the same
change in the wrong repository; that PR will be closed pointing here.

### DIFF
--- a/meta-avocado-distro/recipes-extended/fluentbit/fluentbit/0001-lib-Do-not-use-private-makefile-targets-in-CMakelist.patch
+++ b/meta-avocado-distro/recipes-extended/fluentbit/fluentbit/0001-lib-Do-not-use-private-makefile-targets-in-CMakelist.patch
@@ -1,0 +1,103 @@
+From 8d93bb7c866f88dd7f2cac81d2acfcffae9d555d Mon Sep 17 00:00:00 2001
+From: Niko Mauno <niko.mauno@vaisala.com>
+Date: Sun, 29 Sep 2024 12:00:00 +0000
+Subject: [PATCH] lib: Do not use private makefile targets in CMakelists.txt
+
+By extending the scope of changes introduced in commit
+fc325524d50fe179b76f127243ab9e03ddbdaaa4
+("build: CMakeLists.txt Do not use private makefile targets (#5819)")
+we mitigate the following error produced by BitBake in Yocto
+
+  ERROR: fluentbit-3.1.9-r0 do_package_qa: QA Issue: File /usr/bin/fluent-bit in package fluentbit contains reference to TMPDIR [buildpaths]
+  ERROR: fluentbit-3.1.9-r0 do_package_qa: Fatal QA errors were found, failing task.
+
+stemming from
+
+  $ strings packages-split/fluentbit/usr/bin/fluent-bit
+  ...
+  $(subst /yocto/upstream/build/tmp/work/cortexa57-poky-linux/fluentbit/3.1.9/git/,,$(abspath $<))
+  ...
+
+Signed-off-by: Niko Mauno <niko.mauno@vaisala.com>
+
+Upstream-Status: Submitted [https://github.com/fluent/fluent-bit/pull/9450]
+---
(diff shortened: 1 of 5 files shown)


mobileoverlord (approved):
Approving. Verified rather than taken on trust:

- `fluentbit_5.0.8.bb` is byte-identical to upstream meta-oe master's recipe except the header comment and the `FLB_IN_SYSTEMD` block.
- All four patches are byte-identical to upstream's.
- `SRCREV c2b1cfd…` is exactly upstream tag `v5.0.8` per the GitHub ref API, so the pin is honest.
- The `FLB_IN_SYSTEMD` addition is keyed to the same `DISTRO_FEATURES` check that gates `systemd` entering `DEPENDS`, so the no-systemd container distro can't hit a configure failure hunting a libsystemd that was never a dependency.
- `grep` finds no reference to `fluentbit`, `fluent-bit`, or `td-agent-bit` anywhere in this repo outside the recipe directory — no image or packagegroup consumes it, so the 1.9.7 → 5.0.8 jump and the service rename are unreachable from here today.

Merging with the untested status stated plainly rather than papered over. Nothing has compiled this recipe: the author's note says so, and the label-gated build wouldn't have either — the CI matrix builds `avocado-complete`, which never pulls fluentbit in, so a green run would have proven the recipe parses and nothing more. Blast radius today is parse-time only, and the sole novel line is a standard `bb.utils.contains`. The first change that puts fluentbit in an image is where this actually gets tested.

Two notes for a follow-up, neither blocking:

- `tag=v${PV}` is inert on scarthgap. bitbake 2.8.1's `fetch2/git.py` has no `tag` parameter handling and unknown params are silently ignored, so `SRCREV` alone pins the checkout. Correct today, but it reads as a second pin that isn't one, and it starts meaning something on a bitbake bump.
- The reviewer note says "no Avocado-specific edits" when there is exactly one — the `FLB_IN_SYSTEMD` block.

PR #271, opened 2026-08-12
Secure boot and LUKS2-encrypted /var, unified on wrynose

jetm:
## Problem

The secure-boot and encrypted-`/var` work has been spread across four PRs on two
branches, two of which port the same stack to wrynose independently. Neither
wrynose branch had all the fixes: each carried defects the other had already
solved, and #238 has been conflicting against wrynose since late July.

## Solution

One branch per release branch. This is the wrynose one: the full history from
both wrynose PRs, rebased onto the imx93 wrynose BSP, with every fix that
existed on only one side folded in. Commits keep their original authorship.

Basing on #259 rather than on wrynose directly is deliberate. Scarthgap's
6.6.36 SPL stalls before `Normal Boot` on Rev1.2 FRDM silicon regardless of
boot medium, so the hardware pass this stack has been waiting on is only
possible on the NXP wrynose BSP.

## Key changes

- Set `AVOCADO_VAR_PART_DEV` on imx93. It was the only `encrypted-var` target
  missing the line, so fstab mounted `/var` from the raw partition while
  cryptsetup-var created and opened the LUKS container underneath it
- Promote `BLK_DEV_DM` and `TRUSTED_KEYS` to `=y`. kconfig caps a tristate at
  its weakest parent, so the fragment's `CONFIG_DM_CRYPT=y` was silently
  demoted to `=m` and the initrd unlock had no dm-crypt to call
- Read `gen-sbkeys.sh` from `UNPACKDIR` in sb-keys; `WORKDIR` no longer holds
  unpacked sources on this branch and `do_compile` failed on the chmod
- Restore the var seed to the imx93 stone manifest, fixing the `var.img`
  provisioning failure reported on 2026-08-04
- Reject a `block_size` the qemux86-64 provisioning script cannot honour,
  rather than reading it from the manifest and laying the image out in 512-byte
  units anyway

## Hardware verification

Built and booted on FRDM-IMX93 (`avocado-imx93-frdm`, wrynose BSP), 7975 tasks, 0 errors. **Encrypted `/var` now works on this board, which it never has before.**

```
/dev/mapper/var is active and is in use.
  type:    LUKS2   cipher: aes-xts-plain64   keysize: 512 bits
  device:  /dev/mmcblk1p8
```

Each fix, checked against the produced artifact rather than the source:

| Fix | Evidence on the board |
|---|---|
| `CONFIG_DM_CRYPT=y` | `dmsetup targets` lists `crypt v1.28.0`; `lsblk` shows `mmcblk1p8 -> var 29G crypt /var`. The pre-fix build of the same machine had `BLK_DEV_DM`, `DM_CRYPT` and `TRUSTED_KEYS` all at `=m` |
| `AVOCADO_VAR_PART_DEV` | `findmnt /var` -> `/dev/mapper/var`, not a raw PARTUUID |
| SoC-UID key provider | `/sys/devices/soc0/serial_number` = `b5e6d147d14f24b47192b4911b023cc0`, the 128-bit OCOTP UID, and no `provisioned secret not found` anywhere in the journal |

First boot ran the whole path: `opening LUKS2 container` -> `no filesystem - creating BTRFS (first boot)` -> filesystem created at 29.01 GiB -> `/dev/mapper/var ready` -> `Finished LUKS2 /var unlock and first-boot format`. The partition expanded 112M -> 29G as designed.

A second boot exercised the **unlock** path rather than first-boot format, and `/var` reopened with the same btrfs UUID (`048607b0-...`) across a confirmed reboot (`boot_id` changed). So the SoC-UID key derives reproducibly on the same board, which is the property the whole scheme rests on.

## Reviewer notes

The `var-key.sh` chicken-and-egg reported on 2026-08-04 is fixed here, not deferred. imx93 was the only `encrypted-var` machine with no key provider of its own, so it fell through to the shared one, which reads `/var/private/var-key-secret` - a file on the volume being unlocked, not mounted when the provider runs in the initramfs, and which **nothing in the tree ever creates**. The Argon2id path had therefore never once succeeded on this board.

The replacement derives from the SoC UID and is deliberately weaker than the secret+hw-id design it replaces: a UID-derived key is device-*binding*, not a secret, since anyone who can read the UID can rederive it. That is stated in the commit and in the file rather than left implicit. It is accepted because the stronger design has no provisioning path at all today, and because the Argon2id slot is the documented *recovery* keyslot - a TPM-sealed primary is phase 2. There is no discrete TPM on this board; the route there is the OP-TEE fTPM, which needs `optee-ftpm-init` (currently `COMPATIBLE_MACHINE = "avocado-qemuarm64"`) ported, and no AHAB work.

Testing needs a shell on the board, which the stock image does not give: see #275 for the `ssh-dev` feature that makes that repeatable rather than per-board.

Depends on #259 and cannot merge before it.

Supersedes #238 and #265, which are closed.


jetm:
Doing more testing before it's Ready for Review...

meta-avocado-x86-64/stone/stone-intel-x86-64-v3.json
```diff
@@ -70,7 +69,7 @@
         { "name": "recovery", "partition_type": "EF00", "size": 256, "size_unit": "mebibytes" },
         { "name": "rootfs-a", "image": "rootfs", "partition_type": "8304", "size": 512, "size_unit": "mebibytes" },
         { "name": "rootfs-b", "partition_type": "8304", "size": 512, "size_unit": "mebibytes" },
-        { "name": "var", "image": "var", "partition_type": "8300", "partition_uuid": "4d21b016-b534-45c2-a9fb-5c16e091fd2d", "expand": "true" }
+        { "name": "var", "image": "", "partition_type": "8300", "partition_uuid": "4d21b016-b534-45c2-a9fb-5c16e091fd2d", "expand": "true" }
```
nicksinas: **Major** — five machines now ship an empty `var` partition that nothing formats.

4bdf9b4 restores the imx93 seed because "the removal assumed /var could be created empty and formatted on first boot. It cannot" — no trust root, no avocado-conn, no avocado-grow-var, no route back. That argument is machine-independent, but the seed is still dropped here and in `stone-intel-x86-64-v2.json`, `stone-intel-x86-64-v4.json`, `stone-qemuarm64.json:137`, `stone-qemux86-64.json:108`, and `stone/qemuarm64/rootdisk.conf:204`.

`cryptsetup-var.sh` is the only thing in the tree that ever runs `mkfs.btrfs` on /var, and it is gated on `encrypted-var`, which no kas machine config enables. So on a default build of any of those five, /var has no filesystem and no LUKS header, and `x-systemd.growfs` only grows an existing fs. Restore the seed on all of them, or make the removal conditional on the feature that provides the formatter.

meta-avocado-x86-64/conf/machine/include/avocado-x86-64.inc
```diff
@@ -7,6 +7,7 @@ KERNEL_IMAGETYPE = "bzImage"
 SERIAL_CONSOLES ?= "115200;ttyS0"
 
 EFI_PROVIDER = "systemd-boot"
+AVOCADO_BOOTLOADER = "uefi"
```
nicksinas, line 10: **Major** — intel-x86-64 never sets `AVOCADO_VAR_PART_DEV`.

c64c673 says it "mirrors the qemu enablement" and adds a DMI key provider, but not the `AVOCADO_VAR_PART_DEV` line both qemu machines and imx93 carry. With `encrypted-var` on, cryptsetup-var opens `/dev/mapper/var` while fstab still mounts the raw `PARTUUID=4d21b016-…` base-files default — the exact defect this PR's headline fix corrects for imx93. `cryptsetup-var-udev` is skipped too, since `packagegroup-avocado-rootfs.bb:63` gates on the same variable.

meta-avocado/recipes-kernel/linux/files/dm-crypt.cfg
```diff
@@ -0,0 +1,5 @@
+CONFIG_DM_CRYPT=y
```
nicksinas, line 13: **Major** — the shared fragment lacks the parent-symbol promotions the imx93 one needed.

`meta-avocado-nxp/recipes-kernel/linux/files/imx93-frdm/dm-crypt.cfg` documents that `CONFIG_DM_CRYPT=y` is silently demoted to `=m` unless `BLK_DEV_DM` and `TRUSTED_KEYS` are both `=y`. This fragment — the one every linux-yocto machine gets — sets only `DM_CRYPT=y`. Any of those defconfigs with `BLK_DEV_DM=m` reproduces the bug the PR is fixing, and none of those machines had a boot pass. Add both promotions here and drop the duplication in the imx93 copy.

meta-avocado-x86-64/recipes-core/cryptsetup-var/files/var-key.sh
```diff
@@ -0,0 +1,34 @@
...
+# DMI identifiers are populated by the kernel early and are readable in the
+# initramfs. product_uuid is per-board unique; fall back to product_serial.
+HW_ID=""
+for f in /sys/class/dmi/id/product_uuid /sys/class/dmi/id/product_serial; do
+    if [ -r "$f" ]; then
+        HW_ID=$(tr -d '\0\n' < "$f")
+        [ -n "$HW_ID" ] && break
+    fi
+done
+[ -n "$HW_ID" ] || HW_ID="intel-no-dmi-id"
```
nicksinas: **Major** — this can give a whole fleet one identical /var key.

3901868's own reasoning: "on real hardware a constant would give every board in the fleet the same /var key, silently, with no symptom until someone noticed one device's disk opening on another." This is real hardware and it falls back to the constant `intel-no-dmi-id`.

Worse, the fallback often isn't even reached: `product_serial` is commonly `Default string`, `To Be Filled By O.E.M.` or `0123456789` on whitebox boards, so a non-empty placeholder becomes the key material for every identical box. Refuse the way imx93 does, and reject known DMI placeholders rather than treating non-empty as unique. Latent until `encrypted-var` is switched on, which is the stated plan.

meta-avocado/recipes-kernel/linux/files/dm-crypt.cfg
```diff
@@ -0,0 +1,5 @@
...
+CONFIG_CRYPTO_AES=y
+CONFIG_CRYPTO_XTS=y
+CONFIG_CRYPTO_SHA256=y
+CONFIG_HW_RANDOM_TPM=y
```
nicksinas: **Medium** — this contradicts `tpm.cfg` / `ftpm.cfg` in the same PR.

2af402d turns `CONFIG_HW_RANDOM_TPM` *off* in `meta-avocado-qemu/recipes-kernel/linux/files/tpm.cfg` and `ftpm.cfg` because the hwrng kthread holds `tpm_mutex` across a GetRandom loop and hangs the PCR-7 enroll forever. On qemu with `encrypted-var` on, both fragments land in the same `.config` and the winner is bbappend/SRC_URI order.

It also has nothing to do with dm-crypt — it looks like a leftover from the old `dm-verity.cfg.in`. Drop the line.

meta-avocado/recipes-kernel/linux/linux-yocto_%.bbappend
```diff
@@ -0,0 +1,8 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+SRC_URI:append = "${@bb.utils.contains('DISTRO_FEATURES', 'secureboot', ' file://dm-verity.cfg', '', d)}"
+SRC_URI:append = "${@bb.utils.contains('DISTRO_FEATURES', 'security', ' file://modsign.cfg', '', d)}"
```
nicksinas: **Medium** — `MODULE_SIG_FORCE=y` goes distro-wide, and only to the unverified machines.

`security` and `secureboot` are both unconditionally in `AVOCADO_DEFAULT_DISTRO_FEATURES` (`conf/distro/avocado.conf:8`), so this bbappend puts `CONFIG_MODULE_SIG_FORCE=y` and the dm-verity fragment on every linux-yocto machine by default — while `encrypted-var`, the smaller change, was deliberately made opt-in in this same PR.

Two things worth confirming before this merges: that out-of-tree / feed module RPMs are signed with the same key oe-core hands the kernel, and that the key is stable across kernel rebuilds and A/B kernel updates (an ephemeral per-build `MODULE_SIG_KEY` makes already-installed module packages unloadable). imx93 builds linux-imx and receives none of this, so the verified board is the one board this doesn't touch. Major if either answer is no.

nicksinas (commented):
Reviewed the full branch against a local clone (65 commits, base `imx93-frdm-wrynose-bsp` @ 1d2d44b, head 3901868). No build attempted; findings are from source plus the tree's own wiring (kas configs, DISTRO_FEATURES, packagegroups).

The imx93 fixes are sound and the hardware evidence is real. The pattern in what follows is that most of this PR's own diagnoses apply to the other five machines, none of which got the fix or a boot pass.

**One fact several comments depend on:** `encrypted-var` is defined in `kas/feature/encrypted-var.yml` but no machine config or CI job includes it. Every machine in the repo, imx93 included, builds with encryption off — so nothing in CI exercises any of this, and the imx93 pass required a manual `DISTRO_FEATURES:append`. Worth confirming whether the machine/CI enablement is meant to land here or follow; the answer decides whether several of these are latent or live.

17 inline comments, each tagged **Major** / **Medium** / **Low**. Two more I left off as too small to inline: `gen-sbkeys.sh:42` says the KEK is "signed with the PK private key" when every cert is independently self-signed with `-new -x509` and `PK.key` is never used (conventional for setup-mode enrollment, but the comment says otherwise, and 4a98a62 is already about correcting misleading comments); and it's worth confirming `nativesdk-gptfdisk` / `nativesdk-mtools` / `nativesdk-dosfstools` are all real BBCLASSEXTEND targets on wrynose, since the UEFI provisioning path depends on all three.

meta-avocado/recipes-security/sb-keys/sb-keys.bb
```diff
@@ -0,0 +1,71 @@
...
+        if [ -f "${AVOCADO_SB_KEYS_DIR}/${cert}.der" ]; then
+            install -m 0644 "${AVOCADO_SB_KEYS_DIR}/${cert}.der" \
+                "${D}${datadir}/avocado/sb-keys/${cert}.der"
+        fi
+        # Private .key files are intentionally NOT installed.
+    done
+}
+
+do_deploy() {
+    install -d "${DEPLOYDIR}/sb-keys"
```
nicksinas, line 51: **Medium** — `${DEPLOYDIR}` with no `inherit deploy`, in a recipe nothing builds.

Without `inherit deploy`, `DEPLOYDIR` is unset and this expands to `install -d /sb-keys` at the host filesystem root; `addtask deploy after do_install before do_build` also leaves the output with no sstate handling.

It hasn't surfaced because nothing references `sb-keys` or `key-store` — no RDEPENDS, no IMAGE_INSTALL, no packagegroup — so neither recipe was built by the 7975-task run. Relatedly: `efitools-native` is in DEPENDS but never used, and no `.esl`/`.auth` is produced, so nothing can actually enroll the generated chain. Either wire these in and make them build, or hold them until the enrollment path exists.

meta-avocado/conf/distro/include/avocado-security.inc
```diff
@@ -0,0 +1,8 @@
+# Per-developer key paths are excluded from sstate hash inputs so two build
+# hosts with identical key material but different absolute paths produce the
+# same sstate hash.  Without this, every developer who places keys under a
+# different TOPDIR would get a cache miss even though the resulting binaries
+# are identical.
+AVOCADO_SB_KEYS_DIR ?= "${TOPDIR}/avocado-sb-keys"
+
+BB_BASEHASH_IGNORE_VARS:append = " AVOCADO_SB_KEYS_DIR"
```
nicksinas: **Medium** — this makes sstate serve another developer's certificates.

The comment's premise is "identical key material, different absolute paths", but the certs are generated inside `do_compile` and never enter the task hash, so excluding the path also collapses *different* key material to one hash. With a shared sstate mirror, a build pointing at key dir A gets the package built from key dir B's certs — silently, with the wrong keys.

For a secure-boot key path this needs to be hash-*visible*, not hash-ignored. If path-independence is the real goal, hash the key material (or a digest of it) rather than dropping the variable.

meta-avocado-qemu/recipes-core/systemd/files/avocado-boot.conf
```diff
@@ -1,4 +1,4 @@
-title avocado-linux
+title Avocado OS
 linux /bzImage
-initrd /avocado-image-initramfs-qemux86-64.cpio.zst
-options root=gpt-auto console=ttyS0
+initrd /initramfs.cpio.zst
+options root=PARTLABEL=rootfs-a ro console=ttyS0,115200n8 console=tty0
```
nicksinas, line 8: **Medium** — hardcoding `root=PARTLABEL=rootfs-a` breaks the B slot.

The manifest this PR rewrites declares `sdboot-ab` with `rootfs-a`/`rootfs-b` slot partitions, and there is one loader entry copied into both ESPs. An A→B update therefore boots B's kernel against A's rootfs — the update appears to succeed while running the old rootfs.

Contained to qemux86-64, but that machine exists to stand in for x86 hardware (per the conf comment), so A/B never gets validated. Keep `root=gpt-auto`, or emit a per-slot entry.

scripts/feed-validation-lib.sh
```diff
@@ -324,19 +324,69 @@ fvl_assert_in_target() {
   return 0
 }
 
-# fvl_boot_verify_case <machine> <projdir> <pkg_csv> <lib_csv>
-# One full image cycle: scaffold project -> install packages into the app ext ->
+# fvl_assert_luks_var
+# Checks that /var is mounted from an active LUKS2 dm-crypt container.
+# Requires cryptsetup in the initramfs and a var partition with a LUKS2 header.
+fvl_assert_luks_var() {
+  fvl_qga --run "cryptsetup status var >/dev/null 2>&1" \
```
nicksinas, line 331: **Medium** — these assertions call `cryptsetup` in a rootfs that doesn't ship it.

`fvl_assert_luks_var` / `_tpm` run `cryptsetup status`, `isLuks` and `luksDump` in the booted target, but `packagegroup-avocado-rootfs.bb:63` adds only `cryptsetup-var-udev` — `cryptsetup` itself is in `packagegroup-avocado-initramfs` only. So the assertions fail on a correctly working encrypted /var.

Add `cryptsetup` to the same conditional that pulls in `cryptsetup-var-udev`, or assert via `dmsetup` / `findmnt` / `lsblk`, which the PR description's own board evidence uses.

scripts/feed-validation-cases
```diff
@@ -2,32 +2,40 @@
...
 # (e.g. zeromq, not libzmq - debian renaming is not active here).
 #
-# name              | packages               | expected-libs | boot
-zeromq              | zeromq                 | libzmq.so.5   | yes
+# name              | packages               | expected-libs | boot | assertions
+zeromq              | zeromq                 | libzmq.so.5   | yes  |
+# Boot-only cases: no feed package to install, the assertion set is the test.
+luks-var            |                        |               | yes  | luks-var
+luks-var-tpm        |                        |               | yes  | luks-var-tpm
+dm-verity           |                        |               | yes  | dm-verity
```
nicksinas: **Medium** — three new heavy boot cases that are guaranteed red.

All three need `encrypted-var`, which no machine config enables. `dm-verity` is worse: `vroot`, `veritysetup` and any root-hash plumbing appear nowhere in the tree (I grepped), and the new boot entry uses `root=PARTLABEL=rootfs-a`, so `fvl_assert_dm_verity` can never pass — `dm-verity.cfg` is kernel config for a runtime that doesn't exist yet.

Each `boot: yes` case costs a full image cycle, so this is three heavy failures per run. Gate them on the feature, or land them in the commit that adds the functionality (the same rule the comment block above already applies to the camera/cloud-aws cases).

meta-avocado/recipes-core/cryptsetup-var/files/cryptsetup-var.sh
```diff
@@ -0,0 +1,165 @@
...
+}
+
+# Grow the LUKS container to fill the partition (e.g. after avocado-grow-var).
+# Only the dm-crypt mapping is resized here; the btrfs filesystem is grown at
+# mount time via the x-systemd.growfs option on /var (btrfs resize is an online,
+# mounted-fs-only ioctl and cannot run against the not-yet-mounted mapper in the
+# initrd).
+maybe_resize() {
+    PARTITION_SECTORS=$(blockdev --getsz "$VAR_DEV")
+    DATA_OFFSET=$(dmsetup table "$MAP_NAME" 2>/dev/null | awk '{print $8}')
```
nicksinas, line 165: **Medium** — `maybe_resize` can fail the unit after a successful unlock.

If `dmsetup table` fails or `dmsetup` is absent, `DATA_OFFSET` is empty and line 133 becomes `EXPECTED_DM=$(( PARTITION_SECTORS - ))` — a shell syntax error, fatal under `set -e`. /var is open, formatted and ready at that point, and the unit still fails into emergency mode at the last step.

`dmsetup`, `blockdev` and `blkid` are asserted present by the comment at `cryptsetup-var.bb:20` but not by RDEPENDS. Guard the empty case (skip the resize rather than dying) and add the deps.

meta-avocado/recipes-core/cryptsetup-var/files/var-key.sh
```diff
@@ -0,0 +1,48 @@
...
+#   CAAM present      -> caam-keygen black-key blob (i.MX8M Plus)
+#   ELE present       -> ELE/SMW key-derive (i.MX91/93 without fTPM)
+#   fallback          -> this Argon2id path (no hardware key store)
+# The cryptsetup-var.sh caller depends only on stdout - 64 raw bytes.
+set -eu
+
+SECRET_FILE="/var/private/var-key-secret"
+HW_ID_FILE="/sys/firmware/devicetree/base/serial-number"
+
+if [ ! -f "$SECRET_FILE" ]; then
```
nicksinas, line 18: **Medium** — the known-nonfunctional provider stays the default fallback.

The PR description establishes that nothing in the tree creates `/var/private/var-key-secret` and that it lives on the volume being unlocked, so this path has never once succeeded. Leaving it as the fallback means the next `encrypted-var` machine added without its own provider gets an unbootable /var, discoverable only at first boot on hardware — which is exactly how imx93 got here.

Make it fail at build time (or derive from something universally available) rather than at first boot.

meta-avocado-qemu/conf/machine/avocado-qemux86-64.conf
```diff
@@ -45,14 +44,38 @@ XSERVER = "xserver-xorg \
...
+# "cryptsetup open --token-only" needs to reopen the TPM2-sealed /var on reboot.
+# (systemd-cryptenroll can enroll the token without it, but the reopen handler is
+# a separate libcryptsetup plugin; without it the reopen falls back to Argon2id.)
+PACKAGECONFIG:append:pn-systemd = " tpm2 cryptsetup-plugins"
+
+# barebox is a feed-compatible bootloader recipe, but with MACHINE_FEATURES
+# 'efi' (above) its build emits barebox.efi, which fails objcopy in this
+# toolchain. qemux86-64 boots via systemd-boot + OVMF and never uses barebox, so
+# exclude it from the build. (Parsed globally, but only this machine sets efi.)
+BBMASK += "recipes-bsp/barebox/"
```
nicksinas: **Low** — `BBMASK` from a machine conf is global.

The comment acknowledges it is "parsed globally" — that's the problem rather than the mitigation: a multiconfig build that also targets a barebox machine loses the recipe entirely. Prefer fixing or gating the barebox `efi` build path (or masking from the machine's kas `local_conf_header` where the scope is explicit) over masking another layer's recipe distro-wide.

meta-avocado-qemu/recipes-security/optee-ftpm-init/files/optee-ftpm-setup.sh
```diff
@@ -0,0 +1,57 @@
...
+# TEE_FS_PARENT_PATH=/var/lib/tee, so that exact path must be the persistent
+# mount or the fTPM's NV lands on the initramfs tmpfs and is lost on reboot.
+TEE_STORE=/var/lib/tee
+mkdir -p "$TEE_STORE"
+if ! mount "$TEE_DEV" "$TEE_STORE" 2>/dev/null; then
+    echo "optee-ftpm: first boot - formatting TEE store on $TEE_DEV"
+    # If we cannot format and mount the persistent store, do not fall through:
+    # tee-supplicant would keep the fTPM NV on the initramfs tmpfs, silently
+    # non-persistent. Skip fTPM bring-up instead so /var takes its Argon2id path.
+    if ! mkfs.btrfs -M -f -L teestore "$TEE_DEV" || ! mount "$TEE_DEV" "$TEE_STORE"; then
```
nicksinas: **Low** — `mkfs.btrfs -f` force-formats the `recovery` partition on any mount failure.

`cryptsetup-var.sh` deliberately probes with `blkid -p` and omits `-f` "so an unexpected foreign signature fails loudly rather than being clobbered". This does the opposite, on the partition named recovery, for any mount failure rather than only "no filesystem" — a dirty btrfs needing recovery gets wiped instead.

Harmless today (qemuarm64 declares recovery as `"image": ""`), destructive the moment recovery content exists. Probe first and drop `-f`.

meta-avocado/recipes-core/cryptsetup-var/files/var-key.sh
```diff
@@ -0,0 +1,48 @@
...
+# per-device, non-secret. Parameters: m=65536 (64 MiB), t=3, p=1.
+SALT=$(printf '%s' "$HW_ID" | openssl dgst -sha256 | sed 's/.*= *//' | cut -c1-32)
+
+# openssl kdf -binary emits the 64 raw key bytes directly (no hex round-trip, so
+# no xxd, which the minimal initramfs does not ship). stderr is left attached so
+# a KDF failure surfaces in the journal and trips set -e in the caller rather
+# than yielding a silent empty key.
+openssl kdf -binary \
+    -keylen 64 \
+    -kdfopt pass:"$(printf '%s:%s' "$SECRET" "$HW_ID")" \
```
nicksinas, line 43: **Low** — the secret goes through argv.

`-kdfopt pass:"$SECRET:$HW_ID"` is visible in `/proc/<pid>/cmdline` for the life of the process. Moot while this provider can't run at all, but worth fixing before it's revived — the other providers pass a public hardware id, so this is the one place it matters.

meta-avocado-qemu/recipes-security/optee-ftpm-init/files/optee-ftpm-setup.sh
```diff
@@ -0,0 +1,57 @@
...
+# tee-supplicant services the fTPM's REE-FS storage (and the RPC in general).
+tee-supplicant -d
+
+# Load the fTPM driver now that storage is available; it opens a session to the
+# embedded early TA and registers /dev/tpm0.
+modprobe tpm_ftpm_tee || true
+
+# Give the chip time to appear so the enroll finds it.
+i=0
+while [ ! -e /dev/tpm0 ] && [ "$i" -lt 100 ]; do i=$((i + 1)); sleep 0.1; done
```
nicksinas: **Low** — `sleep 0.1` needs fractional-second support.

Fractional `sleep` is a coreutils/GNU extension. If the initramfs `sleep` is busybox without it, each call errors out instantly, the loop spins its 100 iterations in no time, and the script reports `/dev/tpm0` missing on a board where it would have appeared. `cryptsetup-var.sh:60` uses whole seconds for the same kind of wait — worth matching, or confirming which `sleep` ends up in the initramfs.

meta-avocado-x86-64/stone/stone-intel-x86-64-v3.json
jetm: Confirmed, and the argument in 4bdf9b4 (restoring imx93's seed) applies machine-independently exactly as you said. Traced it: `kas/feature/encrypted-var.yml` exists but no kas machine config anywhere references it, so `cryptsetup-var.sh` - the only thing in the tree that formats /var - never runs on a default build of any of these five. Restored the var image seed on `stone-intel-x86-64-v2/v3/v4.json`, `stone-qemuarm64.json`, `stone-qemux86-64.json`, and `qemuarm64/rootdisk.conf`, matching imx93-frdm exactly (commit `e9bdc94`).

meta-avocado-x86-64/conf/machine/include/avocado-x86-64.inc
jetm: Right - `avocado-x86-64.inc` had no `AVOCADO_VAR_PART_DEV` line at all, so it fell through to base-files' own `PARTUUID=4d21b016-...` default, which never resolves to `/dev/mapper/var`. Added the same `bb.utils.contains('DISTRO_FEATURES', 'encrypted-var', ...)` line qemuarm64/qemux86-64/imx93-frdm already carry, so x86-64 gets `cryptsetup-var-udev` and the mapper path the moment encrypted-var is turned on for it (commit `88e2506`).

meta-avocado/recipes-kernel/linux/files/dm-crypt.cfg
jetm: Confirmed - `meta-avocado/recipes-kernel/linux/files/dm-crypt.cfg` only set `CONFIG_DM_CRYPT=y`, and it reaches every linux-yocto machine that turns on encrypted-var, which is strictly more exposure than the imx93-specific copy. Added `CONFIG_BLK_DEV_DM=y` and `CONFIG_TRUSTED_KEYS=y` there too (commit `7abbe09`). Left the imx93 copy's duplicate promotions alone for now - it feeds a different, unconditionally-included recipe (linux-imx) rather than this encrypted-var-gated linux-yocto fragment, and consolidating the two means touching that wiring, which is out of scope for this fix.

meta-avocado-x86-64/recipes-core/cryptsetup-var/files/var-key.sh
jetm: Agreed, and worse than described - checked, and the fallback isn't even always reached: whitebox/OEM boards commonly populate DMI `product_uuid`/`product_serial` with a literal placeholder ("Default string", "To Be Filled By O.E.M.", the null UUID, etc.) rather than leaving it empty, so "non-empty" alone doesn't mean unique. Made the x86-64 `var-key.sh` refuse exactly like imx93's does: exit 1 on an empty read or a known placeholder, no constant fallback (commit `892ae98`). Independently re-tested this myself against fake sysfs paths: empty and all four placeholder strings correctly exit 1, a real-looking UUID correctly proceeds.

meta-avocado/recipes-kernel/linux/files/dm-crypt.cfg
jetm: Confirmed the conflict and the leftover-from-dm-verity theory - `CONFIG_HW_RANDOM_TPM=y` in `dm-crypt.cfg` has nothing to do with dm-crypt and directly fights the `# CONFIG_HW_RANDOM_TPM is not set` that `tpm.cfg`/`ftpm.cfg` set for the documented tpm_mutex-hang reason. Dropped the line, same commit as the parent-symbol fix above since it's the same 5-line file (`7abbe09`).

meta-avocado/recipes-kernel/linux/linux-yocto_%.bbappend
jetm: Confirmed the mechanics: `security` and `secureboot` are both unconditionally in `AVOCADO_DEFAULT_DISTRO_FEATURES`, and no `MODULE_SIG_KEY` is set anywhere in the tree, so every linux-yocto machine gets `MODULE_SIG_FORCE=y` against the kernel's own ephemeral per-build signing key. I couldn't independently confirm from the tree whether any out-of-tree module feed RPM path exists that would actually hit the unloadable-module failure mode you describe, or what the intended key-stability story is. This is a product/security-posture call rather than a code fix, so leaving it open pending that decision - will follow up here once it's settled.

meta-avocado/recipes-security/sb-keys/sb-keys.bb
jetm: Confirmed - `sb-keys.bb` calls `addtask deploy` and writes to `${DEPLOYDIR}` without `inherit deploy`, latent only because nothing currently references `sb-keys`/`key-store`. Added the missing inherit, matching `avocado-image-var.bb`'s use of the same class (commit `f39ee1e`). Left the RDEPENDS/enrollment-wiring question alone - the "or hold them" branch is the right default while `efitools-native` stays unused and no `.esl`/`.auth` gets produced, so there's nothing yet for an image to consume.

meta-avocado/conf/distro/include/avocado-security.inc
jetm: Agreed, and traced the mechanism: the certs are generated inside `sb-keys`' `do_compile` and never otherwise enter the task hash, so excluding `AVOCADO_SB_KEYS_DIR` from `BB_BASEHASH_IGNORE_VARS` collapses genuinely different key material to the same signature, not just different paths with identical material. Removed the ignore-vars entry (commit `6c471b4`) - the cost is a cache miss per differing key dir, which is the safe failure mode for key material. Content-hashing the actual certs would avoid even that cost but is a larger change than reverting an unsafe shortcut warrants right now.

meta-avocado-qemu/recipes-core/systemd/files/avocado-boot.conf
jetm: Confirmed the regression, and checked both suggested fixes before picking either: `rootfs-a` and `rootfs-b` in `stone-qemux86-64.json` share the same GPT partition type GUID (`8304`), so `root=gpt-auto` cannot disambiguate which slot booted without a distinct type GUID per slot or a real per-slot loader entry. Not confident which approach is correct without boot-testing it, so leaving this open - this is a slot-selection design question, not something to guess at in a fix pass. Will follow up once that's settled.

scripts/feed-validation-lib.sh
jetm: Confirmed - `packagegroup-avocado-rootfs.bb` only added `cryptsetup-var-udev` (just the udev rule, no binary); the `cryptsetup` CLI these assertions call lives only in the initramfs packagegroup. Took the first option rather than rewriting the assertions to dmsetup/findmnt/lsblk: added `cryptsetup` to the same `AVOCADO_VAR_PART_DEV`-gated RDEPENDS as `cryptsetup-var-udev`, so it only ships on a build where encrypted-var is actually wired to the mapper (commit `4b39868`).

scripts/feed-validation-cases
jetm: Confirmed both halves: `MACHINE=qemuarm64` (this script's target) has encrypted-var off in every kas config, and grepped for `vroot`/`veritysetup` - dm-verity has no implementation anywhere in the tree beyond the new kernel config fragment. Removed the three case lines rather than the assertion functions, matching the file's own stated rule for the cameras/cloud-aws groups: add the case in the commit that adds what it tests (commit `4b39868`).

meta-avocado/recipes-core/cryptsetup-var/files/cryptsetup-var.sh
jetm: Confirmed, and not just a theoretical edge case - `dmsetup`'s package (`device-mapper`) genuinely isn't RDEPENDS'd anywhere in this recipe or the initramfs packagegroup, so the comment claiming it was "already there" was wrong. Added `device-mapper` to RDEPENDS and guarded the empty-`DATA_OFFSET` case to skip the resize with a logged warning instead of hitting the malformed arithmetic expression (commit `24a85a7`).

meta-avocado/recipes-core/cryptsetup-var/files/var-key.sh
jetm: Agreed this is a real gap, but it isn't live on this pushed head - every machine that currently turns on encrypted-var (qemu, x86-64, imx93) already ships its own `var-key.sh` override via FILESEXTRAPATHS, so the shared placeholder's failure mode only bites a future machine that forgets to add one. Left it as-is rather than guessing at a build-time detection mechanism for "does this machine have an override" - that's real design work, not a quick fix, and the file's own header already documents a full phase-2 replacement of this provider.

meta-avocado-qemu/conf/machine/avocado-qemux86-64.conf
jetm: Confirmed, and took the second option: moved `BBMASK += "recipes-bsp/barebox/"` from `avocado-qemux86-64.conf` into `kas/machine/qemux86-64.yml`'s `local_conf_header`, matching the existing precedent in `kas/vendor/raspberrypi.yml` and `kas/vendor/nxp-frdm-imx93.yml` (commit `5743285`). Verified with a real `bitbake -p` parse of this machine post-fix: 3549 recipes parsed, 2 masked, 0 errors - the mask still applies to a normal qemux86-64 build, it just no longer rides along with the .conf into a build that also targets a barebox machine.

meta-avocado-qemu/recipes-security/optee-ftpm-init/files/optee-ftpm-setup.sh
jetm: Confirmed the inconsistency with `cryptsetup-var.sh`'s own convention. Mirrored it exactly: probe with `blkid -p` first, and only format (without `-f`) when nothing is there; any other mount failure now exits and falls back to Argon2id instead of clobbering (commit `8bc3913`).

meta-avocado/recipes-core/cryptsetup-var/files/var-key.sh
jetm: Agreed this is real, but tested rather than guessed at a fix: `openssl kdf -kdfopt pass:env:VAR` and `pass:stdin` are both taken as literal password strings, not indirection - `-kdfopt` doesn't support the pass-phrase mini-language other openssl subcommands do. A correct fix needs a different mechanism that isn't verified yet, and this provider can't run at all today (nothing populates `/var/private/var-key-secret`), so left alone rather than land an unverified KDF invocation change in dead code - worth resolving properly whenever this file's documented phase-2 rewrite happens.

meta-avocado-qemu/recipes-security/optee-ftpm-init/files/optee-ftpm-setup.sh
jetm: Confirmed - busybox's `sleep` needs `FEATURE_FANCY_SLEEP` for a fractional argument, not guaranteed on this minimal initramfs, and `cryptsetup-var.sh`'s own TPM wait already uses whole seconds for this exact reason. Changed to `sleep 1` over 10 iterations, same total wait budget as the old 100 x 0.1s (commit `8bc3913`).

meta-avocado/recipes-kernel/linux/linux-yocto_%.bbappend
jetm: Good catch. Fixed in d9394cb.

The modsign.cfg fragment is now gated on a new `module-signing` distro feature instead of `security`, and that feature is deliberately absent from `AVOCADO_DEFAULT_DISTRO_FEATURES`, so it defaults off everywhere. A board turns it on through `kas/feature/module-signing.yml`, the same shape as `kas/feature/encrypted-var.yml`. The `secureboot` (dm-verity.cfg) and `encrypted-var` (dm-crypt.cfg) lines are untouched. A new name rather than a narrower `security` was needed because `security` and `secureboot` both stay unconditionally on for their other purposes.

On the two things worth confirming: both answers are no today, which is why this goes opt-in rather than staying on with a caveat. Nothing in the tree sets `MODULE_SIG_KEY` and nothing signs the package feeds, so oe-core generates a fresh signing key per kernel build. Already-installed module RPMs would stop loading after the next kernel build, and an A/B kernel update would break every out-of-tree module on the device. That is the story that has to exist before this can be turned on by default, and it is not in this PR.

Grepping the tree for other consumers of the `security` feature turned up exactly one: this bbappend line. No machine .conf, kas yml, packagegroup or class reads it, so nothing else was relying on `security` to imply module signing and no other gate changed. `secureboot` has two other references, `linux-yocto_%.bbappend:3` for dm-verity and a `qemuarm64-secureboot` MACHINEOVERRIDE in `meta-avocado-qemu/conf/machine/avocado-qemuarm64.conf:8`; neither is about module signing, and no machine config or doc treats `secureboot` as implying it, so no board's behavior changes beyond losing the forced-signing kernel it could not have satisfied anyway.

meta-avocado-qemu/recipes-core/systemd/files/avocado-boot.conf
jetm: Good catch, and a per-slot entry would not have closed it either. stone's OTA
schema gives each `update.os_artifacts` entry one `image_key` shared across
every slot in `slot_partitions`, so a single build of the `boot` image lands
byte-identical on whichever ESP an update targets. Whatever `root=` says, both
ESPs end up holding the same bytes.

So the decision moves to boot time. A new generator,
`meta-avocado/recipes-core/systemd/avocado-slot-root-generator`, reads
`LoaderDevicePartUUID` in the initramfs, maps the booted ESP to a slot through
a table the recipe generates from the same stone manifest that pins those ESP
partition UUIDs, and writes a `What=` drop-in for `sysroot.mount` into the
early generator directory. `systemd-fstab-generator` still builds the unit from
the command line, so its fsck and `initrd-root-device` wiring survives and only
the device changes. It comes in through `packagegroup-avocado-initramfs` gated
on `AVOCADO_BOOTLOADER = "uefi"`, which covers qemux86-64 and the three intel
SKUs. U-Boot targets read their slot from a U-Boot variable and do not get it.

`root=PARTLABEL=rootfs-a` stays, now as the degrade path rather than the
answer. Missing EFI variable, truncated, non-UUID, or a UUID absent from the
table: the generator logs to `/dev/kmsg` and returns without writing anything,
and the boot goes exactly as it does today. There is a comment above the line
now saying so, because otherwise the next reader files this same review.

One deliberate omission. The rootfs side still resolves by PARTLABEL rather
than a pinned `partition_uuid` on `rootfs-a`/`rootfs-b`. A pinned UUID is
identical on every device of a SKU, same as the label, so it buys no
uniqueness while adding a second table to keep in step with the recipe. Worth
pinning for other reasons if there are some.

Now the part that decides whether this thread closes: none of it is
boot-tested. A qemux86-64 image off this branch is a from-scratch wrynose build
with no usable sstate, and I did not attempt it. What I did run was the
installed script against a faked efivarfs blob, using the real encoding of a
4-byte attribute prefix and a NUL-terminated UTF-16LE value, in both the
uppercase form systemd-boot writes and the lowercase form other Boot Loader
Interface implementations write. Slot A resolved to `rootfs-a` and slot B to
`rootfs-b`, and each degrade path above wrote no drop-in and exited 0.
Separately I lifted the recipe's `do_write_slot_map` out of the `.bb` and ran
it against all four real `sdboot-efi` manifests: every emitted table matches
the `boot-a`/`boot-b` `partition_uuid` values the same manifest pins, and it
fails the build on a `uboot-env` manifest or one with no rootfs slot
partitions. The drop-in override I checked against a running systemd instead of
taking it from the man page, confirming a drop-in in one unit-load-path
directory merges over a fragment found in another.

So the logic is exercised and the boot is not. Slot A and slot B both need a
real boot before this is trusted on hardware, and I would leave this thread
open until someone has done that rather than close it on a harness.

mobileoverlord (approved):
Reviewed in full, including the cryptsetup-var unlock path, the key providers, the capabilities class, the slot-root generator, and the qemu UEFI conversion. The property that matters most holds everywhere I checked: every shared-code change is gated so an untouched board builds identically — AVOCADO_BOOTLOADER defaults to uboot, encrypted-var is kas-feature opt-in, the capabilities check only fires when a security feature is actually requested, and the initramfs/rootfs packagegroup additions are all feature-conditional. The only default-on behavior change outside the imx93/qemu machines is x-systemd.growfs on /var, which is right. The i.MX key provider refusing to derive a non-unique key (rather than the qemu constant fallback) is exactly the correct asymmetry, and the phase-1 SoC-UID caveat is honestly documented with the ELE phase-2 path named. Hardware evidence is artifact-level and covers both first-boot format and reopen.

Housekeeping done on the branch: folded #298 (sb-keys efitools DEPENDS, which was a guaranteed resolution failure) and #301 (bootloader-only fwup task) into the stack, and rebased onto the rebased #259 — 97 commits replayed clean.

Merge order: #259 first, then this retargets to wrynose and follows.

PR #272, opened 2026-08-12
Secure boot and LUKS2-encrypted /var, unified on scarthgap

jetm:
## Problem

The secure-boot and encrypted-`/var` work has been spread across four PRs on two
branches. On scarthgap it was a two-PR stack (#232 with #236 on top of it), so
neither PR showed the whole change and the base PR had drifted from the one
built on it.

## Solution

One branch per release branch. This is the scarthgap one: the full history from
both scarthgap PRs, rebased onto current scarthgap. Commits keep their original
authorship.

The wrynose counterpart is #271 and is where hardware validation happens -
scarthgap's 6.6.36 SPL stalls before `Normal Boot` on Rev1.2 FRDM silicon, so
this branch cannot be booted on the board it targets. It is kept in step so
scarthgap does not silently fall behind on a security feature.

## Key changes

- Set `AVOCADO_VAR_PART_DEV` on imx93. It was the only `encrypted-var` target
  missing the line, so fstab mounted `/var` from the raw partition while
  cryptsetup-var created and opened the LUKS container underneath it
- Promote `BLK_DEV_DM` and `TRUSTED_KEYS` to `=y`. kconfig caps a tristate at
  its weakest parent, so the fragment's `CONFIG_DM_CRYPT=y` was silently
  demoted to `=m` and the initrd unlock had no dm-crypt to call
- Restore the var seed to the imx93 stone manifest, fixing the `var.img`
  provisioning failure reported on 2026-08-04
- Reject a `block_size` the qemux86-64 provisioning script cannot honour,
  rather than reading it from the manifest and laying the image out in 512-byte
  units anyway

## Reviewer notes

Draft. No build has been run against this exact tree.

The sb-keys `UNPACKDIR` fix carried on the wrynose branch is deliberately NOT
here: nothing in this tree uses `UNPACKDIR`, so `WORKDIR` is correct on
scarthgap and porting it would break `do_compile`. This is the one place the
two branches intentionally differ.

`var-key.sh` still reads its secret from the volume it is unlocking, on both
branches.

Supersedes #232 and #236, which are closed.


jetm:
Doing more testing before it's Ready for Review...

jetm:
Closing as superseded. The work continued on the wrynose line and this scarthgap version has diverged behind it; 90 of its 142 files already exist upstream, and the pieces that do not have been relocated to different layers since. Leaving the branch in place rather than deleting it.

PR #46, opened 2025-06-03
E2E Testing for Qemux86-64

mobileoverlord:
This PR does the following:
* Change the `qemux86-64-secureboot` machine name to `qemux86-64`. This was the only system where sometimes there were divergences in naming convention. 
* Brings in WPEWebKit and builds for targets that have OpenGL enabled in `DISTRO_FEATURES`
* Updates the `sdk-test` containers to use `podman-compose`. podman fixes permissions issues with bind mounted directories and enables mapping the root user in the container with the system user outside the container. 
* Fix issues with the `avocado-repo` scripts installing duplicative packages in extension os trees
* Convert qemu to build final images using genimage. Using Genimage allows reuse of partition mapping and image creation outside yocto.
* Fix issues with swtpm throwing a segmentation fault by working around an issue with linking libpcre to sources built wihin yocto. This needs to be investigated further, but is low priority.
* adds `nativesdk-avocado-pkg-images` which packages the entire contents of the deploy dir for use in the sdk with the same directory and relative file structure as within yocto. This also updates `avcoado-repo images` to install this package to `_avocado/images` where deploy dir files are found in `_avocado/images/deploy`.
* Updates the nativesdk toolchain to package tools necessary for genimage and mksquashfs to function out of the box

PR #274, opened 2026-08-12
rpi/sdk-target: investigate missing boot artifacts at stone bundle

jetm:
## Status

Draft, pending validation. The problem statement this change was written against came from reading the stone manifest and the recipe's dependencies, not from a build. A later check against an actual Raspberry Pi 5 build contradicts it. Do not merge until the real failure is located.

## What this change assumed

`avocado build` for a Raspberry Pi target dies at finalize:

```
[ERROR] File 'u-boot.bin' not found in any input directory for FAT image
```

The assumption was that `stone-raspberrypi<n>.json` names boot artifacts the runtime build never produces - `u-boot.bin` (delivered as `kernel_<soc>.img`) and the whole `bootfiles/` set of `start*.elf`, `fixup*.dat`, `bootcode.bin`, `config.txt` and `cmdline.txt` - and that the only input directory stone receives is the runtime directory, which holds just rootfs, kernel, initramfs and var.

## What a build actually shows

Listing the `avocado-img-bootfiles` RPM from a Raspberry Pi 5 build tree:

```
/u-boot.bin
/bootfiles/bootcode.bin
/bootfiles/config.txt
/bootfiles/cmdline.txt
/bootfiles/start.elf
/bootfiles/fixup.dat
    ... 20 entries under /bootfiles in total
```

`avocado-runtime.bb:18` RDEPENDs on `avocado-img-bootfiles`, and the base recipe packages the deploy directory wholesale via `FILES:${PN} = "/*"`, skipping only rootfs, initramfs and var. Those artifacts therefore already reach the runtime directory that `stone bundle` receives as an input, so the premise that nothing produces them does not hold.

The dependencies this recipe already declared are not unconsumed either. `meta-avocado-raspberrypi/recipes-avocado/distro/avocado-stone.bbappend` declares `u-boot:do_deploy` and `rpi-bootfiles:do_deploy`; `avocado-img-bootfiles` declares `do_compile[depends] += "avocado-stone:do_deploy"`; and `do_collect_artifacts` is registered `after do_compile`. The deploys are transitively ordered ahead of the collection step, which is consistent with the artifacts being present in the RPM.

## Why this is on hold rather than closed

The reported error is real and still unexplained. If the files already sit in the runtime directory, the lookup fails somewhere downstream - on the CLI side, or against a stale runtime - rather than in this recipe. Closing outright would lose the report; merging would add a copy step that hides the cause.

The change as written also depends on a consumer that does not exist. It stages into `${SDKPATHNATIVE}/stone`, and nothing reads that path. An earlier reviewer note cited avocado-cli#183 as the consumer half, but that PR covers device-tree overlay provisioning; no avocado-cli PR, open or merged, adds this directory as a `stone -i` input.

## Validation still needed

Reproduce `File 'u-boot.bin' not found in any input directory for FAT image` with a Raspberry Pi build in hand and identify where the lookup actually fails. The RPM listing above came from a wrynose-generation build (u-boot 2025.10); this PR targets scarthgap, and although the packaging recipe is identical on both branches, scarthgap has not been reproduced directly.

### DIFF
--- a/meta-avocado-raspberrypi/recipes-avocado/sdk/avocado-sdk-target.bbappend
+++ b/meta-avocado-raspberrypi/recipes-avocado/sdk/avocado-sdk-target.bbappend
@@ -4,6 +4,28 @@ FILESEXTRAPATHS:prepend := "${@':'.join(['%s/stone' % layer for layer in d.getVa
 do_compile[depends] += "u-boot:do_deploy"
 do_compile[depends] += "rpi-bootfiles:do_deploy"
 
+# Stage the boot artifacts stone-raspberrypi<n>.json names but this build does
+# not produce: u-boot.bin (-> kernel_<soc>.img) and the whole bootfiles/ set
+# (start*.elf, fixup*.dat, bootcode.bin, config.txt, cmdline.txt).
+#
+# The two dependencies above already force those recipes to deploy, but nothing
+# copied the result anywhere stone could see it, so the dependency was declared
+# and never used. The runtime input dir only holds what the runtime build makes
+# (rootfs, kernel, initramfs, var), so `avocado build` died at finalize with
+# "File 'u-boot.bin' not found in any input directory for FAT image".
+#
+# They go beside the manifest that names them, in the SDK's stone dir, which the
+# CLI passes to `stone bundle` as an input directory. Reading DEPLOY_DIR_IMAGE
+# here is safe precisely because of the do_deploy dependencies above - without
+# them this would be a race rather than a copy.
+do_install:append() {
+    install -d ${D}${SDKPATHNATIVE}/stone
+    install -m 0644 ${DEPLOY_DIR_IMAGE}/u-boot.bin ${D}${SDKPATHNATIVE}/stone/u-boot.bin
+
+    install -d ${D}${SDKPATHNATIVE}/stone/bootfiles
+    cp -a ${DEPLOY_DIR_IMAGE}/bootfiles/. ${D}${SDKPATHNATIVE}/stone/bootfiles/
+}
+
 RDEPENDS:${PN}:append = " \
   nativesdk-fwup \
   nativesdk-mkfat \


meta-avocado-raspberrypi/recipes-avocado/sdk/avocado-sdk-target.bbappend
```diff
@@ -4,6 +4,28 @@ FILESEXTRAPATHS:prepend := "${@':'.join(['%s/stone' % layer for layer in d.getVa
 do_compile[depends] += "u-boot:do_deploy"
```
mobileoverlord, line 4: stone bundle is called at the end of the build cycle to validate the contents of the stone manifest are available from the `DEPLOY_DIR_IMAGE`, which is the input path given to stone. I would expect that depending on u-boot:deploy to have occurred before the do_compile would result in u-boot having staged u-boot.bin to the input path and it would resolve. Has the output path for this artifact changed in `DEPLOY_DIR_IMAGE` such that the input manifest is using an incorrect relative path?

jetm: No, the paths in the manifest are still correct. I had the wrong stone invocation in mind when I wrote this.

There are two. The bitbake one in `avocado-stone.bb` (`do_stone_validate` / `do_stone_bundle`) runs with `-i ${DEPLOY_DIR_IMAGE}`, and `meta-avocado-raspberrypi/recipes-avocado/distro/avocado-stone.bbappend` already carries the `u-boot:do_deploy` and `rpi-bootfiles:do_deploy` edges, so `u-boot.bin` and `bootfiles/*` resolve there exactly as the manifest names them. The one that failed for me is the SDK-side bundle `avocado build` runs at finalize, which passes `-i $AVOCADO_PREFIX/runtimes/$RUNTIME_NAME` (avocado-cli `src/commands/runtime/build.rs`, the `STONE_INPUT_DIR` fed into `stone bundle`), not `DEPLOY_DIR_IMAGE`.

What I got wrong is assuming that directory holds only what the runtime build produced. `avocado-img-bootfiles` collects the whole deploy dir and packages it at package root, and `avocado-runtime` pulls it in, so dnf lands it straight in the runtime root. I checked the imx93 RPM staged in our local repo and it carries `imx-boot`, `bl31.bin` and every dtb at `./`, which is the same relative layout the manifest expects. `u-boot.bin` and `bootfiles/` should arrive by that path with nothing extra staged, which makes this PR a second delivery mechanism for artifacts that already have one, plus an avocado-cli change to consume it. I would rather close it than land that.

If the failure reproduces on a real pi tree, the place to look is whether the img-bootfiles RPM is actually missing those files, because `meta-avocado-raspberrypi` has no `avocado-img-bootfiles.bbappend` adding the deploy edges the way meta-avocado-qemu and meta-avocado-nvidia do, and `docs/migrations/scarthgap-to-wrynose.md` #46 documents that exact ordering and sstate trap for u-boot artifacts. That would be a small dep fix in the recipe that owns the delivery rather than a copy in the SDK recipe. I do not have a pi build tree to confirm which it is, and the original commit was never build-verified either, so I would rather not guess at a replacement patch here.

jetm:
Holding this rather than merging, because the Problem statement above doesn't survive a check against a real build.

I listed the `avocado-img-bootfiles` RPM from a Raspberry Pi 5 build tree. It already carries `/u-boot.bin` at the package root and a full `/bootfiles/` directory - 20 entries including `bootcode.bin`, `start*.elf`, `fixup*.dat`, `config.txt` and `cmdline.txt`. `avocado-runtime.bb:18` RDEPENDs on that package, and the base recipe packages the deploy directory wholesale via `FILES:${PN} = "/*"`, skipping only rootfs, initramfs and var. The runtime build therefore does produce these files, and they are already delivered into the runtime directory that `stone bundle` receives as an input.

That makes the `do_install:append` here a second copy of files that are already present by another route, staged into `${SDKPATHNATIVE}/stone`. I could not find anything that reads that directory. The reviewer note points at avocado-cli#183 as the consumer half, but that PR is "Provision and build declared device-tree overlays"; no PR in avocado-cli, open or merged, adds this directory as a `stone -i` input. So the pairing this change depends on does not currently exist.

On ordering: the dependencies above are not unconsumed. `meta-avocado-raspberrypi/recipes-avocado/distro/avocado-stone.bbappend` already declares `u-boot:do_deploy` and `rpi-bootfiles:do_deploy`; `avocado-img-bootfiles` declares `do_compile[depends] += "avocado-stone:do_deploy"`; and `do_collect_artifacts` is registered `after do_compile`. The deploys are transitively ordered ahead of the collection step, which is consistent with the artifacts actually being in the RPM.

Two caveats on what I checked. The build tree I inspected is wrynose-generation (u-boot 2025.10) rather than scarthgap, which this PR targets; the packaging recipe is the same on both branches, but that is not a direct reproduction. And none of this explains the reported failure - if the files are in the runtime directory, `File 'u-boot.bin' not found in any input directory for FAT image` has a cause further downstream, most likely on the CLI side or in a stale runtime rather than in this recipe.

Suggest we leave this open and unmerged until someone reproduces the original error with a build in hand and identifies where the lookup actually fails. Merging it now would add a copy step that hides the real cause.

jetm:
Closing as unnecessary: the premise this PR was written against does not hold.

A full rpi5 E2E now passes end to end with **no change to meta-avocado or avocado-cli** — feed built from merged scarthgap `670c668` (carrying #245's `563d462`), avocado-cli #183 `aa2d49a`, and the committed `tooling/dto-e2e-fixture`:

```
>> PASS: hello-overlay.dtbo compiled, delivered to boot, and bundled into a finalized os-bundle.aos.
```

The `File 'u-boot.bin' not found in any input directory for FAT image` failure came from a hand-rolled test project that declared only `sdk.packages` plus `runtimes.<r>.{target,extensions}` and **no `avocado-runtime`**. Without it, the `avocado-img-bootfiles` RPM — which ships `/u-boot.bin` and ~20 `bootfiles/*` — is never installed into the runtime, so stone's FAT builder finds nothing. `avocado-runtime.bb:18` RDEPENDs on it, exactly as this PR's own body noted when a real build contradicted the original assumption.

So `avocado-sdk-target`'s `do_compile[depends] += "u-boot:do_deploy"` is not a dangling dependency in need of a copy step; delivery already happens through the package path. Details in ENG-2134.

PR #276, opened 2026-08-13
ssh: add an ssh-dev feature and keep host keys on /var for scarthgap

jetm:
A freshly flashed board has no way in. `/etc/shadow` ships `root:*:`, and even
once openssh is installed the image cannot serve SSH at all, so getting a shell
has meant a bespoke sequence per board.

Two commits, two independent gaps.

**Host keys were unreachable.** sshd resolves `HostKey` to `${sysconfdir}/ssh`,
which on this image is the read-only rootfs. `sshd_check_keys` cannot create a
key there, so the unit fails, sshd starts keyless, and every client is dropped
at key exchange with `Connection reset by peer` - a symptom that reads like a
network fault rather than a missing file. On top of that, oe-core's
`read_only_rootfs_hook` sees no pre-generated key in `/etc/ssh` and points sshd
at `sshd_config_readonly` with keys on tmpfs, so the device presents a new host
key on every boot. Both halves are needed: clearing `SSHD_OPTS` alone still
writes keys to the read-only path, and repointing `HostKey` alone is ignored
while sshd is running the readonly config.

**No build-time switch.** `ssh-dev` is a kas overlay, so any board can ask:

```
bakar build meta-avocado/kas/machine/<board>.yml:meta-avocado/kas/feature/ssh-dev.yml
```

Deliberately not an `AVOCADO_FEATURE_GROUPS` entry. A feature group populates
the feed - `networking` builds the openssh RPMs and stops there, because the
base image is minimal by design and gains capability through extensions. That
is right for a product and wrong for a bench board, where the point is to have
a shell before any extension has been installed. `ROOTFS_IMAGE_EXTRA_INSTALL`
is the seam the image recipe already leaves for that.

Dev only - it accepts root with an empty password. The feature file says so at
the top.

## Same files as the wrynose version

Byte-identical to #275. The bbappend branches on whether the recipe wrote the
`HostKey` lines at all, so one file is correct on both openssh versions in the
layer: 10.3p1 writes them from `OPENSSH_HOST_KEY_DIR` one per key type enabled
in `PACKAGECONFIG`, so rewriting in place preserves those gates; 9.6p1, which
scarthgap carries, leaves them commented and rewrites only
`sshd_config_readonly`, so sshd runs on its compiled-in defaults and a sed
matches nothing. Either branch alone fails silently on the other version - the
sed is a no-op on 9.6p1, and an unconditional explicit list would resurrect a
key type `PACKAGECONFIG` had turned off on 10.3p1.

## Verification

Verified on an FRDM-IMX93 against the wrynose BSP: sshd reachable over the
network, and the ed25519 host key fingerprint unchanged across a reboot that
changed `boot_id`, so the key is reused rather than regenerated.

Not built or booted on scarthgap. The 9.6p1 branch above was derived by reading
that recipe's `do_install`, not by running it, so the parse and the produced
`sshd_config` are unconfirmed on this branch.

## Added 2026-08-15: a console-only sibling, and a dedupe

Two extra commits, folded in here rather than opened as their own PR since they
touch the same file and the same base.

**`kas/feature/dev-root-login.yml`** is the console-only counterpart of this
fragment - the toggle and nothing else. It had been carried as an untracked file
on the bench through the whole Jetson device-tree-overlay bring-up, because a
stock image sets root's password field to `*` and `/proc/device-tree`, the only
thing that proves an overlay reached the running kernel, is readable only from a
shell on the board.

It is a separate fragment rather than a flag on this one because ssh-dev appends
`openssh-sshd`, `openssh-keygen` and `openssh-sftp-server` to the base rootfs so
oe-core's image features have a config to edit. On a board reached over a serial
cable that is a larger image and a network service for no gain. Reach for
ssh-dev when you need a shell over the network; reach for dev-root-login when
you need a login prompt and nothing else.

**`ssh-dev.yml` now includes it** rather than setting `AVOCADO_DEV_ROOT_LOGIN`
again, the way `containers.yml` pulls its layer through `virtualization.yml`.

That dedupe closes a gap this PR had on its own, which is the part worth a
reviewer's attention. Flipping the toggle re-signs `avocado-users`' `do_install`,
so the PR service can hand back a revision below one already published;
`packagedata` then fails `version went backwards`, and if the feed holds the
older package at a higher revision, dnf resolves to it and the image keeps
`root:*`. The build stays green and the board cannot be logged into - over SSH
or on the console - and the symptom is indistinguishable from the feature not
working. This PR flipped the same toggle and carried none of that handling.

The QA relaxation now inherited from the new fragment covers it. The `PR` bump
that is sometimes additionally needed ships commented out, with a
`devtool-debt:` marker: the right value depends on the reader's feed rather than
on the file, so a fixed one would be correct once and wrong after.

Not resolved through a real `kas dump` - the include is YAML-valid and matches
the `containers.yml` precedent exactly, but it has not been run through kas on
this branch. Worth a reviewer's eye given every ssh-dev build now depends on it
resolving.

The equivalent change is not yet on the wrynose side (#275); it should follow
once the shape is agreed here.

### DIFF
--- a/kas/feature/dev-root-login.yml
+++ b/kas/feature/dev-root-login.yml
@@ -1,0 +1,55 @@
+header:
+  version: 16
+
+# Console root login with an empty password, on any machine.
+#
+#   bakar build meta-avocado/kas/machine/<board>.yml:meta-avocado/kas/feature/dev-root-login.yml
+#
+# DEV ONLY. Never compose it into anything that ships.
+#
+# A stock image sets root's password field to `*`, which no password can match,
+# so the board boots to a console prompt that cannot be satisfied. That is fine
+# for a product and blocking on a bench: /proc/device-tree, the only thing that
+# proves a device-tree overlay reached the running kernel, can only be read from
+# a shell on the board.
+#
+# Prefer this over kas/feature/ssh-dev.yml when the board is reachable over a
+# serial console. ssh-dev also sets AVOCADO_DEV_ROOT_LOGIN, but it appends
+# openssh-sshd, openssh-keygen and openssh-sftp-server to the base rootfs to
+# give those image features a config to edit. On a UART-attached board that is
+# a larger image and a network service for no gain. Reach for ssh-dev when you
+# need scp or a shell over the network; reach for this when you need a login
+# prompt and nothing else.
+local_conf_header:
+  feature/dev-root-login: |
+    # avocado-users_1.0.bb rewrites /etc/shadow's `root:*:` to `root::` only
+    # when this is 1.
+    AVOCADO_DEV_ROOT_LOGIN = "1"
+
+    # Flipping the toggle re-signs do_install, so the PR service can hand back
+    # a revision lower than one already published, and packagedata fails
+    # "version went backwards". That check exists to keep non-upgradable
+    # packages out of real feeds; a bench feed is a scratch repo rebuilt at
+    # will, so demote it to a warning rather than hand-bumping PR every
+    # iteration.
+    ERROR_QA:remove = "version-going-backwards"
+    WARN_QA:append = " version-going-backwards"
+
+    # Demoting the check is not always sufficient. If your feed ALREADY holds
+    # an avocado-users at a higher PR than this build produces, dnf resolves to
+    # that older package and the image silently keeps `root:*` - the build goes
+    # green and the board still cannot be logged into. The symptom is
+    # indistinguishable from the feature not working.
+    #
+    # The fix is to make this build win, by bumping PR past whatever the feed
+    # holds. It is commented out because the right value depends on your feed,
+    # not on this file: shipping a fixed one here would be correct for the
+    # first person to hit this and wrong for the second.
+    #
+    # devtool-debt: PR must be bumped by hand when the feed is ahead.
+    # Ceiling: a feed whose avocado-users PR is at or below this build's.
+    # Upgrade trigger: someone hits the silent `root:*` a second time - at that
+    # point make the recipe derive PR from AVOCADO_DEV_ROOT_LOGIN instead, so
+    # the toggle cannot lose to its own previous output.
+    #
+    # PR:pn-avocado-users = "r1"
--- a/kas/feature/ssh-dev.yml
+++ b/kas/feature/ssh-dev.yml
@@ -1,0 +1,43 @@
+header:
+  version: 16
+
+# Bench/HITL SSH access, on any machine.
+#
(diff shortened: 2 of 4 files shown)


jetm:
Heads-up from taking this to hardware: the host-key work here is necessary but not sufficient on its own — sshd still will not start, because there is no `sshd` privilege-separation account.

`avocado-users` installs a fixed `/etc/passwd` and `/etc/group` over whatever the useradd class produces, so openssh's `USERADD_PARAM:${PN}-sshd = "... --user-group sshd"` is discarded. OpenSSH refuses to run without that account and exits before writing a banner or a log line, so the symptom is indistinguishable from the missing-key one: connections dropped at `kex_exchange_identification`, and `journalctl -u sshd@` completely empty.

#304 adds the account against this branch. The two changes are independent and neither supersedes the other — a key in a writable directory still gets no sshd without the account, and the account alone still leaves the key on a read-only path — so merge order does not matter, but **sshd will not come up until both have landed**.

Flagging it mainly so this PR does not read as not-working when validated on its own.

mobileoverlord (approved):
Approving. Verified beyond CI, which only parses the container-sdk config and never loads kas/feature/*:

- Resolved `machine/imx93-frdm.yml:feature/ssh-dev.yml` with `kas dump` on a merge of this branch onto current scarthgap. Both `feature/dev-root-login` and `feature/ssh-dev` land in local_conf_header and meta-avocado-distro is in the layer set, so the include works. On the branch as-is it cannot resolve because its base still pins the orphaned oe-core 06b1c4f; that goes away on merge.
- Traced the 9.6p1 path: the shipped sshd_config keeps `#HostKey` commented, so the bbappend takes the printf branch. read_only_rootfs_hook runs in do_rootfs and the IMAGE_PREPROCESS append runs in do_image afterwards, so `SSHD_OPTS=` is the last assignment for both sshd_check_keys (sourced) and sshd@.service (EnvironmentFile). /var is its own partition in fstab.

Non-blocking:
- This diverges from wrynose, which points sshd_config_readonly at /var/lib/ssh and leaves SSHD_OPTS alone. 9.6p1 has no OPENSSH_HOST_KEY_DIR_READONLY_CONFIG, but the same shape is one sed on sshd_config_readonly in the bbappend, with no image hook and no ordering argument. Worth doing if this file is touched again.
- `return 0` inside a `do_install:append` body returns from the whole do_install, so a later append from another layer would be skipped in the no-sshd_config case. An if/else around the two branches avoids that.
- The include precedent on scarthgap is kas/base.yml -> kas/vendor/oe.yml; containers.yml here does not include virtualization.yml.
- The feed's openssh-sshd RPM now ships HostKey lines pointing at /var/lib/ssh, so extensions using the packaged sshd_config on the read-only rootfs get working keygen too.

PR #277, opened 2026-08-13
nvidia: suppress redundant usb twiddle prompts in initrd-flash

lee-reinhardt:
## Problem

When `initrd-flash.sh` runs inside the avocado-vm, every `disconnect_usb_device` call becomes a real user unplug/replug prompt via the vm agent (`AVOCADO_AGENT_SOCK`). Two kinds of call sites issued prompts nobody should answer:

1. **Duplicate prompt mid-provision.** `generate_flash_package` disconnects twice back to back. The existing device-gone idempotency guard cannot catch the second call after a *manual* twiddle: the user's replug re-enumerates the device, so the rescan finds it present and requests another cycle immediately after a correct one. Operators who obey the second prompt yank the device while it is processing commands, which has broken a provision.
2. **Phantom teardown prompts.** The releases in `get_final_status` and at end of script prompted after flashing had already completed. Each request aged out the host's 120s replug waiter before falling back to the sysfs toggle, appending minutes of dead wait to every successful provision.

## Change

`disconnect_usb_device` gains an optional third argument:

- `reuse-cycle` — the caller only needs the device to have observed a disconnect edge since commands were staged. If a successful agent twiddle (manual cable cycle) already delivered one, skip the prompt. `generate_flash_package` clears the tracking flag at entry so a cycle from an earlier stage can never satisfy it.
- `no-prompt` — teardown/cleanup release: never engage the agent path, keep only the free sysfs toggle.

`unmount_and_release` threads the mode through. With `AVOCADO_AGENT_SOCK` unset (bare Linux / CI), the flag is never set and every call site behaves exactly as before.

The agent request now also prints `ACTION NEEDED: unplug and replug the device's USB cable to continue provisioning` to stderr before blocking, so CLI-driven provisions surface the prompt in the terminal instead of stalling silently.

## Validation

- `bash -n` clean; 20-case stub harness (real functions, fake agent transport) covering all three modes, cold/warm flag states, device-gone idempotency, and no-agent parity.
- Full Jetson Orin Nano hardware provision through the desktop: exactly two prompts, no duplicate after either replug, and the final-status stage completed in ~3s (previously ~120s of phantom-request timeout). `Final status: SUCCESS`.
- Adversarial cross-model review returned zero findings.

### DIFF
--- a/meta-avocado-nvidia/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
+++ b/meta-avocado-nvidia/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
@@ -191,6 +191,11 @@ _twiddle_via_agent() {
     req=$(printf '{"loc_hint":"%s"}' "$usb_instance")
     local resp=""
 
+    # Terminal-visible prompt: CLI-driven provisions have no banner UI,
+    # so without this line the flash just stalls silently while the
+    # host waits for the user. Desktop flows show a banner as well.
+    echo "ACTION NEEDED: unplug and replug the device's USB cable to continue provisioning" >&2
+
     # Two transports — prefer nc -U for terseness, fall back to python3.
     # Either should be available in the SDK container. 130s overall
     # timeout: the host has a 120s waiter for user replug, +10s slack.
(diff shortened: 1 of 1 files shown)


mobileoverlord (approved):
All three disconnect call sites classified correctly (reuse-cycle mid-provision after a manual twiddle, no-prompt on both teardowns), the stage-start reset keeps a prior stage's cycle from satisfying the next, and the terminal prompt fixes a real silent stall. Avocado-agent-specific, so nothing to reconcile against upstream meta-tegra.

PR #279, opened 2026-08-13
imx93: AHAB signed boot, end to end on hardware

jetm:
Builds on #271. Takes AHAB from "designed" to a board that boots a fully signed
chain: NXP-signed ELE container, OEM-signed `imx-boot`, OEM-signed OS container
carrying the initramfs, Linux, LUKS2 `/var` underneath it.

```
Authenticate OS container at 0xa8000000
Starting kernel ...
[  OK  ] Found device /dev/mapper/var.

# findmnt -no SOURCE,FSTYPE /var
/dev/mapper/var btrfs
```

There is no `Loading Ramdisk` line, and that absence is the result: the initrd
that derives the `/var` key is inside the signature rather than beside it.

## What was actually wrong

Three variables look like the AHAB switch and none is. `UBOOT_SIGN_ENABLE` is the
FIT/HABv4 path, referenced only in `compile_mx8m()`; `SECO_FIRMWARE_NAME` is the
NXP-signed ELE blob needed to boot at all; `SECOEXT_FIRMWARE_NAME` is defined
upstream only for mx8ulp/mx95/mx943 and its sole consumer is
`COMPATIBLE_MACHINE = "avocado-imx95-frdm"`, so the `"none"` shim is inert here.

Signing is SPSDK, not CST. `cst_signer`'s source refuses AHAB without SPSDK and
its README puts i.MX 8x/8ULP/9 on that path, so `nxpimage ahab sign` takes the
finished container and finds the header and signature block itself - no offsets
scraped from a build log, and no `var-cst-signer` to port.

**`CONFIG_AHAB_BOOT` changes what `booti` means.** Its first argument becomes a
container: `cmd/booti.c` authenticates it, then calls `container_get_image_dst()`
on the same address to find the kernel. Authentication has already copied each
image to the destination recorded inside the container, so staging the container
at the kernel's own destination means authentication destroys what the next line
parses. That hides behind `auth_cntr`, which calls the same function and passes
because nothing reads the container afterwards.

**A second `--data` payload does not carry the initramfs.**
`container_get_image_dst()` returns the first exec image and the first data image
and stops, so a second data payload is authenticated, placed, and then never
located; `booti` still reads `argv[1]`. Keeping the separate load only for its
size is worse: `${filesize}` comes from the unauthenticated copy on disk, so a
larger file leaves unverified bytes past the authenticated region, and appending
a cpio archive is the documented way to overlay files in an initrd. Bundling into
the kernel removes the argument instead of guarding it.

## Contents

- `avocado-ahab-sign.bbclass` - shared signing and one fail-closed check. Two
  copies of that logic could drift, and a drift there is a security bug.
- `avocado-os-container.bb` - packs the initramfs-bundled Image + dtb at this
  machine's addresses and signs it. Asserts at build time that the measured
  kernel reaches neither the device tree nor the container.
- `ahab.cfg` + `u-boot-imx` bbappend - `CONFIG_AHAB_BOOT=y`, verified against the
  produced `.config`. The bbappend also rewrites the boot flow, `fdt_addr` and
  `bootm_size`, for the signed path only.
- `INITRAMFS_IMAGE_BUNDLE` - the initramfs rides inside the signed exec payload,
  so `booti` takes no ramdisk argument. Closes the gap this PR previously listed
  as known.
- Boot partitions 128 -> 256 MiB under AHAB, `recovery` included since it is
  `bootpart 5` and holds a boot filesystem too. The signed container is ~200 MB
  once the initramfs is in it, because oe-core's `copy_initramfs` unpacks the
  `.cpio.zst` and the kernel embeds what it is given.
- `no-efi-capsule-auth.cfg` - **not** gated on AHAB. NXP's FRDM defconfigs enable
  capsule authentication, whose ESL step needs `cert-to-efi-sig-list`, which no
  recipe provides; unsigned builds were equally broken, and earlier builds only
  worked through a stale `tmp/hosttools` symlink. `avocado-imx95-frdm` carries
  the same symbol and was left alone rather than changed unverified.
- `kas/feature/ahab.yml` - opt-in, no working default. Asking for `ahab` without
  keys halts the build rather than shipping an unsigned image that looks signed.
- `scripts/avocado-flash` - the bash tool replaced by the Python port, plus `ums`
  (writes the board's own card in place) and `fuse-srk` verbs.

## What is opt-in, and what is not

Behind the `ahab` feature: the signing class, the OS container recipe, the
imx-boot signing include, the stone boot-file list and partition sizes, the
U-Boot environment rewrite, and `INITRAMFS_IMAGE_BUNDLE`.

Unconditional, deliberately: the capsule-auth fix (an unsigned build was broken
too, so gating it would leave it broken), two names added to
`BB_BASEHASH_IGNORE_VARS`, `AVOCADO_AHAB_FAMILY` / `AVOCADO_KERNEL_DTB` (read
only by gated code), and `scripts/avocado-flash`.

An unsigned build of these boards behaves as it did before this branch. The
signed path's `fdt_addr` and `bootm_size` are applied by the AHAB rewrite rather
than carried in the shared environment, so they do not reach it.

## The memory map had to move

The addresses were chosen for a ~33 MB kernel. At ~200 MB the kernel unpacks
across the device tree, and U-Boot stops with `FDT image overlaps OS image`.
Moving the container then hit a second wall: DRAM on this board is not
contiguous - `bdinfo` reports `0x80000000+0x16000000` and `0x98000000+0x68000000`
with 32 MiB missing between them - and a load spanning the hole is refused with
`Reading file would overwrite reserved memory`, which names neither the hole nor
the size. The container now stages at `0xa8000000`, in the second bank and above
the `no-map` carve-out Linux later claims as `ele-reserved`. NXP's 256 MiB bootm
window moves with it.

The build-time check compares ranges rather than addresses. The previous check
asked whether `cntr_addr` equalled `image_addr`, which is true of any layout that
fails this way and false of every layout that did fail here.

## Fusing

`fuse-srk` is a sibling of `sd`/`emmc` and never reachable from them: flashing is
repeatable, fuses are not. `ahab_close` is absent rather than gated. Dry run is
the default and it refuses on a non-blank fuse field, a non-open lifecycle, an
unsigned bootloader, or a read-back mismatch.

**The lab board has been fused, and wrongly.** `SRK_HASH` was burned
byte-swapped, so it is spent and `ahab_status` reports `BAD_KEY_HASH` on that
unit permanently. The board still boots because it is `OEM Open`, and
**`ahab_close` must never be run on it.** The tool now reads the fuse file
little-endian and asserts the ELE read-back after burning. Neither has been
proven against a passing burn, which needs a second board.

## Known gaps

- `ahab_status` reports `BAD_KEY_HASH` on the lab board because the burned hash
  is wrong, not because none is burned. A board with a correct burn is needed to
  see it clear.
- Signing is not reproducible: ECDSA draws a fresh nonce, so a signed bootloader
  has a new checksum every build. Matters for sstate reuse and OTA deltas.
- The initramfs is embedded uncompressed, because the kernel's compression choice
  depends on `INITRAMFS_SOURCE` and that variable only reaches the kernel as a
  make argument during `do_bundle_initramfs`. Compressing it would need an
  oe-core class override; the boot partition was widened instead.


jetm:
Rebased onto the current tip of `imx93-secure-boot-wrynose` (`81b6b04`). One
conflict, in `meta-avocado/conf/distro/include/avocado-security.inc`, and it
was not a context collision.

`6c471b4` on the base branch removed `BB_BASEHASH_IGNORE_VARS:append = "
AVOCADO_SB_KEYS_DIR"` and replaced the comment above it with the reasoning for
why a key directory path has to stay in the sstate hash: nothing else about the
key material reaches the task signature, so ignoring the path collapses two
different key sets to one hash and a shared sstate mirror can serve a package
built against the wrong keys.

This branch added a second `BB_BASEHASH_IGNORE_VARS:append` for
`AVOCADO_AHAB_KEYS_DIR` and `AVOCADO_AHAB_SPSDK_BINDIR`, written when the
SB_KEYS line was still there and explicitly citing it as precedent. Keeping it
would reintroduce that exact hazard for the AHAB root of trust, where the
consequence is worse: `do_sign_ahab` reads `AVOCADO_AHAB_KEYS_DIR` directly, so
a collision means a `flash.bin` signed by key set B handed to a build pointing
at key set A, and the symptom does not show up until the image is on a board
whose fuses are already burned.

So I dropped the append and kept the documentary half of the hunk as a comment:
both variables stay in the hash, and neither gets a default, for the reason the
original text gave. The cost is an sstate miss for anyone whose AHAB tree or
SPSDK install sits at a different path, which is the same cost the base branch
already accepted for the UEFI keys.

Nothing else moved. The diff against the new base is byte-identical to the
pre-rebase diff everywhere except that file, all 19 commits replayed in order,
and the base's `optee-ftpm-setup.sh` fixes (the `blkid` probe before format, the
whole-second sleep) survived the move to
`meta-avocado/recipes-security/optee-ftpm-init/`.

If the SPSDK-bindir cache miss turns out to be annoying in practice, I'd revert
just that variable to the ignore list on its own - it's a tool path rather than
key material, so the case for excluding it is weaker but not absurd.

jetm:
Superseded by #314, which is the same work rebased onto `wrynose`: 108 commits down to 18, since 86 of the originals were already upstream by patch-id.

Closing this rather than force-pushing over it, so this head stays recoverable on `imx93-ahab-secure-boot`.

#314 carries one blocker this PR did not name: `wrynose` moved imx93 to a single signed FIT booted with `bootm`, and the AHAB environment rewrite here still assumes the three-file `booti` layout. Because that rewrite is `sed -i`, the mismatched expressions fail silently rather than erroring. Details are in #314's Known gaps.

PR #252, opened 2026-08-03
meta-avocado: add meta-avocado-sbom

hiagofranco:

PR #281, opened 2026-08-13
meta-avocado: add meta-avocado-sbom

hiagofranco:
Forward port of scarthgap 62fc186 ("meta-avocado: add meta-avocado-sbom (#252)"), reworked: wrynose OE-core deleted cve-check.bbclass. CVE analysis is now sbom-cve-check, which reads an SPDX 3.0 document, matches its purl/CPE identifiers against the cvelistV5 and NVD databases the sbom-cve-check-update-*-native recipes fetch, applies the VEX statements in that same document, and exports a manifest in the JSON shape the old class wrote per recipe.

The layer scans the whole build in one pass, by appending sbom-cve-check-recipe to OE-core's meta-world-recipe-sbom - the aggregate SBOM of every world target. Per-recipe scanning re-analyses each recipe's full dependency closure once per recipe; per-image scanning covers one image, while avocado ships a package feed that images, extensions and sysexts are composed from. avocado-cve-report then joins that manifest with the package-to-recipe map in pkgdata, unchanged in output: which CVEs affect the packages actually installed.

Two kas fragments, deliberately separate so the build pipeline can take SPDX without paying for the CVE scan:

  kas/feature/sbom.yml       SPDX_INCLUDE_VEX ??= "current"
  kas/feature/cve-check.yml  SPDX_INCLUDE_VEX = "all", AUTOREV on the CVE
                             database recipes, and the meta-avocado-sbom
                             layer

create-spdx is create-spdx-3.0 on wrynose and already inherited through INHERIT_DISTRO, so scarthgap's INHERIT swap is gone. The scan reads VEX to decide Patched and Ignored, so cve-check.yml raises SPDX_INCLUDE_VEX to "all": under "current" the fixed-version and cpe-stable-backport statements never reach the document, and what cve-exclusion_6.18.inc asserts would come back as Unpatched. The weak assignment in sbom.yml keeps fragment include order from deciding the value.

Exports land in ${DEPLOY_DIR_IMAGE}, which is already per MACHINE, so CVE_CHECK_DIR and its machine scoping have no successor here. Presence of a manifest entry, not cvesInRecord, now separates a scanned-clean recipe from an unexamined one: sbom-cve-check sets cvesInRecord from the exported issue list, so a clean recipe carries "No" too.

One manifest per build also changes how errors have to be scoped: a malformed record now costs that entry (counts.entries_unreadable) instead of the file, since cve_files_unreadable >= cve_files is the "nothing was scanned" condition and would fail the build over one bad record. Repeated recipe names are merged by CVE id rather than overwritten, so counts and report body cannot disagree when AVOCADO_CVE_MANIFEST points at a directory. A CVE whose status is none of Unpatched/Patched/Ignored now warns instead of vanishing into a counter.

packagegroup-avocado-sdk-all already inherits nospdx on wrynose.

Not built: report.py verified end to end against a synthetic manifest and pkgdata tree, manifest shape against sbom-cve-check 1.3.1, class and variable names against oe-core 709c4fd6. Details in docs/migrations/scarthgap-to-wrynose.md.

### DIFF
--- a/docs/migrations/scarthgap-to-wrynose.md
+++ b/docs/migrations/scarthgap-to-wrynose.md
@@ -421,3 +421,32 @@ commit subject (`scarthgap` vs the forward-port history).
 - `f347c52`, `3fb9561` — vendor-fork re-pins; see the two concrete
   fork-port items under Open Questions above (oe-core symlink patch,
   bitbake LFS subdir fix). Not distro cherry-picks.
+
+## Forward-Port Round — 2026-08-13
+
+### `meta-avocado: add meta-avocado-sbom` (scarthgap `62fc186`, PR #252)
+
+Not a cherry-pick: wrynose OE-core deleted `cve-check.bbclass`. CVE
+analysis is now `sbom-cve-check` — the external
+[sbom-cve-check](https://github.com/bootlin/sbom-cve-check) tool
+(`python3-sbom-cve-check-native`) reads an SPDX 3.0 document, matches its
+purl/CPE identifiers against the cvelistV5 and NVD databases fetched by
+`sbom-cve-check-update-{cvelist,nvd}-native`, applies the VEX statements
+in that same document, and exports the result. Three classes wrap it:
+`sbom-cve-check` (per image, via `IMAGE_CLASSES`),
+`sbom-cve-check-recipe` (per recipe, whole dependency closure), and
+`sbom-cve-check-common` (shared).
+
+What the port changed against scarthgap's commit:
+
+| Scarthgap | Wrynose | Why |
+|---|---|---|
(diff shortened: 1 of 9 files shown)


hiagofranco:
Keeping as a draft since I need to check on a Wrynose build that I do not have yet.

mobileoverlord (approved):
Green CI on both parse-container-sdk arches. Pure addition (+1022/-0, 9 files): new meta-avocado-sbom layer, kas/feature/{sbom,cve-check}.yml, the avocado_sbom report lib and avocado-cve-report recipe, plus the scarthgap-to-wrynose migration note. No existing recipe or config paths modified, so no regression surface in the current build. Verified in use: this commit is the base of the sbom-wrynose test branch that has been building and provisioning jetson-agx-orin and jetson-orin-nano all day.

PR #283, opened 2026-08-13
nvidia: fix shadowed agent usb disconnect + suppress teardown replug prompts

lee-reinhardt:
## Problem

The wrynose BSP rebase re-imported upstream copies of seven `initrd-flash.sh` helper functions without removing the agent-aware block grafted at the top of the file. Bash keeps the later definition, so the live `disconnect_usb_device` was the plain sysfs-only copy and `_twiddle_via_agent` (plus the device-gone idempotency guard) was dead code. Any provision running inside the avocado-vm would hang at the first cable-cycle point with no prompt anywhere, since the `authorized=0` toggle is a no-op through vhci_hcd.

Separately, the teardown-prompt fix from #277 (scarthgap) had not been ported, so once the agent path is live again, the releases in `get_final_status_t234` and the end-of-script lingering-device deauthorize would each age out the host's 120s replug waiter after flashing already completed.

## Change

Commit 1 — remove the duplicates. Six of the seven pairs are byte-identical modulo whitespace; only `disconnect_usb_device` differs, by exactly the agent path and the device-gone guard. Keep the agent-aware copy, delete the shadowing re-imports. Pure deletion, −202 lines.

Commit 2 — port the relevant half of #277:

- `disconnect_usb_device` gains an optional `no-prompt` mode (threaded through `unmount_and_release`): teardown/cleanup releases never engage the agent, keeping only the free sysfs toggle. Applied to `get_final_status_t234`'s release and the end-of-script deauthorize.
- `_twiddle_via_agent` prints `ACTION NEEDED: unplug and replug the device's USB cable to continue provisioning` to stderr before blocking, so CLI-driven provisions surface the prompt in the terminal instead of stalling silently.
- The `reuse-cycle` mode from #277 is intentionally not ported: the back-to-back double disconnect in `generate_flash_package` that required it does not exist in the wrynose flow.

With `AVOCADO_AGENT_SOCK` unset (bare Linux / CI), every call site behaves exactly as before in all modes.

## Validation

- `bash -n` clean; no duplicate function definitions remain (verified by scan).
- 13-case stub harness (real `_twiddle_via_agent` / `disconnect_usb_device` / `unmount_and_release` extracted from this file, fake agent on a real AF_UNIX socket): mode threading, device-gone idempotency, prompt emission, and no-agent parity all pass.
- Hardware bench on a Jetson Orin Nano is pending a wrynose Orin runtime build; the equivalent scarthgap change was validated end-to-end on real hardware in #277.

---

## Added on top (jschneck)

Three further commits, landed here rather than separately because they were developed and hardware-validated on top of this branch and share its file:

**`nvidia: flash the module's own SDRAM config and the real ODMDATA`** — two bugs in `tegra234-flash-helper.sh`, both the same shape: a value taken from a positional argument where upstream `tegra-flash-helper.sh` reads it from `flashvars`.

- The `BOARDSKU` 0005/0008 branch rewrote `EMC_BCT`, which is *not in scope* in that script (it sources only `./flashvars`; `EMC_BCT` lives in the un-exported `.env.initrd-flash`). Meanwhile `--sdram_config` was given `$sdramcfg_files` (`$3`), never rewritten. A P3701-0005 64GB module was flashed with the `-0000` 32GB SDRAM config, MB1 skipped `MSS CAR: Init PLLM/Init PLLHUB` and halted in `MB1_MSS_CHECKER` at `Task: Check MSS settings`. **jetson-agx-orin could not be flashed at all.** p3767 targets were unaffected because their build-time value already equals what the rewrite would produce.
- ODMDATA arrived as positional `$4`, but `.env.initrd-flash` sets `ODMDATA=""` where there is no carrier override and the argument was unquoted — so it collapsed, `$4` landed on the next parameter, and tegraflash got `--odmdata boot.img`. MB1 logged `Skipping GBE UPHY config`, the MGBE UPHY lanes were never applied, and the AQR113C was unreachable (`phy_id` = 0). **No ethernet on agx-orin.** Now read from the environment when set, else flashvars, and guarded — `initrd-flash` forwards it so `CARRIER_ENV_ODMDATA` overrides still win (bsp-mic-712-ox-16gb and bsp-mic-733-ao5a1/ao6a1 depend on that).

**`meta-avocado-distro: stop skipping the alt-mc merge on every Jetson`** — `do_multikernel_merge` derived the alt kernel version with a glob requiring an RPi-style `-v<digit>` ABI suffix, which Jetson's `kernel-6.8.12-l4t-r39.2.0-…` does not have. It matched nothing and skipped the merge as a `bbwarn`, so 1611 L4T RPMs never reached the unified feed and any extension package scoped to `kernel-6.8.*` was unresolvable at provision time.

**`nvidia: repin meta-tegra/-community/meta-python-ai and adapt`** — vendor pins to branch tips, plus the fallout: carrier-include rename (`devkit-wifi.inc` → `p3768.inc`, new `p3737.inc`), two more OOT entries stripped from `packagegroup-machine-base`, dropped `tegra-wifi`/`tegra-bluetooth` bbappends (upstream deleted both recipes), `systemd-jetson-masks` for the Tegra Combined UART getty preset, and DeepStream 9.1 now that the libnvvpi3 dependency is gone.

## Hardware validation

Covers this PR's own change too, which closes out the pending Orin Nano bench above.

| | jetson-agx-orin (P3701-0005 / P3737-0000) | jetson-orin-nano (P3767-0005 / P3768) |
|---|---|---|
| flash | `Final status: SUCCESS` | `Final status: SUCCESS` |
| `--sdram_config` | `tegra234-p3701-0005-sdram-l4t.dts` | `tegra234-p3767-0001-sdram-l4t.dts` |
| `--odmdata` | `gbe-uphy-config-22,…,gbe0-enable-10g` | `gbe-uphy-config-8,hsstp-lane-map-3,hsio-uphy-config-0` |
| MB1 | `MSS CAR: Init PLLM` / `Init PLLHUB` present, MSS check passes | unchanged |
| ethernet | `eth0` 1000Mb/s full duplex, DHCP, routed traffic | `eth0` up, DHCP, routed traffic |
| agent USB path | clean disconnect/reconnect, no replug prompts, `Device already detached — flash completed cleanly` | same |
| multikernel feed | — | 1611 L4T RPMs merged; `libc6` still appears once |

Note this branch was tested with #281 (meta-avocado-sbom) already merged into wrynose.

### DIFF
--- a/kas/feature/python-ai.yml
+++ b/kas/feature/python-ai.yml
@@ -5,7 +5,7 @@ repos:
   meta-python-ai:
     url: https://github.com/avocado-linux/vendor-meta-python-ai.git
     path: meta-python-ai
-    commit: 596a00168046c5c8db4c240c1fee9e3f73fbc285
+    commit: 91cc20407148acbbe1ec2dbb295c240ec5c2cbeb
     branch: wrynose
     layers:
       .:
--- a/kas/vendor/nvidia.yml
+++ b/kas/vendor/nvidia.yml
@@ -6,7 +6,7 @@ repos:
     url: https://github.com/avocado-linux/vendor-meta-tegra
     path: meta-tegra
     branch: wrynose
-    commit: 6a0294838e8e0180a40cd3d48d7d0fce1deb3ef2
+    commit: 727633deca9b72643112d1aebadafb736dad60c0
     layers:
       .:
 
@@ -14,7 +14,7 @@ repos:
     url: https://github.com/avocado-linux/vendor-meta-tegra-community
     path: meta-tegra-community
     branch: wrynose
-    commit: 5c2a026204d47398f4a12d750f6ebeef4ac01802
+    commit: f9600de48fd7c7dfd711af71430f2afb3705cf33
     layers:
       .:
 
--- a/meta-avocado-distro/classes/avocado-multikernel.bbclass
+++ b/meta-avocado-distro/classes/avocado-multikernel.bbclass
@@ -65,11 +65,22 @@ do_multikernel_merge() {
         # So filter rpm/ to ONLY the alt kernel's version-tagged RPMs. This
         # depends solely on the alt mc's own output — never on the prunable
         # default deploy state — so it is immune to prune/ordering races. The
-        # version is derived from the alt mc's abiversioned kernel package
+        # version is derived from the alt mc's kernel package
         # (e.g. kernel-6.6.63-v8-... -> 6.6.63); it tags every kernel RPM (PV
         # 6.6.63+git..., KERNEL_VERSION 6.6.63-v8) and none of the common NEVRAs
         # (base-files 3.0.14, packagegroup 1.0, shadow-securetty 4.6, ...).
-        altkver="$(ls "${base}/rpm/"*/kernel-[0-9]*-v[0-9]*.rpm 2>/dev/null \
+        #
+        # Match kernel-<N.N.N> without requiring an ABI suffix. The glob used to
+        # be kernel-[0-9]*-v[0-9]*.rpm, which only matched the Raspberry Pi
+        # shape (kernel-6.6.63-v8-...). Jetson's L4T kernel is
+        # kernel-6.8.12-l4t-r39.2.0-1021.21-... with no -v<digit>, so the glob
+        # matched nothing, altkver came back empty, and EVERY Jetson build
+        # skipped the alt-mc merge with only a bbwarn -- leaving the L4T
+        # kernel's RPMs out of the unified feed entirely. kernel-[0-9]* still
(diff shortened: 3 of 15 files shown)


mobileoverlord (approved):
Green CI on both arches. Reviewed the duplicate-removal commit against the kept definitions: the six byte-identical pairs check out, and the disconnect_usb_device pair differs by exactly the agent path plus the device-gone guard, so keeping the agent-aware copy is right. The no-prompt threading through unmount_and_release is correct at both new call sites.

Hardware validation now covers the pending Orin Nano bench and more — the agent USB path ran clean on both jetson-agx-orin and jetson-orin-nano (disconnect/reconnect, no replug prompts, clean detach at end of script) across several full provisions today.

Also approving the three commits added on top; each is described in the PR body with its failure mode and the on-hardware evidence.

PR #287, opened 2026-08-13
avocado-stone: keep image inputs out of rm_work

### LINKED ISSUE #284: rm_work + nostamp stone tasks silently produce an empty initramfs (build exits 0, device panics)
With the default `AVOCADO_RM_WORK=1` (`kas/base.yml` → `USER_CLASSES += "rm_work"`), `avocado-stone.bb` declares:

```
do_stone_bundle[nostamp]    = "1"
do_stone_provision[nostamp] = "1"
do_compile[depends] += "avocado-image-initramfs:do_image_complete"
do_compile[depends] += "avocado-image-rootfs:do_image_complete"
```

`nostamp` re-runs those tasks on every invocation, which re-triggers `do_image_complete` against a workdir rm_work has already deleted. The image is then repacked from an empty tree: `avocado-image-initramfs-<machine>.cpio.zst` comes out as a ~129-byte zstd frame containing a 1 KiB cpio with a 0-byte `/init` — and **bitbake still exits 0**. The device then panics at boot:

```
Failed to execute /init (error -13)
Kernel panic - not syncing: No working init found.
```

Every build signal says success; the failure only shows up on the booted target. `bitbake -c cleansstate <image>` fixes it until the next `stone_provision`.

Reproduce: build any machine's `avocado-stone`, run `-c stone_provision` twice, check the initramfs artifact size in the deploy dir.

Suggested fix (what we run with): `RM_WORK_EXCLUDE += "avocado-image-initramfs avocado-image-rootfs avocado-image-var avocado-stone"` carried in `avocado-stone.bb` itself — or replacing the nostamp+depends pattern with proper task checksums so the re-run is only triggered when inputs actually changed. Happy to send a PR either way.

WarrDoge:
Fixes #284. With default `AVOCADO_RM_WORK=1`, the nostamp stone tasks re-run every invocation and re-trigger `do_image_complete` against workdirs rm_work already deleted — the initramfs repacks from an empty tree (~129-byte cpio.zst, 0-byte /init) while bitbake exits 0, and the device panics at boot. `RM_WORK_EXCLUDE` on the stone inputs is rm_work's supported escape hatch; costs some disk on those four recipes, buys the guarantee a re-bundle always has real workdirs. We've run this exclusion in production builds for a month across three machines.

### DIFF
--- a/meta-avocado/conf/distro/avocado.conf
+++ b/meta-avocado/conf/distro/avocado.conf
@@ -5,6 +5,16 @@ INITRAMFS_IMAGE = "avocado-image-initramfs"
 INITRAMFS_IMAGE_BUNDLE ?= "0"
 INITRAMFS_IMAGE_NAME = "${INITRAMFS_IMAGE}-${MACHINE_SHORT_NAME}"
 
+# rm_work + the nostamp stone tasks in avocado-stone.bb is a silent-data-loss
+# combination: nostamp re-runs do_stone_* on every invocation, which re-triggers
+# the images' do_image_complete against workdirs rm_work has already deleted. The
+# initramfs is then repacked from an empty tree — a ~129-byte cpio.zst holding a
+# 0-byte /init — and the build still exits 0; the failure only appears as an init
+# panic on the booted device. rm_work.bbclass matches RM_WORK_EXCLUDE against
+# ${PN} in each recipe's own datastore, so this has to be set distro-wide, not in
+# avocado-stone.bb. Fixes #284.
+RM_WORK_EXCLUDE += "avocado-image-initramfs avocado-image-rootfs avocado-image-var avocado-stone"
+
 AVOCADO_DEFAULT_DISTRO_FEATURES = "multiarch ptest systemd usrmerge secureboot security pam ldconfig openssl zram"
 DISTRO_FEATURES_EXTRA ?= ""
 DISTRO_FEATURES:append = "${AVOCADO_DEFAULT_DISTRO_FEATURES} ${DISTRO_FEATURES_DEFAULT} ${DISTRO_FEATURES_EXTRA}"


mobileoverlord (changes requested):
The diagnosis is right, but the fix as written doesn't reach the recipes it needs to. `rm_work.bbclass` checks `RM_WORK_EXCLUDE` against `${PN}` inside **each recipe's own datastore**:

```sh
for p in ${RM_WORK_EXCLUDE}; do
    if [ "$p" = "${PN}" ]; then
```

An append inside `avocado-stone.bb` only lands in avocado-stone's datastore — `avocado-image-initramfs`, `avocado-image-rootfs`, and `avocado-image-var` never see it, and their workdirs (the actual inputs the comment describes being deleted) still get rm_work'd. As written this protects only avocado-stone itself, so the empty-initramfs failure in #284 can still reproduce.

The exclusion needs global scope — the distro conf (e.g. `meta-avocado-distro/conf/distro/…`) is the natural home, keeping the excellent comment with it. Everything else about the change (the list of four recipes, the reasoning) is exactly right.

WarrDoge:
Right, thanks - the append only ever landed in avocado-stone's own datastore. Moved to distro scope and rebased on current `scarthgap`.

One correction to the suggested path: `meta-avocado-distro/conf/distro/` doesn't exist. The distro confs are in `meta-avocado/conf/distro/`, so the exclusion now sits in `avocado.conf` just under `INITRAMFS_IMAGE_NAME` — target distro only, `avocado-container.conf` untouched. The comment moved with it.

Verified on `avocado-qemuarm64` with our own identical local.conf workaround removed, so the distro conf is the only source:

```
$ bitbake-getvar -r avocado-image-initramfs --value RM_WORK_EXCLUDE
 avocado-image-initramfs avocado-image-rootfs avocado-image-var avocado-stone
```

which is exactly what the old patch failed to do - that recipe used to come back empty.

And the failure mode from #284 no longer reproduces: build, then a second `bitbake avocado-stone` so the nostamp `do_stone_*` tasks re-run - `avocado-image-initramfs-qemuarm64.cpio.zst` stays 39757181 bytes, same md5, instead of collapsing to the ~129-byte empty frame. The image boots and our full boot test suite is green on it (80/80).

PR #288, opened 2026-08-13
avocado-users: ship the video group

### LINKED ISSUE #285: avocado-users clobbers groups created during rootfs assembly (e.g. video/44 — udev rules dead on target)
`avocado-users` installs static `/etc/passwd`, `/etc/group`, `/etc/gshadow`, `/etc/shadow` in `do_install`, replacing whatever the files contained. Groups created by package preinsts/sysusers earlier in rootfs assembly are silently lost.

Concrete case: `udev` (and several graphics packages) expect the `video` group (gid 44); rules like `SUBSYSTEM=="video4linux", GROUP="video"` then fail to apply on the target because the group doesn't exist in the shipped `/etc/group`, and device nodes fall back to root-only. Nothing fails at build time — the breakage is only visible on the device when an application can't open `/dev/video0` as a non-root user.

Suggested fix: merge with (or at least preserve) groups already present in `${D}${sysconfdir}/group` instead of overwriting, or extend the static files with the stock oe-core group set. We currently carry a bbappend that re-adds the missing entries; a PR is easy if you agree on the direction.

WarrDoge:
Fixes #285 for the one group with a proven on-target failure: udev's `GROUP="video"` rules silently fail and V4L device nodes fall back to root-only, with nothing failing at build time. Adds `video:x:44:` (stock oe-core gid) to the static file. The broader clobber-vs-merge design question from #285 stays open — happy to follow up if you'd rather merge groups created during rootfs assembly instead of extending the static set.

### DIFF
--- a/meta-avocado-distro/recipes-avocado/avocado-users/files/group
+++ b/meta-avocado-distro/recipes-avocado/avocado-users/files/group
@@ -1,4 +1,5 @@
 root:x:0:
+video:x:44:
 systemd-network:x:993:
 systemd-timesync:x:994:
 systemd-resolve:x:995:


mobileoverlord (approved):
Right place for this. It can't come from an extension via sysusers.d: systemd-sysusers runs long before the extension merge, and the real consumer is udev's GROUP="video" at coldplug — a missing group there silently falls back to root on the device nodes, unfixable post-merge without a udevadm trigger sweep. GID 44 is the LSB-conventional value, so the static list stays clean. No collision with the existing 993-997/999 range.

PR #289, opened 2026-08-13
meta-avocado-qemu: stone depends on the TF-A deploy

### LINKED ISSUE #286: qemuarm64: stone tasks missing trusted-firmware-a:do_deploy dependency (flash.bin race)
On `avocado-qemuarm64`, `meta-avocado-qemu`'s `avocado-stone.bbappend` adds only:

```
```

but the machine's stone manifest also consumes `flash.bin` (TF-A) from the deploy dir. With a cold ss-state and enough parallelism, `do_stone_validate`/`do_stone_bundle` can run before `trusted-firmware-a:do_deploy` has produced `flash.bin`, failing the bundle — or worse, picking up a stale `flash.bin` from a previous build if one is lying around.

Suggested fix: `do_compile[depends] += "trusted-firmware-a:do_deploy"` next to the u-boot dep in `meta-avocado-qemu/recipes-avocado/distro/avocado-stone.bbappend`. One-liner PR available.

WarrDoge:
Fixes #286. The qemuarm64 manifest consumes `flash.bin` from the deploy dir, but only the u-boot deploy was declared. Cold parallel builds can bundle before `trusted-firmware-a:do_deploy` runs — or worse, with a stale flash.bin from a previous build. Scoped `:avocado-qemuarm64` so qemux86-64 is untouched.

### DIFF
--- a/meta-avocado-qemu/recipes-avocado/distro/avocado-stone.bbappend
+++ b/meta-avocado-qemu/recipes-avocado/distro/avocado-stone.bbappend
@@ -1,6 +1,12 @@
 inherit stone
 
 do_compile[depends] += "u-boot:do_deploy"
+# The qemuarm64 manifest also consumes flash.bin (TF-A) from the deploy dir;
+# without this dependency a parallel cold build can bundle before (or with a
+# stale copy of) the TF-A deploy. Gated on MACHINE because avocado-qemux86-64
+# shares this .bbappend and has no TF-A provider -- and because varflags take no
+# override suffix on bitbake 2.8.1. Fixes #286.
+do_compile[depends] += "${@bb.utils.contains('MACHINE', 'avocado-qemuarm64', 'trusted-firmware-a:do_deploy', '', d)}"
 
 DEPENDS += " jq-native mkfat-native fwup-native"
 


mobileoverlord (changes requested):
Right dependency, but the syntax doesn't parse on scarthgap's bitbake (2.8.1). Varflags don't take override suffixes — fed through the actual parser:

```
bb.parse.ParseError: unparsed line: 'do_compile[depends]:append:avocado-qemuarm64 = " b"'
```

so this .bbappend would break parsing of the recipe outright (there's no build CI on this repo to catch it). The scarthgap-safe form is an inline override check:

```
do_compile[depends] += "${@bb.utils.contains('OVERRIDES', 'avocado-qemuarm64', 'trusted-firmware-a:do_deploy', '', d)}"
```

(or gate on MACHINE if that reads better). The fix itself — bundling must wait for the TF-A deploy on qemuarm64 — is correct.

WarrDoge:
Confirmed and fixed - rebased on current `scarthgap`.

Worth flagging before anyone else reaches for it: `bb.utils.contains('OVERRIDES', ...)` parses, but it never fires. `bb.utils.contains()` does `set(val.split())` - whitespace - and `OVERRIDES` is colon-separated, so the subset test always fails and the false branch always wins. Measured on `avocado-qemuarm64` with that exact form:

```
$ bitbake-getvar -r avocado-stone --flag depends --value do_compile
 avocado-core:do_compile avocado-image-initramfs:do_image_complete avocado-image-rootfs:do_image_complete avocado-image-var:do_deploy virtual/kernel:do_deploy u-boot:do_deploy

$ bitbake-getvar -r avocado-stone --value OVERRIDES
 linux:aarch64:pn-avocado-stone:layer-meta-avocado-distro:qemuall:aarch64:stone-img:stone-direct:bootvars-ubootenv:qemuarm64-secureboot:qemuarm64:avocado-qemuarm64:avocado:class-target:libc-glibc:forcevariable
```

No TF-A dep - silently the same bug as before, only without the parse error to tell you.

So this takes your other suggestion, `MACHINE`, which is a single whitespace-free token:

```
do_compile[depends] += "${@bb.utils.contains('MACHINE', 'avocado-qemuarm64', 'trusted-firmware-a:do_deploy', '', d)}"
```

```
$ bitbake-getvar -r avocado-stone --flag depends --value do_compile
 ... u-boot:do_deploy trusted-firmware-a:do_deploy
```

`avocado-qemux86-64` takes the empty branch, so it is unchanged. Full build and boot test green on qemuarm64 (80/80).

PR #291, opened 2026-08-15
Fix five silent failures in Jetson provisioning

jetm:
## Problem

Provisioning a Jetson Orin Nano Developer Kit to an SD card fails in several
places, and none of them reports a usable cause. Each one presented as a
hardware or cable fault and was diagnosed by walking a process tree or reading
a serial console rather than by reading an error message.

Step 1 hangs indefinitely. The script writes a `cpp` wrapper into a temp dir and
prepends that dir to `PATH`; the wrapper body named the preprocessor by bare
word, so `exec cpp "$@"` re-resolved through `PATH`, found the wrapper, and
re-exec'd itself. The symptom is "Step 1: Signing binaries" with no output, no
error, and one process pinned at 100% CPU.

Step 4 times out after 120s with no cause. The sd branch told the target to
`export-devices mmcblk1`, a device that only exists on a module that also
carries eMMC: eMMC takes mmc0 and the card lands on mmc1. This developer kit
module has no eMMC, so the SD controller at `3400000.mmc` is the only one
instantiated and the card enumerates as `mmcblk0`.

That timeout then pointed at the wrong component. The target printed
`Export of mmcblk1 failed` to its debug UART and stopped re-enumerating, while
the host printed a bare timeout followed by a list of its own three SATA disks -
none of which have anything to do with the flash.

Running without root wasted the whole prefix of a run. `initrd-flash` mounts the
board's exported USB storage and never elevates on its own, so an unprivileged
run signed every binary and RCM-booted the board before dying ~40s in at
`could not mount USB storage for writing flashing commands`.

A board in Force Recovery is also unreachable without a udev rule, and that
failure is completely silent: the stock `51-android.rules` hands vendor `0955`
to `GROUP="adbusers"`, and a user outside that group makes tegraflash block on
the USB read emitting nothing at all.

## Solution

Resolve the preprocessor to an absolute path before writing the wrapper, and
keep the existing absolute-path guard, since `command -v` yields a bare word for
a builtin or alias.

Default the SD device to `mmcblk0` and expose `AVOCADO_PROVISION_SD_DEVICE` for
eMMC-bearing modules, rather than swapping one hardcoded name for the other.
Both layouts ship across Jetson modules, and a hardcode is what produced this
failure.

Check for root before anything else runs, ahead of even the manifest reads.
Placing it later would be worse than useless under `set -u`: the script would
die on an unbound variable and report that instead.

Make the export timeout name the target as the thing that failed, name the
device it was asked for, and point at the serial console - the host cannot
recover the target's message in that state, because the target never re-exports
the flashpkg LUN the host reads logs from.

## Key changes

- `stone-provision-tegraflash.sh`: wrapper execs a resolved absolute path
- `stone-provision-tegraflash.sh`: sd branch uses
  `${AVOCADO_PROVISION_SD_DEVICE:-mmcblk0}` for `BOOTDEV` and `ROOTFS_DEVICE`
- `stone-provision-tegraflash.sh`: refuses to run without root, before any
  signing work rather than 40s in at a mount failure
- `initrd-flash.sh`: the Step 4 export timeout attributes the failure to the
  target and points at the debug UART, instead of listing the host's own disks
- `adding-a-jetson-carrier.md`: records the root requirement and the `0955`
  udev/`uaccess` prerequisite
- Four standalone test suites, each verified to fail against the pre-fix code

## Reviewer notes

The through-line is that the failing component and the reporting component were
never the same one.

Verified end to end on an Orin Nano Developer Kit (P3767-0005): flash reports
`Final status: SUCCESS` and the board boots to an `Avocado OS` login prompt on
`ttyTCU0` from `mmcblk0p1`.

Modules that carry eMMC must set `AVOCADO_PROVISION_SD_DEVICE=mmcblk1`. That is
the case the BSP's own `generic/cfg/flash_t234_qspi_sd*.xml` describe, and
reading those as authoritative for all modules is what made the device-name bug
hard to see.

`shellcheck` reports SC2064 (line 315), two SC2086 (lines 427 and 533) and
SC2034 (line 444) in `stone-provision-tegraflash.sh`, and comparable
pre-existing findings in `initrd-flash.sh`. All are unchanged from the base
commit and outside the touched ranges, so they are left alone rather than fixed
drive-by.

### DIFF
--- a/meta-avocado-nvidia/docs/adding-a-jetson-carrier.md
+++ b/meta-avocado-nvidia/docs/adding-a-jetson-carrier.md
@@ -315,6 +315,29 @@ cause).
 
 ## Caveats
 
+- **Provisioning must run as root.** `initrd-flash` mounts the board's
+  exported USB storage to write the command package and never elevates
+  on its own, so the flash needs `CAP_SYS_ADMIN`. Run it under
+  `sudo -E` so the SDK environment survives. Unprivileged runs used to
+  sign the binaries and RCM-boot the board before dying at
+  `could not mount USB storage for writing flashing commands`, which
+  reads as a storage or cable fault; the provisioning script now
+  refuses up front instead.
+- **A Jetson in Force Recovery needs a udev rule to be reachable.** The
+  board enumerates as vendor `0955`, which the stock
+  `51-android.rules` hands to `GROUP="adbusers" MODE="0660"`. A user
+  outside that group cannot open the device, and tegraflash blocks on
+  the USB read with no message at all - the run simply sits there.
+  Either join `adbusers`, or drop in a rule granting the logged-in
+  user access directly:
+
+  ```udev
+  # /etc/udev/rules.d/60-jetson-recovery.rules
+  SUBSYSTEM=="usb", ATTR{idVendor}=="0955", TAG+="uaccess"
+  ```
+
+  `uaccess` is preferable: it grants access via ACL on device add, so
+  it needs no group membership and no re-login.
 - **EEPROM SKU mismatch aborts signing.** If the actual SOM SKU
   doesn't match `CHECK_BOARDSKU` in flashvars, tegraflash refuses to
   sign with `actual board SKU X does not match expected board SKU Y`.
--- a/meta-avocado-nvidia/docs/adding-a-jetson-carrier.md
+++ b/meta-avocado-nvidia/docs/adding-a-jetson-carrier.md
@@ -338,6 +338,41 @@ cause).
 
   `uaccess` is preferable: it grants access via ACL on device add, so
   it needs no group membership and no re-login.
+- **SD boot needs the card's device name declared per board.** The
+  `tegraflash-sd` profile has to tell the target which block device to
+  export, and that name depends on the module, not on the carrier: a
+  module without eMMC instantiates only the SD controller, so the card
+  is `mmcblk0`; a module carrying eMMC puts eMMC on `mmc0` and the
+  card on `mmc1`. The BSP's `generic/cfg/flash_t234_qspi_sd*.xml` say
+  `mmcblk1p1` because they are written for the latter. Declare it in
+  the board's stone manifest:
+
+  ```json
+  "storage_devices": {
+    "rootdisk": {
+      "tegraflash": { "sd_device": "mmcblk0" }
+    }
+  }
+  ```
+
+  `AVOCADO_PROVISION_SD_DEVICE` overrides the manifest for a one-off
+  run (a board revision that enumerates differently). Both accept a
+  bare `mmcblkN` name - a `/dev/...` path is rejected. Provisioning
+  with the SD profile and neither set is an error rather than a guess,
+  since guessing wrong writes at the module's internal storage. In
+  tree today, and how far each value is trusted:
+
+  | Board | `sd_device` | Basis |
+  |---|---|---|
+  | `jetson-orin-nano-devkit` | `mmcblk0` | Confirmed on a P3767-0005 |
+  | `recomputer-j301x` | `mmcblk0` | Same Orin Nano module (its machine conf requires `orin-nano.inc`, whose upstream `TNSPEC_BOOTDEV_DEFAULT` is `mmcblk0p1`); not yet flashed |
+  | `jetson-agx-orin-devkit` | `mmcblk1` | The value the tree used before this was declarable; not yet flashed |
+  | `jetson-orin-nx` | `mmcblk1` | Same |
+  | `recomputer-j401x` | `mmcblk1` | Same |
+
+  Check yours against the target's own console
+  (`mmcblkN: mmcN:0001 USD 29.8 GiB`) the first time you flash it, and
+  correct the manifest if it disagrees.
 - **EEPROM SKU mismatch aborts signing.** If the actual SOM SKU
   doesn't match `CHECK_BOARDSKU` in flashvars, tegraflash refuses to
   sign with `actual board SKU X does not match expected board SKU Y`.
--- a/meta-avocado-nvidia/docs/adding-a-jetson-carrier.md
+++ b/meta-avocado-nvidia/docs/adding-a-jetson-carrier.md
@@ -317,12 +317,24 @@ cause).
 
 - **Provisioning must run as root.** `initrd-flash` mounts the board's
   exported USB storage to write the command package and never elevates
-  on its own, so the flash needs `CAP_SYS_ADMIN`. Run it under
-  `sudo -E` so the SDK environment survives. Unprivileged runs used to
-  sign the binaries and RCM-boot the board before dying at
+  on its own, so the flash needs `CAP_SYS_ADMIN`. Unprivileged runs
+  used to sign the binaries and RCM-boot the board before dying at
   `could not mount USB storage for writing flashing commands`, which
   reads as a storage or cable fault; the provisioning script now
-  refuses up front instead.
+  refuses up front instead. Elevate with:
+
+  ```sh
+  sudo -E env "PATH=$PATH" <command>
+  ```
+
+  `sudo -E` on its own preserves the SDK environment but not `PATH`:
+  Debian, Ubuntu and Fedora all ship `Defaults secure_path` in
+  sudoers, which replaces `PATH` whatever `-E` says. The provisioning
+  script looks up its C preprocessor with `command -v` against `PATH`,
+  so under a replaced `PATH` the nativesdk and cross branches stop
+  matching - the run silently falls back to the host `cpp`, or exits
+  at `No C preprocessor (cpp) found in PATH` on a host without one.
+  The explicit `env "PATH=$PATH"` puts the caller's `PATH` back.
 - **A Jetson in Force Recovery needs a udev rule to be reachable.** The
   board enumerates as vendor `0955`, which the stock
   `51-android.rules` hands to `GROUP="adbusers" MODE="0660"`. A user
--- a/meta-avocado-nvidia/stone/tegra/stone-provision-tegraflash.sh
+++ b/meta-avocado-nvidia/stone/tegra/stone-provision-tegraflash.sh
@@ -304,19 +304,31 @@ trap "rm -rf '$temp_bin_dir'" EXIT
 # The nativesdk cpp may not be installed, so fall back to the
 # cross-compiler cpp which works fine for preprocessing (target
 # architecture is irrelevant for text preprocessing).
+#
+# Every branch stores the ABSOLUTE path, never a bare name. The wrapper written
+# below is itself called `cpp` and its directory is prepended to PATH, so a body
+# of `exec cpp "$@"` re-resolves through PATH, finds the wrapper again, and
+# re-execs itself forever. The flash then sits in "Step 1: Signing binaries"
+# with no output and no error while one process spins at 100% CPU, which reads
+# as a dead board or a bad USB cable rather than a bug in this script. Only the
+# plain-cpp fallback could collide - the nativesdk and cross names differ from
+# the wrapper's own basename - and that is precisely the branch taken whenever
+# provisioning runs outside the SDK.
 SDK_HOST_CPP=""
 _nativesdk_cpp="${AVOCADO_SDK_ARCH:-$(uname -m)}-avocadosdk-linux-cpp"
-if command -v "$_nativesdk_cpp" >/dev/null 2>&1; then
-    SDK_HOST_CPP="$_nativesdk_cpp"
+if _resolved_cpp=$(command -v "$_nativesdk_cpp" 2>/dev/null); then
+    SDK_HOST_CPP="$_resolved_cpp"
 elif [ -n "${CC:-}" ]; then
     _cc_bin="${CC%% *}"
     _cross_cpp="${_cc_bin/gcc/cpp}"
-    if command -v "$_cross_cpp" >/dev/null 2>&1; then
-        SDK_HOST_CPP="$_cross_cpp"
+    if _resolved_cpp=$(command -v "$_cross_cpp" 2>/dev/null); then
+        SDK_HOST_CPP="$_resolved_cpp"
     fi
 fi
-if [ -z "$SDK_HOST_CPP" ] && command -v cpp >/dev/null 2>&1; then
-    SDK_HOST_CPP="cpp"
+if [ -z "$SDK_HOST_CPP" ]; then
+    if _resolved_cpp=$(command -v cpp 2>/dev/null); then
+        SDK_HOST_CPP="$_resolved_cpp"
+    fi
 fi
 if [ -z "$SDK_HOST_CPP" ]; then
     echo "ERROR: No C preprocessor (cpp) found in PATH"
@@ -324,6 +336,18 @@ if [ -z "$SDK_HOST_CPP" ]; then
     echo "Ensure the SDK toolchain is properly configured."
     exit 1
 fi
+# Refuse to write a wrapper that could exec itself. command -v returns a bare
+# word for a shell builtin or an alias, and the resulting wrapper would recurse
+# exactly as described above; failing here costs one clear message instead of an
+# unexplained hang partway through a flash.
+case "$SDK_HOST_CPP" in
+    /*) ;;
+    *)
+        echo "ERROR: resolved C preprocessor '$SDK_HOST_CPP' is not an absolute path."
+        echo "The generated cpp wrapper would re-exec itself and hang the flash."
+        exit 1
+        ;;
+esac
 
 cat > "$temp_bin_dir/cpp" << CPPWRAPPER
 #!/bin/bash
--- a/meta-avocado-nvidia/stone/tegra/stone-provision-tegraflash.sh
+++ b/meta-avocado-nvidia/stone/tegra/stone-provision-tegraflash.sh
@@ -63,6 +63,11 @@ bpmp_dtb_file=$(jq -r '.storage_devices.rootdisk.images.bpmp_dtb // empty' "$AVO
 tegraflash_overlays=$(jq -r '.storage_devices.rootdisk.tegraflash.overlays // [] | join(",")' "$AVOCADO_STONE_MANIFEST")
 tegraflash_dce_overlays=$(jq -r '.storage_devices.rootdisk.tegraflash.dce_overlays // [] | join(",")' "$AVOCADO_STONE_MANIFEST")
 
+# Block device the SD card enumerates as on this module - a board property, so
+# the board's manifest is where it is declared. Consumed by the sd) branch of
+# the boot-media case below.
+tegraflash_sd_device=$(jq -r '.storage_devices.rootdisk.tegraflash.sd_device // empty' "$AVOCADO_STONE_MANIFEST")
+
 echo "Runtime images from manifest:"
 echo "  rootfs: $rootfs_file"
 echo "  var: $var_file"
@@ -79,6 +84,7 @@ echo "  dtb: $dtb_file"
 echo "  bpmp_dtb: $bpmp_dtb_file"
 echo "  overlays: $tegraflash_overlays"
 echo "  dce_overlays: $tegraflash_dce_overlays"
+echo "  sd_device: ${tegraflash_sd_device:-<not declared>}"
 
 # Create build directory for tegraflash working area
 build_dir="${AVOCADO_STONE_BUILD_DIR}/tegraflash"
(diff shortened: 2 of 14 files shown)


meta-avocado-nvidia/stone/tegra/stone-provision-tegraflash.sh
```diff
@@ -463,9 +501,15 @@ case "$boot_media" in
         fi
         ;;
     sd)
+        # An eMMC-less module instantiates only the SD controller, so the card
+        # is the first MMC device: 3400000.mmc -> mmc0 -> mmcblk0. A module that
+        # carries eMMC puts eMMC on mmc0 and the card on mmc1, which is what the
+        # BSP's generic/cfg/flash_t234_qspi_sd*.xml assume; set
+        # AVOCADO_PROVISION_SD_DEVICE=mmcblk1 for those.
+        sd_device="${AVOCADO_PROVISION_SD_DEVICE:-mmcblk0}"
```
nicksinas: **Major** — this default points eMMC-bearing modules at their eMMC.

Five manifests offer a `tegraflash-sd` profile: `stone-jetson-orin-nano-devkit.json`, `stone-jetson-agx-orin-devkit.json`, `stone-jetson-orin-nx.json`, `stone-recomputer-j301x.json`, `stone-recomputer-j401x.json`. Only the Orin Nano devkit was verified. The AGX Orin devkit's module carries eMMC, so by this PR's own reasoning its card is `mmcblk1` and `mmcblk0` is the eMMC.

The collision is visible a few lines up in this same `case`: the `emmc)` branch writes exactly `BOOTDEV="mmcblk0p1"` / `ROOTFS_DEVICE="mmcblk0"`. So on such a module `tegraflash-sd` and `tegraflash-emmc` now produce identical device names, and the SD profile writes at the internal storage.

Nothing in the tree sets `AVOCADO_PROVISION_SD_DEVICE` — I grepped; the only hits are this line, the comment above it, and `tests/test-sd-device-name.sh`. So the correct value rests entirely on the operator remembering it, per board. Set it where the board is described instead: the stone manifests already carry a `provision.envs` block, and there's `conf/machine/include/avocado-jetson.inc`. As it stands this swaps a hardcode that broke one board for a default that breaks another, which is the outcome a0fd8fd's message argues against.

meta-avocado-nvidia/stone/tegra/stone-provision-tegraflash.sh
```diff
@@ -463,9 +501,15 @@ case "$boot_media" in
...
         sed -i \
-            -e 's/^BOOTDEV=.*/BOOTDEV="mmcblk1p1"/' \
-            -e 's/^ROOTFS_DEVICE=.*/ROOTFS_DEVICE="mmcblk1"/' \
+            -e "s/^BOOTDEV=.*/BOOTDEV=\"${sd_device}p1\"/" \
+            -e "s/^ROOTFS_DEVICE=.*/ROOTFS_DEVICE=\"${sd_device}\"/" \
```
nicksinas, line 553: **Medium** — the override goes into `sed` replacement text unvalidated.

`AVOCADO_PROVISION_SD_DEVICE=/dev/mmcblk1` is the natural mistake — every other device reference an operator sees is a `/dev` path — and it turns this into `s/^BOOTDEV=.*/BOOTDEV="/dev/mmcblk1p1"/`, i.e. `sed: -e expression #1: unknown option to 's'`. A value containing `&` would silently substitute the matched text instead.

That's a new confusing failure in a PR about removing them. Validate against `^mmcblk[0-9]+$` and reject with a message naming the variable and the expected form.

meta-avocado-nvidia/docs/adding-a-jetson-carrier.md
```diff
@@ -315,6 +315,29 @@ cause).
 
 ## Caveats
 
+- **Provisioning must run as root.** `initrd-flash` mounts the board's
```
nicksinas, line 318: **Medium** — the variable this PR introduces isn't documented.

The Caveats list gains the root requirement and the udev prerequisite, but `AVOCADO_PROVISION_SD_DEVICE` appears only in a code comment and the PR description. It's the difference between flashing the SD card and flashing the eMMC on an AGX Orin, so it belongs here alongside the other two — which modules need it, what value, and why the BSP's `flash_t234_qspi_sd*.xml` says `mmcblk1p1`.

This is the doc a carrier author reads; a knob that exists only in the commit history isn't discoverable from it.

nicksinas (commented):
Reviewed against a local checkout (head 595c274, base `scarthgap` @ 670c668). No hardware, no build.

All five fixes are correct as far as they go, and I checked the parts that could have been subtly wrong: every branch of the cpp resolution stores an absolute path and `export PATH="$temp_bin_dir:..."` happens only *after* the wrapper is written, so the recursion is genuinely closed; the root check does land ahead of the first `set -u` dereference as claimed; and all three `boot_media` branches reach `initrd-flash`, so the privilege requirement isn't over-broad. The udev rule is right too — a `60-` rule runs before `73-seat-late.rules`, so the `uaccess` tag gets its ACL.

9 comments, tagged **Major** / **Medium** / **Low**. The substantive one is #1: four of the five boards that offer `tegraflash-sd` weren't in the verification, and on the eMMC-bearing one the new default names the eMMC.

Two smaller things I left off: `test-export-timeout-message.sh:19` cites `initrd-flash.sh:937-945` by line number in a comment, which will rot (naming the function would survive); and it's worth knowing whether `pre-commit` passes here, since the hook is `shfmt -i 2` while both touched scripts are 4-space throughout — neither workflow runs it, so this may be local-only by design.

meta-avocado-nvidia/stone/tegra/tests/test-sd-device-name.sh
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env bash
```
nicksinas, line 1: **Medium** — four test suites with no way to run them.

Nothing references any of the four: no runner, no aggregator, and neither `.github/workflows/pr-build-labeled.yml` nor `pr-parse-container-sdk.yml` invokes them. The repo already has the pattern — `scripts/run-feed-validation.sh` driving `scripts/feed-validation-cases` — so there's a shape to copy.

As it stands, "each verified to fail against the pre-fix code" is a one-time manual check, and these five failure modes (each of which cost a serial-console or process-tree debugging session) regress unnoticed. A `run-tegra-tests.sh` that execs the four and sums exit codes is most of the work; wiring it into a workflow is the rest.

The tests themselves are good — lifting the `sd)` branch and `wait_for_exported_storage` out and executing them, rather than asserting on source text, is the right call.

meta-avocado-nvidia/stone/tegra/tests/test-preflight-checks.sh
```diff
@@ -0,0 +1,73 @@
...
+  exit 1
+}
+
+echo "test-preflight-checks: $TARGET"
+
+if [ "$(id -u)" -eq 0 ]; then
+  echo "  skip  running as root; cannot exercise the non-root path" >&2
+  echo "test-preflight-checks: PASS"
+  exit 0
+fi
```
nicksinas, line 43: **Medium** — this reports PASS while asserting nothing.

The skip notice goes to stderr, then the suite prints `test-preflight-checks: PASS` to stdout and exits 0. Two likely environments hit this: a root CI container, and anyone running the suite under the same `sudo` the script now requires.

So the one test covering 0af1a15 reads as a passing regression guard precisely when it didn't run. Make skip a distinct outcome — exit 77, or a `SKIP` line a runner counts separately — so it can't be mistaken for coverage.

meta-avocado-nvidia/docs/adding-a-jetson-carrier.md
```diff
@@ -315,6 +315,29 @@ cause).
...
+  exported USB storage to write the command package and never elevates
+  on its own, so the flash needs `CAP_SYS_ADMIN`. Run it under
+  `sudo -E` so the SDK environment survives. Unprivileged runs used to
```
nicksinas: **Medium** — `sudo -E` doesn't preserve PATH where `secure_path` is set.

Debian, Ubuntu and Fedora all ship `Defaults secure_path` in sudoers, which replaces PATH regardless of `-E`. The nativesdk-cpp lookup is a `command -v` against PATH, so under the remedy documented here that branch silently never matches and the run falls through to plain `cpp` — or exits at "No C preprocessor (cpp) found in PATH" on a host without one.

Worth checking what actually happened on the P3767-0005 host: if PATH survived there, the sudoers config is doing the work and that's worth a sentence; if it didn't, `sudo -E env "PATH=$PATH" <command>` is the form to document. Either way the interaction is that the new root requirement puts every run through sudo, which is what makes this reachable.

meta-avocado-nvidia/stone/tegra/stone-provision-tegraflash.sh
```diff
@@ -304,26 +318,50 @@ trap "rm -rf '$temp_bin_dir'" EXIT
...
+case "$SDK_HOST_CPP" in
+    /*) ;;
+    *)
+        echo "ERROR: resolved C preprocessor '$SDK_HOST_CPP' is not an absolute path."
+        echo "The generated cpp wrapper would re-exec itself and hang the flash."
+        exit 1
+        ;;
+esac
 
 cat > "$temp_bin_dir/cpp" << CPPWRAPPER
```
nicksinas, line 383: **Low** — the generated wrapper interpolates the path unquoted.

Line 368 is `exec ${SDK_HOST_CPP} "$@"`, which was safe while the value was a bare command name. Now that every branch stores an absolute path, an SDK unpacked under a directory containing a space makes the wrapper try to exec its first word — and the failure surfaces inside tegraflash's DTS preprocessing, not here.

`exec "${SDK_HOST_CPP}" "$@"` closes it. Same change dad1e04 already made necessary, one line further down.

meta-avocado-nvidia/stone/tegra/tests/test-cpp-wrapper.sh
```diff
@@ -0,0 +1,91 @@
...
+}
+
+echo "test-cpp-wrapper: $TARGET"
+
+# --- 1. the script must never put a bare name in the wrapper -----------------
+# Matches an assignment of a literal name that is neither absolute (/...) nor a
+# variable ($...). The empty initialiser SDK_HOST_CPP="" is excluded by
+# requiring a character other than the closing quote. The historical bug was
+# exactly SDK_HOST_CPP="cpp".
+if grep -qE 'SDK_HOST_CPP="[^/$"]' "$TARGET"; then
```
nicksinas, line 45: **Low** — the lint is narrower than the surrounding comment implies.

`SDK_HOST_CPP="[^/$"]` catches the historical `SDK_HOST_CPP="cpp"`, but not `SDK_HOST_CPP='cpp'` and not an assignment through an intermediate variable that holds a bare name — which is the shape the fixed code now uses (`SDK_HOST_CPP="$_resolved_cpp"`).

The absolute-path-guard assertion below is what actually covers that case, so the pair is sound in aggregate. Worth saying so in the comment, since "SDK_HOST_CPP is only ever assigned a resolved path" reads as a stronger guarantee than the grep can give.

meta-avocado-nvidia/stone/tegra/tests/test-cpp-wrapper.sh
```diff
@@ -0,0 +1,91 @@
...
+trap 'rm -rf "$tmp_bad" "$tmp_good"' EXIT
+
+real_cpp=$(command -v cpp 2>/dev/null || true)
+if [ -z "$real_cpp" ]; then
+  echo "  skip  no host cpp available; cannot exercise the wrapper" >&2
+else
+  # bare name, wrapper dir first on PATH -> must NOT terminate
+  printf '#!/bin/bash\nexec cpp "$@"\n' >"$tmp_bad/cpp"
+  chmod +x "$tmp_bad/cpp"
+  if PATH="$tmp_bad:$PATH" timeout 5 "$tmp_bad/cpp" --version >/dev/null 2>&1; then
```
nicksinas: **Low** — "non-zero exit" is doing a lot of work here.

This passes on any failure, not only on the recursion: `timeout` absent, `cpp` rejecting `--version`, the wrapper not being executable. `test-export-timeout-message.sh:73` ("the handler still fails on timeout") has the same shape — a failed function extraction or an unbound variable in the stub harness satisfies it too.

Both would be stronger asserting the specific signal: exit 124 from `timeout` here, and the timeout string in the output there. The output greps that follow do most of the real checking, so this is about the two that don't.

meta-avocado-nvidia/stone/tegra/stone-provision-tegraflash.sh
jetm: ## PRRT_kwDOOIx8Rs6aT1-1 - SD default points eMMC modules at their eMMC

You're right, and the collision with the `emmc)` branch a few lines up makes it
concrete: on a module with eMMC the two profiles would have written the same
device names. Fixed in 9bba6fa by taking the value out of the script entirely.

The board declares it: `storage_devices.rootdisk.tegraflash.sd_device`, read
with the same jq pattern as `tegraflash.overlays` right above it.
`AVOCADO_PROVISION_SD_DEVICE` stays as a one-off override for a board revision
that enumerates differently, and if neither is set the SD profile now stops
with a message naming both places rather than guessing.

On `provision.envs`, which you suggested first: I checked avocado-cli, and
those blocks are read only by `commands/profiles.rs`, which walks them for
`${VAR}` placeholders to build the form-field list. Nothing exports a block's
values as env vars for the provision script, so a literal there would never
reach it. That's what pushed the declaration into the manifest key the script
already reads.

Values in tree, and how far each is trusted (the same table is now in the
Caveats section):

| Board | `sd_device` | Basis |
|---|---|---|
| `jetson-orin-nano-devkit` | `mmcblk0` | Confirmed on a P3767-0005 |
| `recomputer-j301x` | `mmcblk0` | Same Orin Nano module - its machine conf requires `orin-nano.inc`, whose upstream `TNSPEC_BOOTDEV_DEFAULT` is `mmcblk0p1`. Not flashed |
| `jetson-agx-orin-devkit` | `mmcblk1` | The pre-PR value, kept. Not flashed |
| `jetson-orin-nx` | `mmcblk1` | Same |
| `recomputer-j401x` | `mmcblk1` | Same |

So no board's behaviour changes except the Orin Nano devkit, which was proven
broken, and `recomputer-j301x`, which is the same module. If you know the
AGX/NX values from a flash, say so and I'll mark them confirmed.

meta-avocado-nvidia/stone/tegra/stone-provision-tegraflash.sh
jetm: ## PRRT_kwDOOIx8Rs6aT1-6 - override reaches sed's replacement text unvalidated

Fixed in 9bba6fa. The value has to match `^mmcblk[0-9]+$` - the same shape
`initrd-flash.sh:895` matches when it decides a device is external storage -
and anything else exits with a message naming the variable, the manifest key
and the expected form.

Both failure modes you named are now cases in `test-sd-device-name.sh`, along
with a partition name, and each asserts that `.env.initrd-flash` is left
untouched. Removing the guard makes them fail exactly as described: `mmcblk0p1`
writes `BOOTDEV="mmcblk0p1p1"`, and `mmcblk0&` writes
`BOOTDEV="mmcblk0BOOTDEV="placeholder"p1"`.

meta-avocado-nvidia/docs/adding-a-jetson-carrier.md
jetm: ## PRRT_kwDOOIx8Rs6aT1-8 - the variable isn't documented

Fixed in 9bba6fa. The Caveats list gains an "SD boot needs the card's device
name declared per board" entry covering both the manifest key and the env
override, why `mmcblk0` and `mmcblk1` differ by module, why the BSP's
`flash_t234_qspi_sd*.xml` say `mmcblk1p1`, the accepted value form, and the
per-board table above with what is confirmed on hardware and what isn't. It
also says how to check on a board you haven't flashed - the target's own
`mmcblkN: mmcN:0001 USD ...` console line.

meta-avocado-nvidia/stone/tegra/tests/test-sd-device-name.sh
jetm: ## PRRT_kwDOOIx8Rs6aT1_A - four suites with no way to run them

Fixed in 0924bed. `stone/tegra/tests/run-tegra-tests.sh` execs every
`test-*.sh` beside it and reports passed/skipped/failed, following the shape
`scripts/run-feed-validation.sh` already sets, and
`.github/workflows/pr-tegra-tests.yml` runs it on every PR (ubuntu-latest, no
board, no BSP, no SDK - the suites lift the code under test out of the scripts
and execute it).

No path filter, matching `pr-parse-container-sdk.yml`: the job takes seconds
and a filtered required check that never runs is its own problem. Say the word
if you'd rather scope it to `meta-avocado-nvidia/**`.

Thanks for the note on the approach - the extraction is the part that makes
these runnable in CI at all.

meta-avocado-nvidia/stone/tegra/tests/test-preflight-checks.sh
jetm: ## PRRT_kwDOOIx8Rs6aT1_D - reports PASS while asserting nothing

Fixed in 0924bed. The root path now prints `test-preflight-checks: SKIP` and
exits 77, which the runner counts in its own column and never as a pass.
Verified under `unshare -r`: exit 77, and the runner reports
`3 passed, 1 skipped, 0 failed`.

`test-cpp-wrapper.sh` had a milder version of the same problem - a host with no
`cpp` skips two of its five checks and still printed a bare PASS. Its summary
line now says which checks were skipped (7f53532). It stays exit 0, since the
source-text checks did run.

meta-avocado-nvidia/docs/adding-a-jetson-carrier.md
jetm: ## PRRT_kwDOOIx8Rs6aT1_F - `sudo -E` doesn't preserve PATH under secure_path

Fixed in 77de70d, taking the second of your two options.

I can't reconstruct what PATH did on the P3767-0005 host - the flash predates
this branch and I don't have the shell history - so I didn't want to write "it
survived there" into the docs on a guess. The documented form is now
`sudo -E env "PATH=$PATH" <command>` in both the Caveats entry and the script's
own error message, with a sentence on why the shorter form isn't enough, so it
is correct whether or not that particular host had secure_path set.

Your causal point is in the commit body: requiring root put every run through
sudo, which is what made this reachable.

meta-avocado-nvidia/stone/tegra/stone-provision-tegraflash.sh
jetm: ## PRRT_kwDOOIx8Rs6aT1_I - generated wrapper interpolates the path unquoted

Fixed in 7f53532: `exec "${SDK_HOST_CPP}" "$@"`. `test-cpp-wrapper.sh` now
asserts the quoted form is present, and the test fixture that models a working
wrapper quotes it too, so the fixture matches what the script generates.

meta-avocado-nvidia/stone/tegra/tests/test-cpp-wrapper.sh
jetm: ## PRRT_kwDOOIx8Rs6aT1_O - the lint is narrower than the comment implies

Agreed, and the pass message was making the same overclaim. Reworded in
7f53532: the comment now says the grep catches only the historical literal
`SDK_HOST_CPP="cpp"`, that it misses the single-quoted form and any assignment
through a variable holding a bare name, and that the runtime absolute-path
guard is what covers those - the pair is sound in aggregate, and the comment
now says so instead of implying the grep alone is the guarantee. The pass
message reads "SDK_HOST_CPP is not assigned a bare literal name".

meta-avocado-nvidia/stone/tegra/tests/test-cpp-wrapper.sh
jetm: ## PRRT_kwDOOIx8Rs6aT1_T - "non-zero exit" is doing a lot of work

Both tightened, in 7f53532 and 0924bed.

`test-cpp-wrapper.sh:69` now requires exit 124 - only `timeout`'s own kill
produces it, so a missing or non-executable wrapper (126/127) or a `cpp` that
rejects `--version` fails the check instead of passing as a reproduced
recursion. Verified: the bare-name wrapper does exit 124 here.

`test-export-timeout-message.sh:73` now requires the timeout line for the
device it asked for (`Timeout waiting for exported storage device mmcblk0`)
alongside the non-zero exit, so a failed function extraction or an unbound
variable in the stub harness no longer satisfies it.

mobileoverlord (approved):
All five fixes fail fast with the actual cause where they used to hang or mislead — the absolute-path cpp guard makes the self-exec loop structurally impossible, sd_device moves from a hardcoded wrong guess to a per-carrier manifest key with validation and a one-off override, and the timeout message now points at the target UART instead of the host's own disks. Tests per failure plus the tegra PR CI workflow. The initrd-flash.sh hunk is disjoint from #277's and composes.

PR #292, opened 2026-08-15
Deliver device-tree overlays on Jetson

jetm:
## Problem

`meta-avocado-nvidia` had no implementation of the fixed-path delivery hook the
avocado CLI runs before `stone bundle`. A runtime declaring
`device_tree_overlays` on a Jetson target could compile a `.dtbo` and then had
nowhere to put it, so the overlay went unclaimed and failed the build.

The obvious mechanism does not work. `stone-provision-tegraflash.sh:49` reads
`.storage_devices.rootdisk.tegraflash.overlays`, which looks like a delivery
key, but `:275-285` only tests `[ -f "$build_dir/$overlay" ]` and warns - it
never applies anything. The lists that do apply overlays live in flashvars
(`OVERLAY_DTB_FILE`, `BOOTCONTROL_OVERLAYS`) and are consumed by the classic
`flash.sh` path through `tegra-flash-helper.sh:196-198`. `initrd-flash.sh`, the
path Avocado uses, never reads them: it consumes binaries already signed at
Yocto build time. And there is nowhere to put loose overlays either - the
flashed partition set has no DTBO partition, and the ESP carries only
`EFI/BOOT/bootaa64.efi`.

## Solution

Merge the overlays into the base DTB with `fdtoverlay` and publish the result as
`storage_devices.rootdisk.images.dtb`. The provisioning script reads that key at
`:43` and copies it into place during the flash.

Because `images.dtb` is a scalar rather than a string array, the
`deep_merge_json` array-replacement problem does not arise and no
`overlays_append` primitive is needed.

Merging into the base DTB means reading it, which is why the first commit stages
the Tegra BSP into the SDK stone dir - see "Staging the Tegra BSP" below.

## Key changes

- `avocado-sdk-target.bbappend` stages `tegraflash-bsp` into the SDK stone dir,
  so the base DTB is resolvable at CLI build time at all
- Hook merges staged `.dtbo` files into the base DTB in declared order and
  publishes `images.dtb`
- The base DTB is selected by `DTBFILE` from the BSP's `.env.initrd-flash`, the
  same selector `initrd-flash.sh:922` uses
- The merged tree is published **un-prefixed**, never as `kernel_<name>.dtb`
- `fdtoverlay` failure, a missing base DTB, a missing `rootdisk`, a missing
  `DTBFILE`, and a DTB that would shadow the merged one are all hard errors
- `tegraflash.overlays` is populated as provenance only, which `:275-285`
  verifies
- One-line `SDK_SYSROOT_DEPENDS` append, matching meta-avocado-raspberrypi and
  meta-avocado-qemu
- 15-check suite that decompiles the published DTB to prove the merge happened

## Reviewer notes

Every failure mode here is silent by construction: publishing the unmerged base
DTB yields a board that boots perfectly with none of the requested overlays
applied, while the build succeeds and the overlay is marked claimed. That is why
`fdtoverlay` failures are fatal rather than falling back, and why the suite
decompiles the DTB rather than asserting on the emitted fragment - removing the
`fdtoverlay` call turns three checks red.

`rootdisk` is hardcoded to match the consumer's own hardcoded read; a discovered
destination would be invisible to it. A drift check asserts both sides still
name the same jq path.

Changing an overlay changes the DTB, so it requires a reflash rather than an
ordinary image update.

### Staging the Tegra BSP (commit 1, previously #293)

avocado-cli adds `$AVOCADO_SDK_PREFIX/stone` as a stone input dir so a BSP's
`stone-<arch>.json` can reference boot artifacts the runtime build does not
produce. Its own comment in `runtime/build.rs` describes that as half a change:

> This is the consumer half of a two-part change and is inert without the
> other: today meta-avocado's `avocado-sdk-target` installs only the stone JSON
> into this directory, so nothing new resolves from it yet.

Confirmed at `avocado-sdk-target.bb:33-38`, whose `do_install` stages exactly
three files. The overlay hook is the first consumer: without this it fails with
`tegraflash_bsp directory not found`, after the SDK wrapper has already compiled
the overlay and the hook has been discovered and run.

**This is not meta-avocado#274, and the evidence differs.** #274 is closed
because its premise was refuted: a real Raspberry Pi 5 build showed
`avocado-img-bootfiles` already ships `u-boot.bin` and `bootfiles/` via
`avocado-runtime`'s RDEPENDS, so those artifacts were reachable all along. I
checked the equivalent claim here rather than reasoning by analogy. The Orin
Nano base DTB is genuinely absent from the runtime input dir: the 24 `.dtb`
files under `runtimes/<rt>/devicetree/` are AGX Orin (`p3737`/`p3701`), and every
`p3768` entry there is a `.dtbo`. Only `tegraflash-bsp` carries
`tegra234-p3768-0000+p3767-0005-nv-super.dtb`.

Ownership is normalised to root after the copy: the deploy dir is owned by the
build user and `do_package` rejects a packaged path whose uid has no passwd
entry (`getpwuid(): uid not found`). An absent BSP is a `bbfatal`, not a silent
skip.

**The cost is real and worth a decision rather than a nod: 808 files, including
the flashing binaries, landing in every Jetson SDK image.** The whole BSP is
staged rather than only the DTB the overlay path needs, because the same
missing-input gap is what stops a manifest naming `u-boot.bin` or `bootfiles/`
from resolving, and solving it once avoids a second round when the next consumer
appears. If that trade is unacceptable, the narrow alternative is staging only
`${DTBFILE}` and its `kernel_` alias, which unblocks overlays and leaves the
general gap open. Say so on review and I will narrow it.

### Two commits that came from hardware, not from review

The last two commits exist because the earlier ones were not enough, and both
failures were invisible to every build-path assertion.

**`select by DTBFILE, not by uniqueness`.** The hook originally identified the
base DTB as the only `kernel_*.dtb` present. A real BSP ships one per module SKU
- five on an Orin Nano dev kit (p3767-0000/0001/0003/0004/0005) - so the
uniqueness rule refused every real build. Selection now reads `DTBFILE` from
`.env.initrd-flash`, which is what the flash itself uses, so the hook and the
flash cannot disagree about which SKU is being built.

**`publish the un-prefixed base DTB`.** The merged DTB was being published as
`kernel_<name>.dtb`. `tegra-flash-helper.sh:567-572` does `rm -f kernel_$base`
and recreates that file from the un-prefixed `$base` during signing, so the
merged tree was written to the one filename guaranteed to be deleted before it
was read. **Reviewers should treat the un-prefixed publication as load-bearing
rather than stylistic.** Under the prefixed form the overlay compiled, the right
SKU was selected, the merge succeeded, the fragment named a real file, the
bundle finalized, and the flash reported `Final status: SUCCESS` - and the board
applied nothing. The only thing that caught it was reading `/proc/device-tree`
on the running kernel.

### Verification

Verified on hardware, not by the test suite alone. On a Jetson Orin Nano
Developer Kit (P3767-0005) flashed via the tegraflash-sd path:

    # tr -d '\0' < /proc/device-tree/hello-overlay/avocado,marker
    eng-2134-dto-e2e

and absent on a paired build with the `device_tree_overlays` declaration
removed and nothing else changed.

The env-var contract the hook implements (`AVOCADO_DTBO_STAGING`,
`AVOCADO_OVERLAYS_MANIFEST`, `AVOCADO_DELIVERY_FRAGMENT`,
`AVOCADO_STONE_MANIFEST`) is the one avocado-cli#183 emits.

Branched fresh from `scarthgap` rather than continuing
`eng-2134-devicetree-overlay`, since #245 was squash-merged and that branch's
earlier commits duplicate upstream content without sharing ancestry.

Supersedes #293, whose commit is now the first here. It was split out
originally, but a 34-line change with no independently observable outcome is not
a separate review - merging it alone changes nothing, and merging this without
it produces a hook that cannot find its input.

### DIFF
--- a/meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
+++ b/meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
@@ -104,12 +104,19 @@ def find_base_dtb(stone, data_dir):
     if not os.path.isdir(bsp_dir):
         die(f"tegraflash_bsp directory not found: {bsp_dir}")
 
-    base_name = f"kernel_{read_dtbfile(bsp_dir)}"
+    # Un-prefixed, deliberately. The BSP ships both <name>.dtb and the
+    # kernel_<name>.dtb alias, but they are not interchangeable here:
+    # tegra-flash-helper.sh:567-572 does `rm -f kernel_<name>.dtb` and recreates
+    # it from the un-prefixed <name>.dtb (or its .signed form) at signing time.
+    # Publishing the merged tree as kernel_<name>.dtb therefore hands it to a
+    # file that is deleted before it is read, and the board boots the stock tree
+    # with no overlay applied and nothing reported. Merging into the un-prefixed
+    # base means the signer regenerates the alias FROM the merged tree.
+    base_name = read_dtbfile(bsp_dir)
     if not os.path.isfile(os.path.join(bsp_dir, base_name)):
         die(
             f"{base_name} not found in {bsp_dir}: DTBFILE names it as the tree "
-            "this machine boots, but the BSP does not ship it. tegraflash-bsp.bb "
-            "only creates the kernel_ alias for DTBs matching nv-super"
+            "this machine boots, but the BSP does not ship it"
         )
     return bsp_dir, base_name
 
--- a/meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
+++ b/meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
@@ -57,6 +57,31 @@ def env(name):
     return value
 
 
+def read_dtbfile(bsp_dir):
+    """Return the DTB basename this machine actually boots.
+
+    A BSP ships one base DTB per module SKU - five on the Orin Nano developer
+    kit (p3767-0000/0001/0003/0004/0005) - so the tree cannot be identified by
+    being the only one present. initrd-flash.sh:922 resolves the flash XML's
+    DTB_FILE to kernel_$DTBFILE, sourcing DTBFILE from .env.initrd-flash, which
+    is the only thing in the BSP that names the SKU this machine boots. Use the
+    same selector, so the merged tree is by construction the tree that gets
+    flashed."""
+    env_path = os.path.join(bsp_dir, ".env.initrd-flash")
+    if not os.path.isfile(env_path):
+        die(
+            f"{env_path} not found: it names the DTB this machine boots "
+            "(DTBFILE), and a BSP ships one base DTB per module SKU, so there "
+            "is no safe way to pick one without it"
+        )
+    with open(env_path) as f:
+        for line in f:
+            key, sep, value = line.partition("=")
+            if sep and key.strip() == "DTBFILE":
+                return value.strip().strip('"').strip("'")
+    die(f"no DTBFILE entry in {env_path}; cannot tell which base DTB to merge into")
+
+
 def find_base_dtb(stone, data_dir):
     """Locate the BSP's base kernel DTB.
 
@@ -79,20 +104,14 @@ def find_base_dtb(stone, data_dir):
     if not os.path.isdir(bsp_dir):
         die(f"tegraflash_bsp directory not found: {bsp_dir}")
 
-    candidates = sorted(
-        name
-        for name in os.listdir(bsp_dir)
-        if name.startswith("kernel_") and name.endswith(".dtb")
-    )
-    if not candidates:
-        die(f"no base DTB (kernel_*.dtb) in {bsp_dir}")
-    if len(candidates) > 1:
+    base_name = f"kernel_{read_dtbfile(bsp_dir)}"
+    if not os.path.isfile(os.path.join(bsp_dir, base_name)):
         die(
-            "more than one base DTB in "
-            f"{bsp_dir}: {', '.join(candidates)}. Refusing to guess which tree "
-            "the overlays belong to"
+            f"{base_name} not found in {bsp_dir}: DTBFILE names it as the tree "
+            "this machine boots, but the BSP does not ship it. tegraflash-bsp.bb "
+            "only creates the kernel_ alias for DTBs matching nv-super"
         )
-    return bsp_dir, candidates[0]
+    return bsp_dir, base_name
 
 
 def main():
--- a/meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
+++ b/meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
@@ -129,6 +129,25 @@ def main():
     # derives kernel_<basename>.dtb from it, and the flash layout expects the
     # name the BSP shipped.
     merged = os.path.join(staging, base_name)
+
+    # images.dtb is resolved by name against stone's input dirs, first match
+    # wins, and the merged DTB has to keep the base name because the flash
+    # layout refers to the BSP's filename. Today the base DTB exists only
+    # nested inside tegraflash_bsp/, so the staged copy is the only top-level
+    # match and it wins. If anything ever deploys that name beside the manifest,
+    # stone resolves to the unmerged tree instead and the board boots cleanly
+    # with no overlay applied - the exact failure this hook exists to prevent,
+    # and one that reports nothing. The Raspberry Pi hook refuses by name for
+    # the same class of shadowing.
+    shadow = os.path.join(data_dir, base_name)
+    if os.path.exists(shadow):
+        die(
+            f"{shadow} would shadow the merged DTB: stone resolves images.dtb "
+            "by name against its input dirs and would pick that copy over the "
+            "merged one in staging, shipping a tree with no overlays applied. "
+            "Remove it or rename the deployed DTB"
+        )
+
     shutil.copy2(os.path.join(bsp_dir, base_name), merged)
 
     # fdtoverlay applies in argument order, and order is load-bearing: a later
--- a/meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
+++ b/meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
@@ -76,9 +76,38 @@ def read_dtbfile(bsp_dir):
         )
     with open(env_path) as f:
         for line in f:
+            line = line.strip()
+            if not line or line.startswith("#"):
+                continue
             key, sep, value = line.partition("=")
-            if sep and key.strip() == "DTBFILE":
-                return value.strip().strip('"').strip("'")
+            if not sep:
+                continue
+            # `export DTBFILE=...` and a trailing `# comment` are both valid
+            # shell and both appear in hand-edited env files. Neither is in the
+            # BSP's current output, but this is the single selector the whole
+            # merge hinges on: an unrecognised form matches nothing and falls
+            # through to the die() below, which reports a MISSING selector when
+            # the truth is an unparsed one - and sends the reader looking for a
+            # BSP that failed to write the value rather than a format change.
+            key = key.strip()
+            if key.startswith("export "):
+                key = key[len("export ") :].strip()
+            if key != "DTBFILE":
+                continue
+            value = value.strip()
+            if value[:1] in ("'", '"'):
+                quote = value[0]
+                end = value.find(quote, 1)
+                if end != -1:
+                    return value[1:end]
+                # Unterminated quote - fall through to the bare-word path
+                # rather than silently returning the rest of the line.
+                value = value[1:]
+            else:
+                value = value.split("#", 1)[0].strip()
+            if value:
+                return value
+            die(f"DTBFILE in {env_path} is set but empty; cannot pick a base DTB")
     die(f"no DTBFILE entry in {env_path}; cannot tell which base DTB to merge into")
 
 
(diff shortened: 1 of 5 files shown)


meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
```diff
@@ -0,0 +1,225 @@
...
+    # wins, and the merged DTB has to keep the base name because the flash
+    # layout refers to the BSP's filename. Today the base DTB exists only
+    # nested inside tegraflash_bsp/, so the staged copy is the only top-level
+    # match and it wins. If anything ever deploys that name beside the manifest,
+    # stone resolves to the unmerged tree instead and the board boots cleanly
+    # with no overlay applied - the exact failure this hook exists to prevent,
+    # and one that reports nothing. The Raspberry Pi hook refuses by name for
+    # the same class of shadowing.
+    shadow = os.path.join(data_dir, base_name)
+    if os.path.exists(shadow):
```
nicksinas, line 222: The shadow check covers `data_dir` only, but `images.dtb` is published by **basename** and — per the comment at :199-207 — stone resolves it across *all* input dirs, first match wins. If that basename appears top-level in another input dir stone orders before the staging dir (the SDK stone dir where the BSP itself is staged is the obvious candidate), the unmerged tree ships, which is exactly the silent no-overlay boot this hook exists to prevent — and this guard won't catch it.

Can we confirm the staging dir is guaranteed to win over every other input dir, or should the guard walk the same input-dir set stone does rather than just `data_dir`? As written, a top-level collision in any dir stone orders first reintroduces the failure the whole hook is built to stop.

meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
```diff
@@ -0,0 +1,225 @@
...
+            "by name against its input dirs and would pick that copy over the "
+            "merged one in staging, shipping a tree with no overlays applied. "
+            "Remove it or rename the deployed DTB"
+        )
+
+    shutil.copy2(os.path.join(bsp_dir, base_name), merged)
+
+    # fdtoverlay applies in argument order, and order is load-bearing: a later
+    # overlay may target a node an earlier one created.
+    dtbo_paths = [os.path.join(staging, ov["file"]) for ov in overlays]
```
nicksinas, line 234: Every other failure here goes through `die()` with a clean `[ERROR] device-tree-overlay-deliver` prefix, but `ov["file"]` / `ov["name"]` (also :204, :217) raises a raw `KeyError` traceback on a malformed manifest entry — the one class of bad input that escapes the otherwise-uniform fail-clean handling.

Consider validating each overlay entry has `file`/`name` up front (or `ov.get("file")` + `die`) so a malformed manifest fails with the same clean message as everything else rather than a bare traceback.

meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
```diff
@@ -0,0 +1,225 @@
...
+    env_path = os.path.join(bsp_dir, ".env.initrd-flash")
+    if not os.path.isfile(env_path):
+        die(
+            f"{env_path} not found: it names the DTB this machine boots "
+            "(DTBFILE), and a BSP ships one base DTB per module SKU, so there "
+            "is no safe way to pick one without it"
+        )
+    with open(env_path) as f:
+        for line in f:
+            key, sep, value = line.partition("=")
```
nicksinas, line 82: `key, sep, value = line.partition("=")` + `key.strip() == "DTBFILE"` silently skips an `export DTBFILE=...` form and doesn't strip a trailing `# comment`. Fine for the BSP's current `.env.initrd-flash` format — but this is the single selector the whole merge hinges on, so a format drift fails by matching nothing and falling through to the "no DTBFILE entry" die, which reads as a *missing* selector rather than an *unparsed* one.

Worth either tolerating those forms or documenting the exact expected line format so a future BSP change doesn't misattribute the failure.

meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
```diff
@@ -0,0 +1,225 @@
...
+    if manifest.get("version") != 1:
+        die("overlays.manifest.json: unsupported version")
+    overlays = manifest.get("overlays", [])
+    if not overlays:
+        die("hook invoked with no overlays to deliver")
+
+    with open(stone_path) as f:
+        stone = json.load(f)
+    bsp_dir, base_name = find_base_dtb(stone, data_dir)
+    platform = stone.get("runtime", {}).get("platform", "tegra")
```
nicksinas, line 196: `platform` falls back to `"tegra"` when `runtime.platform` is absent, and that value is written as `claimed_by` at :213. If the CLI's post-hook check compares `claimed_by` against a specific platform id rather than just non-null, the fallback path would fail the claim check — harmless only if `runtime.platform` is always present (the test assumes it is).

Does the post-hook claim check require a specific value or just non-null? If the former, worth confirming `runtime.platform` is always populated, or dropping the fallback so a missing platform fails loudly here instead of downstream.

meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
jetm: You are right, and it is worse than the guard being incomplete. I traced the actual `-i` order the CLI builds:

1. the project's own `stone_include_paths`
2. `$AVOCADO_PREFIX/runtimes/$RUNTIME_NAME`
3. `$AVOCADO_SDK_PREFIX/stone`
4. `$DTBO_STAGING` — appended last, `src/utils/device_tree_overlay.rs:326`

stone takes the first match (`create.rs` `find_file_in_dirs`, `dir.join(name)`), so the directory holding the merged DTB had the **lowest** precedence of the four, and the guard was checking `data_dir` — which is #3, one of the three that outranked it. So the answer to "can we confirm the staging dir wins" is no: it lost to everything, and it only worked because nothing happened to collide.

`stone_include_paths` being first is what makes this reachable rather than theoretical. A user compiling their own DTBs into a directory they add to that list is following documented usage, and one file with the BSP's base name there ships an unmerged tree.

Fixed in avocado-cli#183 by prepending instead: the merged tree now wins by construction rather than by the absence of a collision. That is the version a guard cannot silently miss, which was your point. There is a test asserting the prepended form and rejecting the appended one, and I confirmed it fails when the order is reverted.

On walking the same input-dir set here: the hook cannot. The `-i` list is assembled after the hook returns and the hook is never told what is on it. So I have kept this check as a second line over the one directory it can see, and rewritten the comment, which previously claimed the staged copy was the only top-level match and therefore won — the premise you just disproved.

meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
jetm: Fixed. Every overlay entry is now validated once up front - each must be an object carrying a non-empty `name` and `file` - so a malformed manifest exits through `die()` with the same prefixed one-line message as everything else, before anything is copied.

Confirmed the old behaviour was what you described rather than assuming it: against the previous hook the new test produces

```
    dtbo_paths = [os.path.join(staging, ov["file"]) for ov in overlays]
                                        ~~^^^^^^^^
KeyError: 'file'
```

which is the traceback, in a build-container log, that this was the only path capable of producing. The test asserts both that the hook fails and that the output contains no `Traceback`, so a future refactor that reintroduces bare indexing trips it.

meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
jetm: Tolerating the forms rather than documenting the current one, since your reasoning about the failure mode is the part that decided it: the parser now skips blanks and comment lines, strips an `export ` prefix, handles single and double quotes, and drops a trailing `# comment` on a bare word. An empty value gets its own message instead of falling through.

Your prediction about misattribution is exactly what the old parser did. Against the previous hook the new test gives:

- `export DTBFILE="tegra234-test-0005.dtb"` → `no DTBFILE entry in .../.env.initrd-flash` — reports a **missing** selector when the line is right there
- `DTBFILE=tegra234-test-0005.dtb  # the SKU this board boots` → `tegra234-test-0005.dtb  # the SKU this board boots not found in .../tegraflash-bsp` — the comment ends up inside the filename

The second is the worse one: it sends the reader looking for a DTB the BSP failed to ship. Both now parse. The single-quoted form already worked and is covered so it stays that way.

meta-avocado-nvidia/recipes-avocado/avocado-dtc-overlay-deliver/files/device-tree-overlay-deliver
jetm: Checked the consumer rather than reasoning about it. The post-hook check compares for non-null only, not against a platform id - `avocado-cli` `src/utils/device_tree_overlay.rs:238`:

```python
unclaimed = [o.get("name") for o in data.get("overlays", []) if not o.get("claimed_by")]
```

So the fallback cannot fail the claim check; any non-empty string passes.

Leaving it in place rather than dropping it. Since the check is truthiness-only, a missing `runtime.platform` has no downstream failure to fail loudly *instead of* - dropping the fallback would add a hard error to a path that currently works, which is a cost with nothing bought. And `"tegra"` is not a guess here: this hook only ships in `meta-avocado-nvidia`, so it is the correct label for anything it claims. The value's other use is the delivery log line, where it reads correctly too.

Worth noting the coupling is one-directional and undefended: if that check ever grows to compare a specific id, this fallback becomes wrong silently. The hook cannot see that check, so the guard belongs on the CLI side if it changes.

mobileoverlord (approved):
Right mechanism: merging overlays into the base DTB at delivery time (fdtoverlay) instead of boot-time application means the flashed artifact is one signed-able blob — this composes with FIT verified boot later, which the boot-time-apply approach would not. DTB selection reuses the flash path's own DTBFILE selector and refuses when the SKU is ambiguous rather than guessing. Scalar images.dtb neatly sidesteps the deep_merge array-replacement trap.

PR #294, opened 2026-08-18
ENG-2347 + ENG-2364: freeze the report envelope, and split unscanned_recipes

hiagofranco:
I tested locally and seens good, tests are passing. Here is the Claude Opus 5 summary of the 3 commits for the AI reviewer:

```
# PR description

Freezes the CVE report envelope — keys, types, counter namespace, derived
invariants — behind a schema, a golden fixture and a mutation-proved checker,
so the two independent parsers have something to build against. Contents stay
free to move.

Then fixes a counter the freeze would otherwise have frozen wrong:
`unscanned_recipes` was counting recipes cve-check had scanned and found no NVD
record for, alongside recipes nothing had looked at. Those are now separate.

**Reviewers: one deliberate departure from the compatibility rules.** The split
changes what an existing counter counts, which `README` reserves for a new
major, and it is made at version 1 anyway. The reasoning is in the last section
below and recorded in the README. It is the one thing here a green CI run
cannot tell you is right.

Tests pass locally (59) and in CI. The recipe path is not covered — see
"Known gaps".

---

# Summary for the reviewer

Four commits:

    8548d64  meta-avocado-sbom: freeze the report contract
    54a3147  avocado-cve-report: check every report the build produces
    64e0f3a  ci: check the report contract on every meta-avocado-sbom change
    2128451  avocado-cve-report: split unscanned_recipes from no_cve_record_recipes

## Part 1 — the freeze (ENG-2347, first three commits)

The CVE report JSON has two independent parsers — `avocado-cve-match.py` and
Peridio's consumer — and neither had a contract to build against. Both
hand-rolled their own version gate against a shape nobody validated.

This freezes the **envelope**: keys, types, the counter namespace, and the
derived invariants. It does not freeze **contents** — PURLs, CVE IDs, counts,
digests are all free to move. That split is the point: ENG-2196's mapping
corrections change contents on purpose, and they must never turn CI red.

What lands:

- `schema/avocado-cve-report-v1.schema.json` — JSON Schema 2020-12, for
  consumers who can't run our Python.
- `tests/fixtures/avocado-cve-report.json` — 44 KB golden fixture, trimmed from
  a real qemuarm64 report by `make-fixture.py`. Regenerate it, don't hand-edit.
- `lib/avocado_sbom/verify.py` — the checker, shared by CI, the CLI and the
  recipe, so the three can't drift. `packages_digest` moved into `report.py`
  for the same reason.
- `tests/test_verify.py` — 52 tests. Every check is mutation-proved, and
  controls assert that CVE churn, mapping corrections, a wholly different
  report, and a build with zero unpatched CVEs all still pass.
- `.github/workflows/pr-sbom-report-contract.yml` — runs on any
  `meta-avocado-sbom/**` change.
- `do_cve_report` now checks every report the build produces.

**Two severities.** *Malformed* (shape is wrong) is always fatal — a report
that fails its own contract is not shippable. *Incomplete* (well-formed, but
missing data it should carry) is governed by `AVOCADO_CVE_REPORT_STRICT`,
default `1`. Set `0` to warn instead. The check runs after the task's existing
diagnostics, which explain what to do about it.

## Part 2 — the counter split (ENG-2364, fourth commit)

`unscanned_recipes` exists to separate "nothing looked at this recipe" from
"something looked and found nothing". It was not doing that.

A recipe entered the scanned set only if `_has_cve_record()` was true, which is
false when every product carries `cvesInRecord: "No"`. That flag is cve-check
reporting that it looked the product up in the NVD and the database holds no
record under that name — the strongest available evidence a recipe *was*
scanned. It was being read as the opposite.

Measured on qemuarm64, 393 cve-check files, 222 recipes shipping packages:

    unscanned_recipes, before                              108
      scanned, every product cvesInRecord "No"              97   <- miscounted
      no <PN>_cve.json at all                               11   <- opt-outs

The 97 include `attr`, `kmod`, `diffutils`, `libcap-ng`, `bash-completion` —
all with a cve-check result on disk. After the split, `unscanned_recipes` reads
**11**: six packagegroups (`packagegroup.bbclass` sets `CVE_PRODUCT = ""`),
four avocado meta-recipes, and `glibc-locale`.

The report now distinguishes three states, and the checker enforces that they
partition the shipped recipes: the two lists are disjoint, both are scoped to
recipes present in `packages`, and together with the clean ones they account
for every shipped recipe exactly once. On this build: 11 + 97 + 89 clean + 25
carrying CVEs = 222.

Also in this commit: entry presence now decides `scanned`, recorded before the
stale-version drop. A recipe whose cve-check entry was discarded as stale was
previously counted as unscanned as well; it was scanned, just at a version this
build did not produce.

`README`'s world-build floor of 68 and its "all of them opt-outs" claim were
measured under the earlier definition. They are dropped rather than renumbered
— a figure taken under a definition that no longer exists cannot be repaired by
changing its value. There is no replacement world-build figure yet; nobody has
run `bitbake world` on this tree since the split.

## Compatibility — read this one

The rules in `README` say adding keys or counters needs no major bump, while
removals, renames, type changes and **meaning changes** require a new major.

**This PR makes a meaning change at version 1.** `unscanned_recipes` counts
something different than it did, and both new keys are `required` rather than
optional, so a pre-split report is now rejected. Under the stated rules that is
a major bump.

It was made at version 1 because version 1 had no consumer at the time: the
report is produced by this layer and read by nothing outside it —
`avocado-cli` has no reference to it, and neither does anything else in the
tree. The alternative was to freeze a counter we already knew was reporting the
opposite of the truth, which seemed worse than spending the exemption before
anyone could be broken by it.

The exemption is recorded in `README` alongside the rules, with the reasoning
and an explicit note that the rules apply without it from here.

If you disagree, the fix is mechanical: `REPORT_VERSION`, `SUPPORTED_MAJORS`,
the schema `$id` and filename, and a fixture regeneration.

## Validation

`--strict` passes on the fixture and on a fresh qemuarm64 build. Diffing that
fresh build against the 11 Aug one is the real test — eight days of churn:

| full report      | 11 Aug | 19 Aug |
|------------------|--------|--------|
| packages         | 6224   | 6227   |
| recipes          | 50     | 51     |
| cves             | 1015   | 1074   |
| packaged_recipes | 24     | 25     |
| packaged_cves    | 847    | 906    |

Every content axis moved; the envelope is the same shape and the checker stays
green. The fixture digest moved with it, `sha256:2fd3b865` to `sha256:e56d3805`
over the same 260-package trimmed set.

**`--strict` no longer passes on the 11 Aug report, and that is intended.** It
predates the split, so it lacks `no_cve_record_recipes` and is rejected with
`missing top-level key`. Any pre-split report is. This is the backward-
incompatible half of the exemption above, stated here so it isn't mistaken for
a regression.

The schema tests that need `jsonschema` are skipped locally — it isn't
installable on either machine here — so they were checked by hand against the
fixture and a full report: both validate, an added counter is accepted, and a
removed counter, a removed top-level key and an empty recipe name are each
rejected. CI installs `jsonschema` and runs them for real.

## Known gaps

- **The recipe path is not covered by tests.** `report.py`, `verify.py`, the
  schema and the fixture all are. `do_cve_report` needs bitbake, so the new
  `bb.note` and the counter flowing through the task have been read but never
  run.
- **`do_cve_report` validates the in-memory document, not the file it wrote.**
  `write_report()` happens first, then the checks run against the same objects
  `build_report` derived them from, so the derived assertions are close to
  tautological in-process and nothing ever reads `out_path` back. A truncated
  write would pass. Pre-existing in `54a3147`, called out here rather than left
  to be found; worth a follow-up rather than a fix in this PR.
- **`avocado-cve-match.py` is still untracked**, so its own
  `SUPPORTED_REPORT_VERSIONS = ("1",)` gate can't be guarded by this CI.
  Getting it into the repo is the follow-up that closes that.

## Not in scope

- The `unscanned_recipes` ingest band and CI assertion. ENG-2364 establishes
  what the counter means and what to gate on — the opt-out count, not the
  composite, which folds in NVD coverage of the package set — but the band
  itself belongs to ENG-2363 and a later ENG-2347 follow-up.
- A world-build figure for the counter, as above.
- Dependency edges (`dependsOn`), deferred past M1 on purpose.
```

### DIFF
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -251,6 +251,46 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
     stats.recipes = len(recipes)
     return recipes, scanned, scanned - with_record
 
+OPTOUT_VERSION = "1"
+
+def read_optouts(cve_dir):
+    """Read the opt-out markers avocado-cve-optout.bbclass writes beside the
+    cve-check results.
+
+    Returns (declared, unreadable): a recipe name -> reason map, and the number
+    of markers that could not be read. A recipe with no cve-check result and no
+    marker was not scanned and nothing says why, which is the case worth
+    failing a build over; the reason string is carried so the failure can say
+    which mechanism a declared opt-out used.
+
+    Deliberately not part of build_report: the markers explain the report
+    rather than belonging to it, and the version 1 envelope is frozen.
+    """
+    declared = {}
+    unreadable = 0
+
+    for path in sorted(glob.glob(os.path.join(cve_dir, "*_optout.json"))):
+        try:
+            with open(path) as f:
+                data = json.load(f)
+        except (OSError, json.JSONDecodeError):
+            unreadable += 1
+            continue
+
+        if not isinstance(data, dict):
+            unreadable += 1
+            continue
+
+        name = data.get("name")
+        reason = data.get("reason")
+        if not isinstance(name, str) or not name or not isinstance(reason, str):
+            unreadable += 1
+            continue
+
+        declared[name] = reason
+
+    return declared, unreadable
+
 def packages_digest(packages):
     """Fingerprint of the package set. Consumers re-derive it, so the input
     format is part of the frozen contract.
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -220,6 +220,18 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
     stats.recipes = len(recipes)
     return recipes, scanned
 
+def packages_digest(packages):
+    """Fingerprint of the package set. Consumers re-derive it, so the input
+    format is part of the frozen contract.
+    """
+    digest = hashlib.sha256()
+    for name in sorted(packages):
+        pkg = packages[name]
+        digest.update(
+            ("%s\t%s\t%s\n" % (name, pkg["recipe"], pkg["version"])).encode()
+        )
+    return "sha256:" + digest.hexdigest()
+
 REPORT_VERSION = "1"
 
 def build_report(cve_dir, pkgdata_dirs, machine=None, status="Unpatched", summary=True):
@@ -238,20 +250,13 @@ def build_report(cve_dir, pkgdata_dirs, machine=None, status="Unpatched", summar
     unscanned = sorted({p["recipe"] for p in packages.values()} - scanned)
     stats.unscanned_recipes = len(unscanned)
 
-    digest = hashlib.sha256()
-    for name in sorted(packages):
-        pkg = packages[name]
-        digest.update(
-            ("%s\t%s\t%s\n" % (name, pkg["recipe"], pkg["version"])).encode()
-        )
-
     report = {
         "version": REPORT_VERSION,
         "generated": datetime.datetime.now(datetime.timezone.utc)
         .replace(microsecond=0)
         .isoformat()
         .replace("+00:00", "Z"),
-        "packages_digest": "sha256:" + digest.hexdigest(),
+        "packages_digest": packages_digest(packages),
         "status": status,
         "counts": {
             "recipes": len(recipes),
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -205,9 +205,17 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
                 break
 
             issues = []
+            seen_here = set()
             for issue in raw_issues:
                 issue_status = issue.get("status")
                 if issue_status == status:
+                    # An id repeated inside one entry is one CVE, the same as
+                    # one repeated across two entries - see the merge below.
+                    # Dropping it here covers both, and covers the first entry
+                    # for a recipe, which the merge never sees.
+                    if issue.get("id") in seen_here:
+                        continue
+                    seen_here.add(issue.get("id"))
                     issues.append({k: issue[k] for k in fields if k in issue})
                     continue
 
@@ -229,6 +237,15 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
                 # exceed what it carries fails its own check as malformed,
                 # which no setting overrides. Merge instead, by id: the same
                 # CVE reported twice is one CVE.
+                #
+                # The version already recorded stands, and the merged CVEs are
+                # carried under it. Entries at two versions only reach here
+                # unpackaged: a packaged recipe has a known version, and an
+                # entry at any other one was dropped as stale above. Nothing
+                # this build ships is described by that version, so the first
+                # is as good as the second - first meaning first in sorted
+                # filename order, then file order, which makes it stable
+                # across runs rather than correct.
                 seen_ids = {c.get("id") for c in existing["cves"]}
                 added = [c for c in issues if c.get("id") not in seen_ids]
                 existing["cves"].extend(added)
@@ -249,7 +266,14 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
                 stats.packaged_cves += len(issues)
 
     stats.recipes = len(recipes)
-    return recipes, scanned, scanned - with_record
+    # cvesInRecord "No" means no record matched, not that none exists:
+    # cve-check.bbclass skips the products of a CVE it ignored or found
+    # already patched, so at AVOCADO_CVE_REPORT_STATUS "Patched" or "Ignored" a
+    # recipe can carry issues and still be written with no record on every
+    # product. Reporting it as having none while its CVEs sit in "recipes"
+    # would put it in two legs of the partition at once, so what the report
+    # carries decides.
+    return recipes, scanned, scanned - with_record - set(recipes)
 
 def read_optouts(optout_dir):
     """Read the opt-out markers avocado-cve-optout.bbclass writes beside the
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -19,6 +19,7 @@ class Stats(dict):
         "package_collisions",
         "stale_dropped",
         "unscanned_recipes",
+        "no_cve_record_recipes",
         "unpatched_cves",
         "ignored_cves",
         "patched_cves",
@@ -130,6 +131,12 @@ def _strip_pe(version):
     return version
 
 def _has_cve_record(entry):
+    """Whether the NVD holds any CVE record for this entry's products.
+
+    False is a scan result, not the absence of one: cve-check looked the
+    product up and the database had nothing under that name. Do not use it to
+    decide whether a recipe was scanned - presence of the entry decides that.
+    """
     products = entry.get("products")
     if not isinstance(products, list) or not products:
         return True
@@ -145,6 +152,7 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
     )
     recipes = {}
     scanned = set()
+    with_record = set()
 
     paths = sorted(glob.glob(os.path.join(cve_dir, "*_cve.json")))
     stats.cve_files = len(paths)
@@ -170,6 +178,16 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
                 stats.cve_files_unreadable += 1
                 break
 
+            # Presence of an entry is what separates a recipe something
+            # examined from one nothing looked at, so this is recorded before
+            # the stale-version drop below: those recipes were scanned, just at
+            # a version this build does not have. cvesInRecord says something
+            # narrower - the NVD holds no record for the product - which is
+            # tracked separately and must not be read as unscanned.
+            scanned.add(name)
+            if _has_cve_record(entry):
+                with_record.add(name)
+
             known = recipe_versions.get(name)
             stripped = _strip_pe(version)
             if not known:
@@ -179,9 +197,6 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
                     stats.stale_dropped += 1
                     continue
                 version = stripped
-
-            if _has_cve_record(entry):
-                scanned.add(name)
 
             raw_issues = entry.get("issue", [])
             if not isinstance(raw_issues, list) or not all(
@@ -218,7 +233,7 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
                 stats.packaged_cves += len(issues)
 
     stats.recipes = len(recipes)
-    return recipes, scanned
+    return recipes, scanned, scanned - with_record
 
 def packages_digest(packages):
     """Fingerprint of the package set. Consumers re-derive it, so the input
@@ -243,12 +258,18 @@ def build_report(cve_dir, pkgdata_dirs, machine=None, status="Unpatched", summar
 
     stats = Stats()
     packages, recipe_versions = read_pkgdata(pkgdata_dirs, stats)
-    recipes, scanned = read_cve_data(
+    recipes, scanned, no_record = read_cve_data(
         cve_dir, recipe_versions, stats, status=status, summary=summary
     )
 
-    unscanned = sorted({p["recipe"] for p in packages.values()} - scanned)
+    # Both lists are scoped to recipes that shipped a package, so together with
+    # the recipes scanned and found in the NVD they partition that set: a
+    # consumer can account for every shipped recipe exactly once.
+    shipped = {p["recipe"] for p in packages.values()}
+    unscanned = sorted(shipped - scanned)
     stats.unscanned_recipes = len(unscanned)
+    no_cve_record = sorted(no_record & shipped)
+    stats.no_cve_record_recipes = len(no_cve_record)
 
     report = {
         "version": REPORT_VERSION,
@@ -267,6 +288,7 @@ def build_report(cve_dir, pkgdata_dirs, machine=None, status="Unpatched", summar
             "cve_files": stats.cve_files,
             "stale_dropped": stats.stale_dropped,
             "unscanned_recipes": stats.unscanned_recipes,
+            "no_cve_record_recipes": stats.no_cve_record_recipes,
             "cve_files_unreadable": stats.cve_files_unreadable,
             "pkgdata_unreadable": stats.pkgdata_unreadable,
             "package_collisions": stats.package_collisions,
@@ -278,6 +300,7 @@ def build_report(cve_dir, pkgdata_dirs, machine=None, status="Unpatched", summar
         "recipes": recipes,
         "packages": packages,
         "unscanned_recipes": unscanned,
+        "no_cve_record_recipes": no_cve_record,
     }
     if machine:
         report["machine"] = machine
@@ -506,6 +529,7 @@ def main():
     )
     for key in (
         "unscanned_recipes",
+        "no_cve_record_recipes",
         "stale_dropped",
         "package_collisions",
         "cve_files_unreadable",
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -251,9 +251,7 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
     stats.recipes = len(recipes)
     return recipes, scanned, scanned - with_record
 
-OPTOUT_VERSION = "1"
-
-def read_optouts(cve_dir):
+def read_optouts(optout_dir):
     """Read the opt-out markers avocado-cve-optout.bbclass writes beside the
     cve-check results.
 
@@ -269,7 +267,7 @@ def read_optouts(cve_dir):
     declared = {}
     unreadable = 0
 
-    for path in sorted(glob.glob(os.path.join(cve_dir, "*_optout.json"))):
+    for path in sorted(glob.glob(os.path.join(optout_dir, "*_optout.json"))):
         try:
             with open(path) as f:
                 data = json.load(f)
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -221,6 +221,22 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
                 continue
 
             packaged = name in recipe_versions
+            existing = recipes.get(name)
+            if existing is not None:
+                # Two entries for one recipe - two files naming it, or two
+                # entries in one file. Overwriting would leave the counters
+                # holding the replaced entry's CVEs, and a report whose counts
+                # exceed what it carries fails its own check as malformed,
+                # which no setting overrides. Merge instead, by id: the same
+                # CVE reported twice is one CVE.
+                seen_ids = {c.get("id") for c in existing["cves"]}
+                added = [c for c in issues if c.get("id") not in seen_ids]
+                existing["cves"].extend(added)
+                stats.cves += len(added)
+                if packaged:
+                    stats.packaged_cves += len(added)
+                continue
+
             recipes[name] = {
                 "version": version,
                 "packaged": packaged,
--- a/meta-avocado-sbom/lib/avocado_sbom/verify.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/verify.py
@@ -2,8 +2,10 @@
 
 """Check a CVE report against the frozen envelope.
 
-Shape and derived counters only. Nothing here asserts on which packages,
-recipes or CVEs a report carries, so a mapping correction never turns CI red.
+Shape and derived counters only. Nothing here asserts on an expected set of
+packages, recipes or CVEs: every check is the document against itself, or
+against markers the same build produced. So a mapping correction never turns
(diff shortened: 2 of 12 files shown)


hiagofranco:
~~Draft: I am still testing this code~~

Edit: tested. See comment above.

jetm (commented):
Reviewed the SBOM/CVE-report change end to end: read the diff, traced `do_cve_report`'s
`AVOCADO_CVE_REPORT_STRICT`/`AVOCADO_CVE_REPORT_UNSCANNED_EXPECTED` wiring against the
README's stated contract, and ran the PR's own test suite (72/72 pass, including the
jsonschema-vs-checker cross-consistency tests). Nothing blocking.

Two things noted but not worth a line comment - the change's own README already
discloses the `unscanned_recipes` semantics change and reasons about who consumed the
old shape, and `verify.py`'s uniqueness/dedup checks are redundant with guarantees
`report.py`'s generator already provides, not a live gap.

meta-avocado-sbom/lib/avocado_sbom/report.py
```diff
@@ -206,6 +221,22 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
                 continue
 
             packaged = name in recipe_versions
+            existing = recipes.get(name)
```
jetm, line 232: Two entries naming one recipe at different versions merge into an entry that claims the first one's version while carrying both entries' CVEs. It is reachable on the unpackaged path, where `recipe_versions.get(name)` is falsy, so the stale-version branch never runs and neither entry is dropped. A two-entry file at 1.0 and 2.0 yields `version: "1.0"` carrying both CVEs. `entry()` defaults to 1.0, so all three merge tests are single-version and nothing covers the differing case. 18 of the 44 recipe entries in the August qemuarm64 report are unpackaged, which is where this lands. Worth a note on which version wins, if not a fix.

meta-avocado-sbom/lib/avocado_sbom/report.py
```diff
@@ -206,6 +221,22 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
...
+            if existing is not None:
+                # Two entries for one recipe - two files naming it, or two
+                # entries in one file. Overwriting would leave the counters
+                # holding the replaced entry's CVEs, and a report whose counts
+                # exceed what it carries fails its own check as malformed,
+                # which no setting overrides. Merge instead, by id: the same
+                # CVE reported twice is one CVE.
+                seen_ids = {c.get("id") for c in existing["cves"]}
+                added = [c for c in issues if c.get("id") not in seen_ids]
```
jetm, line 250: Dedup compares against the existing entry only, so an id repeated inside one entry's own `issue` list survives and gets counted twice. Counts and list still agree afterwards, so `check_report` sees nothing wrong. I went looking in the fixture and in the 44-recipe report and found zero occurrences in either, so this is hardening rather than a defect.

jetm (approved):
Ran both CI steps locally on a box that has `jsonschema` (4.26.0), so the schema tests that skip on yours actually executed: 72 tests, zero skips, green. Also diffed the schema against `verify.py` in all four directions - `required` against `TOP_LEVEL_TYPES`, `properties` against required-plus-optional, and the counts `required` list against `REPORT_COUNTERS`. Every difference is empty. Five mutations on the fixture each fail with a message naming the real discrepancy, exit 1 mutated and 0 clean, so the check can actually fail.

Separately, the counter split reproduces on another machine. A feed-scoped qemux86-64 build here reports `unscanned_recipes` 11 with the same composition you measured on qemuarm64: six packagegroups, four avocado meta-recipes, `glibc-locale`. The shipped-recipe denominator differs though, 218 against your 222, so the opt-out set looks structural rather than per-MACHINE - which means the declared value may generalise further than the comment claims. The partition closes on real data as well: 218 shipped = 11 unscanned + 96 no-record + 17 carrying CVEs + 94 clean.

On the compatibility exemption: agree. Freezing a counter that reported the opposite of what its name said would have been the worse trade, and the exemption is cheap precisely while nothing outside the layer reads the report. Two small notes inline.

meta-avocado-sbom/lib/avocado_sbom/verify.py
```diff
@@ -0,0 +1,500 @@
...
+    # Both lists partition the shipped recipes with the ones scanned and found
+    # in the NVD, so a consumer can account for every shipped recipe exactly
+    # once. Overlap would mean a recipe reported as both examined and not.
+    shipped = {
+        p["recipe"] for p in packages.values()
+        if isinstance(p, dict) and isinstance(p.get("recipe"), str)
+    }
+    unscanned_names = {n for n in unscanned if isinstance(n, str)}
+    no_record_names = {n for n in no_record if isinstance(n, str)}
+    both = sorted(unscanned_names & no_record_names)
```
jetm, line 227: **Partition invariant unenforced when `AVOCADO_CVE_REPORT_STATUS` is not `Unpatched`**

`_check_derived` enforces that `unscanned_recipes` and `no_cve_record_recipes` are disjoint and both scoped to shipped recipes, matching the README (line 330) and the schema description. It never checks the third leg of that partition: a shipped recipe must not land in both `recipes` and `no_cve_record_recipes`.

That holds by construction only when `AVOCADO_CVE_REPORT_STATUS` is `Unpatched` - an issue can only be Unpatched if its product had a CVE record. Under the documented, supported `Patched` (and `Ignored`) settings it does not hold: `check_cves()` only sets `cves_in_product` on a CVE that is neither ignored nor already in `patched_cves`, so a product whose every DB record was patched (backported patch or `CVE_STATUS`) is written as `cvesInRecord: "No"` while the recipe still carries `Patched` issues.

Reproduced against this head: with `AVOCADO_CVE_REPORT_STATUS = "Patched"`, a shipped recipe landed in both `recipes` and `no_cve_record_recipes`, `check_report(doc, health=False)` returned `[]`, and the partition summed to 2 against 1 shipped recipe.

A build on `Patched` publishes a report the checker calls clean while a recipe appears twice in the partition; a consumer following the README's "account for every shipped recipe exactly once" contract double-counts it. Either add the overlap check alongside the existing `both` check, or scope the README/schema claim to `Unpatched` explicitly.

meta-avocado-sbom/tests/test_verify.py
```diff
@@ -0,0 +1,586 @@
...
+            )
+
+    def assertClean(self):
+        self.assertEqual(verify.check_report(self.report), [])
+
+    def test_fixture_passes(self):
+        self.assertClean()
+        self.assertEqual(verify.additions(self.report), [])
+
+    # ENG-2317: packages disappear, the counter does not notice.
```
jetm: **Bare Linear issue ID committed to public source**

This references `ENG-2317`; the same pattern recurs at line 239 (`ENG-2196`) and line 406 (`ENG-2364`), and the HEAD commit body ends `Refs ENG-2347, ENG-2364.` A grep across the whole tree at this head shows these are the only such references in the repo - new with this PR, not an existing convention.

`avocado-linux/meta-avocado` is public. A bare `ENG-####` is unresolvable to anyone outside the originating workspace, and it rots independently of the code around it - nothing here validates it still points at a real, current ticket. Each comment already states the failure shape it's guarding without the ID; consider dropping the ticket references and keeping the behavioral description.

jetm (commented):
Re-review after the head moved. Ran the full test suite (89 passed), the CI contract command, and reproduced both findings below against synthetic cve/pkgdata fixtures.

2 blocking findings posted inline. One more, not tied to a diff line so it's here instead: every one of this PR's 10 commits (`670c6682..5a29cd42`) carries a `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>` trailer, which breaks with the last 40 commits on `origin/scarthgap` (human co-authors only, no AI trailer). GitHub preserves `Co-authored-by` through a squash merge, so this lands in permanent public history unless it's stripped before merge.

8 additional advisories (docstring gaps, an unpinned `jsonschema` dependency, a couple of minor edge-case mismatches) came up but didn't clear the bar for a blocking comment - not posted.

meta-avocado-sbom/lib/avocado_sbom/verify.py
jetm: Measured how far this reaches on a real tree, since "reachable under a supported setting" and "reaches something we ship" are different claims.

Feed-scoped qemux86-64 build (`avocado-distro`, SPDX 3.0 plus cve-check), 396 cve-check files, 218 shipped recipes. Ten entries have every product at `cvesInRecord: "No"` while still carrying `Patched` or `Ignored` issues, which is the shape that lands a recipe in both lists. Two of those ten ship a package: `qemu-user-static` and `util-linux-libuuid`.

So on this tree, a build with `AVOCADO_CVE_REPORT_STATUS = "Patched"` publishes a report where 2 of 218 shipped recipes appear twice in the partition, and `check_report` returns no failures for it.

The oe-core mechanism is in `cve-check.bbclass`: the `cve_ignore` and `patched_cves` branches both `continue` before `cves_in_product` is set, so a product whose every database record was already patched or ignored is written as `cvesInRecord: "No"` while its issues survive at those statuses. Nothing about that depends on the recipe, so the count moves with the advisory database rather than staying at two.

Worth noting which way the fix cuts. Adding the overlap check makes a `Patched` build fail on data it accepts today, so it is not purely additive for anyone already running that setting - though a report that double-counts is the thing being caught, not a regression introduced. Scoping the README and schema claim to `Unpatched` instead leaves the report honest at the cost of a narrower contract, and `unscanned_recipes` plus `no_cve_record_recipes` stop partitioning anything a consumer can rely on.

meta-avocado-sbom/lib/avocado_sbom/report.py
hiagofranco: Took the note. dfaf19a documents the merge as first-entry-wins, and says first means sorted filename order then file order - stable across runs rather than correct.

The reason it stays a note: two entries at different versions only reach the merge unpackaged, where nothing this build ships is described by either version, so there is no version to prefer. A packaged recipe has a known version and any entry at another one was dropped as stale above.

Added a test pinning the documented behaviour, since `entry()` defaulting to 1.0 left all three merge tests single-version as you noted.

meta-avocado-sbom/lib/avocado_sbom/report.py
hiagofranco: Fixed in dfaf19a, though one step earlier than the thread suggests. Deduping in the merge only covers the second entry onward; an id repeated inside the *first* entry for a recipe never reaches the merge at all, because that path goes straight to `recipes[name] = ...`. So the dedup now happens as `issues` is built, which covers both shapes with one seen-set.

Agreed it is hardening - I also found zero occurrences. Test added.

meta-avocado-sbom/lib/avocado_sbom/verify.py
hiagofranco: Fixed in dfaf19a, taking neither of the two options - the overlap is generator data, so it is fixed there:

    return recipes, scanned, scanned - with_record - set(recipes)

`cvesInRecord: "No"` means no record *matched*, not that none exists; as you traced, `cve_ignore` and `patched_cves` both `continue` before `cves_in_product` is set. That makes the field unusable on its own, so what the report carries decides: a recipe reporting CVEs is not one with no CVE record, whatever the product says.

This keeps the README contract at full width, and a `Patched` build stops publishing a double-counted report rather than starting to fail on one - which answers the "not purely additive" concern in your follow-up. The ten entries you measured now sit in `recipes` only.

Also added the third leg to `_check_derived` as you suggested, so it is enforced rather than true by construction. With the generator fixed it never fires on our data, but the pair that broke was lists-vs-`recipes` and nothing was testing it.

Tests: the two shapes at `read_cve_data` level plus an end-to-end `build_report` at `status="Patched"` asserting the partition sums to the shipped set. All three fail against the previous head.

meta-avocado-sbom/tests/test_verify.py
hiagofranco: Removed in dfaf19a - all three sites now state the failure shape without the id, which as you say is what they were already doing.

Left the two commit bodies below HEAD alone; rewriting those means a force-push mid-review, which seemed worse than the bare refs. Say if you would rather I did.

meta-avocado-sbom/lib/avocado_sbom/verify.py
```diff
@@ -0,0 +1,509 @@
...
+    undeclared = sorted(
+        n for n in unscanned if isinstance(n, str) and n not in declared
+    )
+    if not undeclared:
+        return []
+
+    return [
+        "%d recipe(s) shipped a package, have no cve-check result and declare "
+        "no reason for it: %s. Either something stopped scanning them - a "
+        "cleared CVE_PRODUCT, an interrupted build, a layer that fell out of "
```
jetm: **Opt-out gate cannot catch a cleared `CVE_PRODUCT` - the diagnostic and README both claim it does**

The undeclared-unscanned failure message names three causes: "a cleared `CVE_PRODUCT`, an interrupted build, a layer that fell out of `CVE_CHECK_LAYER_INCLUDELIST`". Two of the three can never produce this failure. `avocado_cve_optout_reason()` deliberately mirrors the same predicates `cve-check.bbclass` bails on (bbclass:29-45), so an accidentally emptied `CVE_PRODUCT` writes a `CVE_PRODUCT is empty` marker, and a layer dropped from `CVE_CHECK_LAYER_INCLUDELIST` writes its own marker - both are declared opt-outs and both pass cleanly.

This isn't just the error string: README:161 states the opposite as this design's whole justification ("a recipe whose `CVE_PRODUCT` got cleared fails the build naming itself"), directly contradicting the paragraph above it that lists `CVE_PRODUCT = ""` as one of the four declared opt-out mechanisms - and the bbclass header (lines 9-16) frames the same clobbered-`CVE_PRODUCT` case as the hole the markers close.

What the gate actually catches is "the scan did not run at all" (interrupted build, class not inherited, recipe absent from this build's graph but present in pkgdata) - worth having, but a narrower guarantee than three places in this PR claim. Suggest fixing the diagnostic to name only causes it can actually have, and correcting the README/bbclass rationale to describe what the marker set does and does not separate.

Concretely: land `CVE_PRODUCT = ""` on a real recipe (a bad rebase, a copied bbappend). `cve-check` writes no `*_cve.json`, `avocado-cve-optout` writes an indistinguishable-from-legitimate marker, `do_cve_report` emits only a `bb.note`, and the report's `unscanned_recipes` silently hides a recipe nobody is scanning any more - the exact event the README says this mechanism converts into a named build failure.

jetm (commented):
Re-review after the last two blocking findings. Both verified independently, not taken on the "resolved" thread status alone: the SBOM partition-overlap bug is fixed (`report.py:276` and `verify.py:232-238` now cover the third leg; reproduced against `Patched`/`Unpatched`/`Ignored` fixtures), and the bare Linear issue IDs are gone (grepped the tree, workflow, and `kas/feature/cve-check.yml` - nothing left).

1 new blocking finding posted inline: the opt-out gate's own diagnostic and the README both claim it catches a cleared `CVE_PRODUCT`, but it structurally cannot - that case is one of the four declared opt-outs the same gate treats as legitimate.

5 advisories (a README claim about counter-change CI coverage that doesn't hold for `report.py` alone, a minor stale-recipe accounting note, a no-op test assertion, a `cancel-in-progress` race on the contract workflow, and an edge case in issue-id deduplication) came up but didn't clear the bar for a blocking comment - not posted.

meta-avocado-sbom/lib/avocado_sbom/verify.py
jetm: Reproduced on a real report, and the marker data makes this narrower than a diagnostic fix.

Feed-scoped qemux86-64, 218 shipped recipes: appending a genuinely-shipped recipe to `unscanned_recipes` with reason `CVE_PRODUCT is empty` yields zero gate failures. Same with `layer ... is not in CVE_CHECK_LAYER_INCLUDELIST`. One failure when no marker exists. So the gate's real guarantee is "no cve-check result and no marker", which is the interrupted-build case only.

The complication is that all 15 markers this build writes carry the same reason, `CVE_PRODUCT is empty` - `glibc-locale` and the 14 image and packagegroup recipes alike. The other three mechanisms are unused on this tree. An accidentally cleared `CVE_PRODUCT` is therefore indistinguishable from every legitimate deferral, so the gate cannot be repaired by keying on the reason string without failing all 15.

That leaves the doc correction as the only fix available, unless the marker starts recording intent rather than mechanism. `avocado_cve_optout_reason()` mirrors the predicates `cve-check` bails on and the class header calls that deliberate, which is exactly why the reason cannot separate a deferral from a regression.

meta-avocado-sbom/lib/avocado_sbom/verify.py
jetm: Ran the pre-fix and post-fix modules over the same 396 cve-check files from a feed-scoped qemux86-64 build, 218 shipped recipes, to see what the change does to a real report rather than a fixture.

At `Patched`, pre-fix: three shipped recipes land in two legs at once - `libusb1`, `qemu-user-static`, `util-linux-libuuid` - and `check_report` returns no failures on that document. Post-fix: none. `no_cve_record_recipes` goes 95 to 92, and the legs sum to exactly 218 against 218 shipped.

At the default `Unpatched` the two documents are byte-identical, 2,044,836 bytes each with `generated` stripped. So the subtraction cannot regress what we ship, which was the concern about fixing this generator-side rather than in the checker.

Agreed on `cvesInRecord`. "No record matched" and "no record exists" sharing one field is the root of it, and letting the reported CVEs decide is the right tiebreak.

meta-avocado-sbom/tests/test_verify.py
jetm: Source is clean - `ENG-[0-9]+` across the tree at 2634188 returns nothing.

The two commit bodies are worth fixing, because the merge does not drop them. The `main` ruleset restricts merges to squash only, and the repo's squash message is set to `COMMIT_MESSAGES`, so every commit body is concatenated into the commit that lands on `main`. The refs arrive intact.

No dedicated force-push needed though. `dismiss_stale_reviews_on_push` is false on that ruleset, so a push costs no approvals - roll it into whatever the next one is. The squash body is also editable in the merge dialog, but that relies on whoever merges remembering.

meta-avocado-sbom/lib/avocado_sbom/report.py
jetm: Ack, and the stability claim checks out: `paths = sorted(glob.glob(...))` at report.py:157, so first is deterministic filename order and the merged entry's version does not move between runs. Agreed unpackaged is the only way in, which makes "stable rather than correct" the honest framing. The new test covers the case `entry()`'s 1.0 default was hiding.

meta-avocado-sbom/lib/avocado_sbom/report.py
jetm: Right, and one step earlier is the better place for it. The first entry for a recipe goes straight to `recipes[name] = ...` and never reaches the merge, so a seen-set at the merge would have missed the shape most likely to occur. One set covering both is also less to keep in sync.

meta-avocado-sbom/lib/avocado_sbom/verify.py
jetm: Fixed in 3cd2c23, as a claims correction rather than a mechanism change.

The diagnostic now names only the cause that can reach it - the scan not having run, so an interrupted build, this class not inherited, or a recipe still in pkgdata the graph no longer reaches. README:161 says a recipe that stops being scanned *and declares nothing* fails, and a new paragraph states what the marker cannot separate. The class header no longer frames the clobbered case as the hole it closes.

Making the gate catch it instead does not work, which is why it went this way. All 15 markers a feed-scoped qemux86-64 build writes carry the reason `CVE_PRODUCT is empty`, `glibc-locale`'s deliberate deferral included, so failing on that reason fails every legitimate opt-out in the build. The reason mirrors the predicate cve-check bailed on, deliberately, and that is exactly what leaves intent unrecorded. Catching an accidental opt-out needs the marker to carry what was intended, which is a different mechanism than this class collects.

Two tests hold it: one asserts the message names none of the four mechanisms `avocado_cve_optout_reason()` can report, so it cannot regrow a cause that writes a marker; the other pins the limitation so the docs cannot drift back to claiming it is caught.

jetm (approved):
Approving. The last blocking finding is addressed and every earlier one was verified against a real build rather than the fixture.

Disclosure on scope: the head commit 3cd2c23 is mine, pushed to this branch to close the opt-out-claims finding. So this approval covers Hiago's twelve commits, and 3cd2c23 itself carries no second reader. It is a diagnostic string, two doc sites and two tests, with no logic or control-flow change.

What was checked by running it, on a feed-scoped qemux86-64 build with 396 cve-check results over 218 shipped recipes:

The partition fix holds where it matters and changes nothing where it should not. At `Patched`, the pre-fix generator put three shipped recipes - `libusb1`, `qemu-user-static`, `util-linux-libuuid` - in two legs of the partition at once, and `check_report` returned no failures on that document. Post-fix there are none, `no_cve_record_recipes` moves 95 to 92, and the legs sum to exactly 218 against 218 shipped. At the default `Unpatched` the two documents are byte-identical at 2,044,836 bytes, so the shipped path cannot regress.

The opt-out gate's reachable causes are now the only ones it names. Appending a genuinely-shipped recipe to `unscanned_recipes` with reason `CVE_PRODUCT is empty` yields zero gate failures, likewise for a dropped `CVE_CHECK_LAYER_INCLUDELIST`, and one failure when no marker exists - so "no result and no marker" is the whole of it, and the docs now say that.

Every figure the README gained is right: 11 unscanned as 6 packagegroups, 4 avocado meta-recipes and glibc-locale, over 218 shipped, with 15 markers for those 11. The merge-order stability claim checks out at report.py:157, `sorted(glob.glob(...))`.

Suite is 98 tests green, up from 89 at the first pass, no skips. All three CI checks pass on the head.

jetm (approved):

nicksinas (approved):

PR #300, opened 2026-08-22
Port rust-native uninative rpath fix to wrynose

jetm:
## Problem

rust-native's snapshot bootstrap fails under uninative:
`rust-snapshot/bin/rustc: error while loading shared libraries:
libz.so.1`. do_rust_setup_snapshot repoints the snapshot's binaries at
the uninative loader but never sets an rpath, and libLLVM needs
libz.so.1, which the uninative tarball doesn't ship. zlib-native is
already staged in recipe-sysroot-native - the loader just has no path
to it. wrynose has no fix for this at all; an older variant exists
only on unmerged feature branches.

## Solution

Port the fix already landed on scarthgap, verbatim. It appends
STAGING_LIBDIR_NATIVE to each affected binary/library's existing
RUNPATH (never asserting a fixed one), guarded on UNINATIVE_LOADER,
using patchelf-uninative rather than bare patchelf so no extra
build dependency enters rust's critical path. Confirmed directly
against the wrynose-pinned oe-core commit that do_rust_setup_snapshot
still exists under that name with the same
${WORKDIR}/rust-snapshot/bin,lib layout this fix assumes.

## Key changes

- `meta-avocado/recipes-devtools/rust/rust_%.bbappend` (new): ports
  the fix, placed alongside the existing `libstd-rs_%.bbappend` in
  the same directory.

## Reviewer notes

An older, less robust variant of this fix exists uncommitted on a
couple of feature branches (a bare-patchelf form with a task-name-
bound `[depends]`). Left those alone rather than hand-patching them -
they'll pick up this version on their next rebase from wrynose.

Verification: this exact content was already build-verified with a
full native-closure build (rust-native/1.94.1 built successfully). I
have not re-run that build myself for this PR.

### DIFF
--- a/meta-avocado/recipes-devtools/rust/rust_%.bbappend
+++ b/meta-avocado/recipes-devtools/rust/rust_%.bbappend
@@ -1,0 +1,104 @@
+# do_rust_setup_snapshot repoints the prebuilt snapshot's three binaries at the
+# uninative loader but sets no rpath, and never touches the libraries beside
+# them. Those carry RUNPATH=$ORIGIN/../lib, and libLLVM needs libz.so.1 - which
+# the uninative tarball does not ship: uninative-tarball.bb packages glibc,
+# patchelf, libxcrypt, libnss-nis and libgcc, nothing more. The uninative loader
+# searches RUNPATH and its own sysroot and never the host default paths, so
+# rustc cannot start and bootstrap fails on `rustc -vV` returning 127:
+#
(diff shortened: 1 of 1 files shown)


mobileoverlord (approved):
Byte-identical to the bbappend already landed on scarthgap (verified with a literal diff), targeting wrynose where the fix is absent. Guarded on UNINATIVE_LOADER, appends rather than asserts RUNPATH — nothing to add.

PR #302, opened 2026-08-23
Add fuse-free FIT verified boot for avocado-imx93-frdm

jetm:
## Problem

`avocado-imx93-frdm` had no way to establish that the kernel, device tree and
initramfs it boots are the ones we built. AHAB provides that, but it requires
burning SRK fuses — irreversible, and the lab board's `SRK_HASH` is already
spent (burned byte-swapped), so AHAB cannot be re-validated on it at all. That
leaves every non-fused board with no boot-integrity story.

## Solution

U-Boot FIT signature verification as a fuse-free route: one FIT carrying kernel,
device tree and initramfs, signed at the configuration node, with the public key
embedded in U-Boot's control DTB and `required = "conf"` making verification
mandatory rather than advisory.

The initramfs rides as its own `ramdisk-1` node rather than being bundled into
the kernel. `conf/distro/avocado.conf` already defaults `INITRAMFS_IMAGE_BUNDLE`
to `0`; bundling was tried first and abandoned on hardware because it produced a
single 167.9 MiB decompression target and bootm aborted with `Image too large:
increase CONFIG_SYS_BOOTM_LEN`. Unbundled, what bootm inflates is the bare
33.9 MiB kernel, inside the existing limit. Signature coverage is unchanged —
`oe/fitimage.py` adds `ramdisk` to the configuration node's signed entries
beside `kernel` and `fdt` — so one signature still spans all three and the
initramfs cannot be substituted independently of its kernel. It also puts the
`CONFIG_SYS_BOOTM_LEN` ceiling on the kernel rather than on the initramfs, which
is the component that grows as cryptsetup, fTPM and dm-verity tooling land in it.

## Key changes

- New `avocado-imx93-frdm-fitimage.bb` inheriting OE-core's `kernel-fit-image`.
  Note for anyone working from older notes: `kernel-fitimage.bbclass` does not
  exist in this oe-core release — it was replaced by `kernel-fit-image.bbclass`,
  used as a standalone recipe depending on `virtual/kernel:do_deploy`.
- A fifth `FIT` key pair in `gen-sbkeys.sh`; the private half is never installed.
- `CONFIG_FIT` split into an unconditional `fit.cfg` from the feature-gated
  `fit-verify.cfg`, so the container format is unconditional for this machine
  while signature verification stays opt-in via `kas/feature/verified-boot.yml`.
- Boot env switched to a single `bootm` of the FIT, and `image_addr` moved to
  `0x90000000` (see reviewer notes).
- `stone-imx93-frdm.json` boot files reduced to `fitImage` alone, so no
  unverified payload remains on the boot partition.

## Reviewer notes

**Verified on hardware** (FRDM-IMX93, slot B, `/var` on `/dev/mapper/var`):

| Payload | Result |
| --- | --- |
| Intact signed | `Verifying Hash Integrity ... sha256,rsa2048:FIT+ OK`, boots to a login prompt |
| One byte flipped in the kernel | refused — `Bad hash value for 'hash-1' hash node in 'kernel-1' image node` |
| One byte flipped in the initramfs | kernel verifies OK first, then refused naming `ramdisk-1` |
| Byte-identical payload, unrelated signing key | refused at the config node — `Failed to verify required signature 'key-FIT'` |
| No payload | `Failed to load 'fitImage'` — distinct from any verification failure |

Only the fourth row proves authenticity. Rows two and three are hash refusals,
which FIT gives you with no key at all — the identical refusals appear on a
bootloader carrying no key, so they are not evidence of verified boot.

**Two bugs only hardware could surface.** The first: `image_addr` was
`0x80200000` while the FIT's kernel subimage decompresses to `0x80400000`, so
bootm inflated from inside the FIT to a destination 2 MiB into its own
compressed source, overwriting its input and returning `inflate() -3` *after*
the hashes had read OK — it presents as payload corruption rather than the
address collision it is. The old raw-image path never hit it because an
uncompressed kernel is copied, not inflated. The collision is independent of the
bundling decision. The second is the `CONFIG_SYS_BOOTM_LEN` overflow described
above.

**`AVOCADO_SECURITY_CAPABILITIES` deliberately does NOT declare
`verified-boot`.** Hardware verification also showed the capability is
bypassable by the attacker the threat model names: `CONFIG_ENV_IS_IN_MMC=y` puts
the environment in the unsigned, writable `uboot-env` partition where it
overrides the compiled-in default, `CONFIG_CMD_BOOTI=y` and
`CONFIG_LEGACY_IMAGE_FORMAT=y` leave unsigned boot paths available, and there is
no `CONFIG_ENV_WRITEABLE_LIST`. So an attacker with offline write access to the
card can repoint `bootcmd` at a raw unsigned kernel without touching the FIT.
Declaring the capability while that is open would over-claim, so it is deferred
to a follow-up that closes the env hole first. The A/B update path needs a
writable env (`avocado_boot_slot` and the slot stamps are written by fwup and by
the update agent), so the fix is a selective whitelist rather than a read-only
env.

**Measured boot is not provided.** Both u-boot versions here have `CONFIG_TPM`
unset and no `CONFIG_MEASURED_BOOT`, so FIT verification extends no PCRs at all.
Any motivation resting on sealing a `/var` key to a PCR reflecting verified-boot
state has no basis in this implementation and has been withdrawn rather than
left asserted.

**Stacking.** This branch is cut from the local tip of
`imx93-secure-boot-wrynose`, which is 7 commits ahead of its pushed state, so
this PR currently shows those commits too. They belong to the in-flight
capability-declaration and fTPM work on #271, not to this change. This PR's own
12 commits run `f262fd2a..995e8783`; the diff resolves itself once #271 is
updated. Hardware verification was performed with those 7 commits applied, which
is why the branch was not rebased to produce a cosmetically cleaner diff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,10 @@
+# Both spellings. `build-*/` alone left a bare `build/` untracked-but-not-ignored,
+# and that is where sb-keys generates its private keys - AVOCADO_SB_KEYS_DIR
+# defaults to ${TOPDIR}/avocado-sb-keys. Since FIT verified boot made FIT.key
+# load-bearing, one `git add -A` in a clone using a bare build dir would commit
+# the key that signs every bootable payload for those devices into a public repo.
 build-*/
+build/
 _repo
 # Python bytecode from helper scripts
 __pycache__/
--- a/kas/feature/verified-boot.yml
+++ b/kas/feature/verified-boot.yml
@@ -1,0 +1,6 @@
+header:
+  version: 16
+
+local_conf_header:
+  feature/verified-boot: |
+    DISTRO_FEATURES:append = " verified-boot"
--- a/meta-avocado-nxp/conf/machine/avocado-imx93-frdm.conf
+++ b/meta-avocado-nxp/conf/machine/avocado-imx93-frdm.conf
@@ -23,21 +23,28 @@ require conf/machine/include/fsl-imx-preferred-env.inc
 # same version for the same reason (see its own PREFERRED_VERSION_u-boot-imx).
 PREFERRED_VERSION_u-boot-imx = "2025.04"
 
-# INITRAMFS_IMAGE/INITRAMFS_IMAGE_BUNDLE must be set here, not from
-# linux-imx_%.bbappend: kernel.bbclass's OWN early anonymous python checks
-# both to decide whether to addtask do_bundle_initramfs/do_transform_bundled_initramfs
-# AT ALL - the same parse-order problem KERNEL_CLASSES had. Setting them from
-# a bbappend's later-running anonymous python left both tasks never added, so
-# uboot_prep_kimage fell back to a bare vmlinux and the FIT shipped with no
-# initramfs coverage at all - confirmed by a real build (mkimage -l showed no
-# Ramdisk: line in any configuration). Bundling into vmlinux.initramfs (kernel
-# recipe's own mechanism), not a separate FIT ramdisk node, is why
-# avocado-imx93-frdm-fitimage.bb also sets INITRAMFS_IMAGE_BUNDLE - it tells
-# kernel-fit-image.bbclass NOT to add a redundant separate ramdisk section,
-# since the initramfs is already folded into the kernel-1 image by the time
-# that recipe reads it.
-INITRAMFS_IMAGE = "avocado-image-initramfs"
-INITRAMFS_IMAGE_BUNDLE = "1"
+# This machine deliberately does NOT override INITRAMFS_IMAGE_BUNDLE, so the
+# distro default in conf/distro/avocado.conf ("0") applies and the initramfs
+# rides in the FIT as its own signed ramdisk-1 node rather than being folded
+# into the kernel image.
+#
+# An earlier revision of this file set INITRAMFS_IMAGE_BUNDLE = "1" here. That
+# bundled a 134 MiB uncompressed initramfs into the kernel, giving a single
+# 167.9 MiB decompression target - past CONFIG_SYS_BOOTM_LEN on both the
+# board's u-boot (0x8000000) and our own build (0x4000000), so bootm aborted
+# with "Image too large: increase CONFIG_SYS_BOOTM_LEN" on hardware. Keeping
+# the two payloads separate drops what bootm inflates to the bare 33.9 MiB
+# kernel, which fits the existing limit and needs no Kconfig change.
+#
+# It is also the better place for the cap to sit long-term: bundling charges
+# CONFIG_SYS_BOOTM_LEN for the initramfs, which is the component that grows as
+# cryptsetup, fTPM and dm-verity tooling land in it. Unbundled, the limit only
+# ever has to cover the kernel and initramfs growth is unbounded by it.
+#
+# Signature coverage is unchanged, which is the property that matters here:
+# oe/fitimage.py appends "ramdisk" to the configuration node's sign_entries
+# alongside "kernel" and "fdt", so the initramfs is authenticated by the same
+# configuration-node signature either way.
 
 # FIT signing is gated here, not per-recipe: avocado-imx93-frdm-fitimage.bb
 # (kernel-fit-image.bbclass) and u-boot-imx_%.bbappend (uboot-sign.bbclass)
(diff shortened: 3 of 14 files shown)


jetm:
Converted to draft. A high-effort review of this branch found defects that
neither the build nor the hardware pass caught, because both ran against a tree
that already held generated keys and stale kernel artifacts. Fixed in `b74ea96d`;
recording them here because two were security-relevant and because the pattern
matters more than the individual bugs.

## Fixed

- **`DEPENDS += sb-keys` had been deleted.** An earlier commit added it precisely
  so key generation was ordered before signing; the kernel-fit-image redesign
  removed the block carrying it and the replacement recipe declared none. Nothing
  sequenced the two, so `bitbake avocado-imx93-frdm-fitimage` on a tree without
  keys fails in `run_mkimage_sign`. A tree that already has `FIT.crt` hides it
  entirely, which is why two full builds and a board pass all went green.
- **Key generation was gated on `PK.crt`, not per key.** Any directory created
  before the FIT role existed would never gain it, and `AVOCADO_SB_KEYS_DIR`
  survives `cleanall`, TMPDIR wipes and sstate purges. Both that and a partial
  keyset end in the same place: the kernel side refuses loudly, but
  `uboot-sign.bbclass`'s `concat_dtb` has no key-existence check, so `mkimage`
  embeds nothing, the signature check passes vacuously, and the bootloader ships
  with `CONFIG_FIT_SIGNATURE=y` and no trust anchor - accepting any FIT.
- **The FIT signing private key was not ignored.** Only `build-*/` was in
  `.gitignore`, so with a bare `build/` the key sat untracked and unignored. One
  `git add -A` would publish the key that signs every payload for those devices.
- **The Kconfig `cat` lines were dead.** `UBOOT_DEFCONFIG` expands to the literal
  `['sd']` (its `.split((',', 1)[0])` is `.split(',')`, returning a list), so
  every cat line in that bbappend appends to a file of that name and the real
  defconfig is never touched - confirmed in `run.do_configure`. The fragments
  reach `.config` through `cml1`'s `merge_config.sh` over `SRC_URI`, which is why
  it looked correct. Apparatus deleted; the broken expansion is left alone
  (fixing it would relocate two pre-existing fragments) but now carries a warning.
- **`kernel-fit-extra-artifacts.bbclass` was hand-copied.** That class is what
  `kernel-fit-image` names as its requirement, and the copy passed `${DEPLOYDIR}`
  where the class passes `$deployDir`, dropping `KERNEL_DEPLOYSUBDIR` - which
  would put `linux.bin` out of reach on any multi-kernel machine. Now selected
  via `KERNEL_CLASSES` from the machine conf.
- **`FIT_CONF_DEFAULT_DTB` was unset**, so the default configuration node was
  whichever DTB came first in `KERNEL_DEVICETREE` across two vendor layers that
  append eight variants. Pinned.
- **`key-store.bb`** packages the same directory as `sb-keys.bb` and had been
  left at four key roles.

Verified against a clean kernel rebuild: `.init.data` back to 236 KB from 140 MB,
`linux.bin` 15,581,719, one `ramdisk-1` node, config node signed
`sha256,rsa2048:FIT`, default configuration pinned.

## Left open deliberately

- **`image_addr` not moved.** It sits 60% into a 96 MiB window bounded by bank0's
  edge at `0x96000000`, which is also OP-TEE's `CFG_TZDRAM_START`. `load` does not
  bounds-check, so outgrowing it presents as a hash mismatch rather than an
  overflow. The ceiling is now documented and `0x84000000` recorded as the escape
  hatch - but moving there needs its own board pass, and swapping a
  hardware-verified address for an unverified one in this change would defeat the
  point of it.
- **`verified-boot` is still absent from `AVOCADO_SECURITY_FEATURES`**, so
  requesting it on a machine that cannot deliver it remains a silent no-op. The
  remedy is coupled: registering the feature without also declaring the capability
  turns the intended build into a `bb.fatal`, and declaring the capability is the
  over-claim this PR intentionally defers while the unsigned-env bypass is open.
  Both belong together in the follow-up.
- **No OTA migration path.** The update payload never replaces `uboot-env`, so a
  device provisioned before this change cannot boot the new `boot.img` - it rolls
  back and survives, but can never take this or a later update without physical
  re-provisioning. That is a deployment-policy decision, not a code fix.

## Why it is a draft

Three of the findings above were invisible to the hardware verification this PR
rests on, and two of those were security-relevant. Every wrong conclusion reached
while reviewing came from trusting a dirty build tree - stale keys hid the
`sb-keys` race, a stale `vmlinux` produced a phantom regression. Worth a reviewer
who has not been in that tree.

Also still stacked: this branch is cut from the local tip of the base, so the diff
shows commits belonging to the in-flight capability-declaration work. It resolves
when the base is updated.

PR #303, opened 2026-08-23
Add a bringup target overlay for new-hardware bringup

jetm:
## Problem

Bringing up a new board means waiting on the full `avocado-distro` meta-target - Qt5, chromium, gstreamer, opencv, nodejs, docker/k3s, openjdk and the cross SDK - none of which tells you whether the board boots.

## Solution

Add a board-agnostic kas overlay that swaps the meta-target for `avocado-image-rootfs` and adds a bringup toolset (network, bus and peripheral probing, storage, device tree, tracing). It layers onto any `kas/machine/<board>.yml` and touches no packagegroups, so boards that do not use it are unaffected:

```
bakar build kas/machine/<board>.yml:kas/target/bringup.yml
```

## Key changes

- New `kas/target/bringup.yml`; one file, no changes to any existing path
- Login features spelled out rather than `debug-tweaks`, which is defined in `core-image.bbclass` and makes bitbake skip `avocado-image-rootfs` (surfacing as the misleading `Nothing PROVIDES 'avocado-image-rootfs'`)
- `AVOCADO_DEV_ROOT_LOGIN = "1"`, which the image features do not imply: `avocado-users` ships root locked as `root:*:`, and `empty-root-password` only preserves an already-empty password
- `read-only-rootfs` declared, without which `sshd_check_keys` fails writing host keys into the erofs `/etc/ssh` and the shipped sshd listens with no key

## Reviewer notes

No production impact - nothing references the overlay unless a build names it.

Verified on `avocado-imx93-frdm`: builds via `avocado-stone -c stone_provision` at 5935 tasks with no errors, the produced rootfs carries `root::` rather than `root:*:` and ships `sshd_config_readonly`, and the board reaches a working console and ssh login.

`kernel-modules` is kept as the default with its cost documented inline: on `avocado-imx93-frdm` it pulls `kernel-module-hci-uart`, which claims the console UART about 84 seconds into boot and makes a still-running board look hung. Boards in that position should name the modules they need instead.

### DIFF
--- a/kas/target/bringup.yml
+++ b/kas/target/bringup.yml
@@ -1,0 +1,101 @@
+# Bringup target overlay - minimal, fast, bootable image for new-HW bringup.
+#
+# Layer on top of any kas/machine/<board>.yml to build the base Avocado rootfs
+# (avocado-image-rootfs) instead of the full avocado-distro meta-target. That
+# skips the shared extras packagegroup (Qt5, chromium, gstreamer, opencv,
+# nodejs, docker/k3s, openjdk, ...) and the cross SDK, so a fresh board boots
+# and is debuggable without paying for any of it. Touches no packagegroups -
+# everything here is image-scoped.
+#
+# Usage (board-agnostic) - the overlay carries its own target, so no image
+# argument is needed; both commands build avocado-image-rootfs:
+#   bakar build \
+#     meta-avocado/kas/machine/<board>.yml:meta-avocado/kas/target/bringup.yml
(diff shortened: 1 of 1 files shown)


mobileoverlord (approved):
Purely additive opt-in kas overlay — no packagegroup or shared-config changes, so boards not using it are untouched. The image.bbclass-vs-core-image feature pitfalls are exactly the ones that bite, and the comments document them where the next person will look.

PR #304, opened 2026-08-24
avocado-users: add the sshd privilege-separation account for scarthgap

jetm:
## Problem

Any image built with the networking feature installs `openssh-sshd`, and that sshd cannot start. It listens on port 22, then every connection dies at `kex_exchange_identification` — and `journalctl -u sshd@` reports **no entries at all**, so the unit looks healthy while nothing works.

There is no `sshd` account. oe-core's openssh creates one through the useradd class (`USERADD_PARAM:${PN}-sshd = "--system --no-create-home --home-dir /var/run/sshd --shell /bin/false --user-group sshd"`), and `avocado-users` installs a fixed `/etc/passwd` and `/etc/group` over the top, discarding it along with any other useradd-created system user. OpenSSH refuses to run without its privilege-separation account and exits before writing a banner or a log line — which is what makes this expensive to diagnose rather than merely broken.

## Solution

Add the account to the shipped user database, mirroring that `USERADD_PARAM`'s home directory and shell so the two do not drift. uid and gid 992 are free on this branch and stay in the system range.

## Relationship to #276

**Complementary, not overlapping.** #276 carries the host-key work for this branch — installing `openssh-sshd` and moving the keys onto `/var` — and this is the one piece it does not have.

The two are independent, and neither supersedes the other:

- a host key in a writable directory still gets no sshd without the account
- the account alone still leaves the key on a read-only path

So the merge order does not matter, but **sshd will not actually come up until both have landed.** Worth knowing when validating #276, since without this change its fix will appear not to have worked — the symptom is identical.

## Reviewer notes

**The clobber is the real bug and is deliberately not fixed here.** Shipping a whole `/etc/passwd` discards every useradd-created system account, so the next package needing one hits the identical wall. The proper fix is to pin only the uid/gid this distro cares about (sysusers.d or `USERADDEXTENSION`) and let useradd own the rest — that changes how every Avocado image gets its users and wants its own verification, so it carries a `devtool-debt` marker with a ceiling and an upgrade trigger instead.

**Found on the wrynose line** while diagnosing an sshd that reported `Finished OpenSSH Key Generation` and still dropped every connection. Adding the account there produced `sshd:x:992:992` in the built rootfs and a running server. scarthgap reaches the same failure by the same route, since it ships the same `avocado-users` and installs openssh through `packagegroup-avocado-feature-networking`.

### DIFF
--- a/meta-avocado-distro/recipes-avocado/avocado-users/avocado-users_1.0.bb
+++ b/meta-avocado-distro/recipes-avocado/avocado-users/avocado-users_1.0.bb
@@ -15,6 +15,24 @@ SRC_URI = " \
   file://gshadow \
 "
 
+# The shipped files REPLACE /etc/passwd and /etc/group wholesale, so any system
+# user a package creates through the useradd class is silently discarded. That
+# is why sshd is listed here: oe-core's openssh declares
+# USERADD_PARAM:${PN}-sshd = "--system --no-create-home --home-dir /var/run/sshd
+# --shell /bin/false --user-group sshd", and without the account OpenSSH refuses
+# to start at all - it exits before writing a banner or a log line, so the only
+# symptom is every ssh connection dying at kex_exchange_identification with
+# nothing in the journal to explain it. The entry below mirrors that
+# USERADD_PARAM's home directory and shell; keep them in step.
+#
+# devtool-debt: sshd is hand-copied here rather than the clobber being fixed.
+# Ceiling: exactly the packages whose users someone has noticed and transcribed.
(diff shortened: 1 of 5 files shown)


mobileoverlord (approved):
uid/gid 992 is collision-free against the shipped list (44, 993-999, 65534), and the entry mirrors oe-core's USERADD_PARAM for openssh (home, shell, system user-group) as it must. The debt comment honestly scopes the real fix (stop clobbering /etc/passwd wholesale) as its own change. Note: may need a trivial rebase over #288's video line in files/group.

PR #305, opened 2026-08-24
ENG-2484: SPDX publication filter, gated on CVE identifiers

hiagofranco:
Splits the "produce a publishable SPDX document" half out of ENG-2356; the publish step that *calls* this is ENG-2279 box 3.

## What it does

`avocado_sbom.publish` reads a build tree and emits a copy of its SPDX documents with the CVE assessment removed, then refuses to write anything that still carries a CVE identifier.

```sh
PYTHONPATH=meta-avocado-sbom/lib python3 -m avocado_sbom.publish \
    --in build/tmp/deploy/spdx/3.0.1 -o publish/spdx
PYTHONPATH=meta-avocado-sbom/lib python3 -m avocado_sbom.publish \
    --in build/tmp/deploy/images/<machine> --check
```

**The gate is on CVE identifiers anywhere in the output, not on field names.** That is the whole design and it is not the obvious one:

| Build | Where the assessment lives | A `security_Vex*` gate | This gate |
| -- | -- | -- | -- |
| 2.2 | `sourceInfo` free text | passes, leaks | catches it |
| 3.0.1 | `security_Vex*` objects | catches it | catches it |

`create-spdx-2.2.bbclass` writes `recipe.sourceInfo = "CVEs fixed: " + patched_cves` and never consults `SPDX_INCLUDE_VEX`, so on 2.2 the field cannot be turned off at build time.

Patch filenames name the CVE they fix and survive any structural filter, so those names are redacted rather than dropped — the file and its checksum stay in the inventory. It is the only scrub; everything else is structural, so anything the gate catches is a shape nobody has handled.

## Both inputs

`deploy/spdx/<version>/` is the per-recipe tree. The document the publish step actually hands over is the flattened per-image one in `deploy/images/<machine>/`, which no recipe document resembles — it is rooted in a `software_Sbom` and carries both VEX flavours. It is only produced when the image recipe is built.

## Verified on hiagodk, qemuarm64, SPDX 3.0.1 with `SPDX_INCLUDE_VEX = "current"`

| Run | Result |
| -- | -- |
| `deploy/spdx/3.0.1` | 13,091 documents, 31,221 security nodes and 4,652 relationships dropped, 860 name fields redacted, 6,867 symlinks skipped, exit 0 (8s) |
| `deploy/images/avocado-qemuarm64` | 2 documents, 2,001 nodes and 235 relationships dropped, 921 names redacted, exit 0 |
| rootfs document, before → after | 9,543 → 8,614 nodes, 7.4 → 5.9 MB; all 862 `security_*` gone, all 659 packages, all 4,203 files and the `Sbom` root kept |
| `--check` re-run over both outputs | exit 0 — the gate holds on what it emits |

Scope box 5's discrepancy is chased and recorded in the README: of the 4,762 documents carrying an identifier, 4,760 carry a `security_*` node, and the two that do not hold six patch filenames between them. The ticket guessed "probably a patch filename" — confirmed. That residue is why the redaction step exists.

No pruning of `rootElement`: across 13,091 tree documents and both image documents, no node carries `element` or `import` and no `rootElement` entry has ever named a vulnerability. If that changes the gate fails closed.

## Note on `report.py`

`write_report` grows an `indent` argument so `publish` can share the atomic writer instead of restating it. **Default is unchanged**, so `avocado-cve-report.bb:131` and `do_cve_report` are untouched. Not re-run under bitbake; covered by the test suite.

## Not in scope

- The `*.cve`, its `*.json` twin and `*.spdx.tar.zst` sitting beside the image document are assessment and are not filtered here — keeping them off the feed is the publish step's job (ENG-2279). The `.tar.zst` carries ~58 documents that leak on 2.2, so it needs an explicit decision there.
- `packages/` and `packages-staging/` are both filtered and both written; which copy is finalised is a publish-step decision (ENG-2356).

122 tests, `unittest discover` picks the file up with no workflow change.

### DIFF
--- a/meta-avocado-sbom/README
+++ b/meta-avocado-sbom/README
@@ -20,6 +20,26 @@ They are joined on the recipe name. CVEs are stored once per recipe and
 packages only reference it, so a recipe shipping many packages does not
 duplicate its CVE list.
 
+A format, not a tool
+--------------------
+
+The Patched and Ignored subtraction in read_cve_data reads the cve-check JSON
+shape - package[].issue[].status - and nothing else. It does not import, link
+against or extend whatever wrote those files. Keep it that way.
+
+The correlation engine selected to replace in-build cve-check, sbom-cve-check,
+is GPL-2.0 and exports yocto-cve-check-manifest. Pointing --cve-dir at that
+export is a path change, so the engine ships unmodified and the only obligation
+that ever attaches is conveying its public source. Implementing this same
+subtraction as a plugin inside that engine would make it a derivative work
+instead, and shipping it into a customer's on-prem install is a distribution -
+which would put our subtraction under GPL-2.0 too. This layer is Apache-2.0 and
+remains so exactly as long as the boundary holds.
+
+So extend the reader here, never the engine there. If the export turns out to
+be one merged document rather than <PN>_cve.json per recipe, widen the glob in
+read_cve_data; that is still the right side of the line.
+
 Requirements
 ------------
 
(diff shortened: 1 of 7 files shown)


jetm (commented):
Reviewed the filter and its suite against the 3.0.1 and 2.2 fixtures. Nothing blocking.

**What was checked.** The 122-test suite passes at `230310e`. I then mutation-tested the gate itself, since that is what this change freezes - eight mutations, each breaking one property the change claims:

| mutation | caught |
|---|---|
| gate made case-sensitive | yes |
| patch-name redaction removed | yes (20 tests) |
| `leaks()` stops recursing into lists | yes |
| alias-URI refs no longer treated as dead | yes |
| symlinks followed instead of skipped | yes |
| 2.2 `SECURITY` externalRefs kept | yes |
| leaking document written anyway | yes |
| `dropped_ids.discard(None)` removed | **no** |

Seven of eight fail the suite. The identifier-keyed gate holds up particularly well: planting an identifier in an unhandled field, and lowercasing a patch name, both fail closed rather than passing clean.

**The one that survives is a coverage gap, not a defect** - the guard is present and correct, and I am not asking for a code change. Removing `dropped_ids.discard(None)` leaves all 122 green because every fixture gives its `security_*` nodes an `spdxId`, so `None` never enters the set. On a document where a dropped node lacks one, a relationship that omits `from` is then dropped too:

```
with    discard(None): ['urn:pkg:example', 'urn:file:1', 'urn:rel:1']
without discard(None): ['urn:pkg:example', 'urn:file:1']
gate sees a leak either way?  with: False  without: False
```

A `contains` edge disappears from the published inventory and the gate cannot see it - dropping data does not leak an identifier. Worth one fixture with a security node that has no `spdxId`, given the comment above that line already names the hazard.

**Two things I checked and found fine**, so they need no reply: the `Stats` subclass overriding `_NAMES` works because the parent reads `self._NAMES` in `__init__` and `__setattr__`; and a kept node whose `spdxId` embeds a CVE would fail the gate rather than publish, which matches the fail-closed intent stated for `rootElement`.

jetm (approved):

jetm (approved):

jetm:
@nicksinas I've approved this - flagging it to you for the merge.

It's currently BLOCKED despite the approval and green checks. The branch ruleset on `scarthgap` requires CodeQL code-scanning results, and this branch predates CodeQL default setup being enabled on the repo, so CodeQL never ran against it. Repo admin doesn't clear that - the ruleset's only bypass actor is org admin.

The clean fix is a rebase onto `scarthgap` so CodeQL runs on the branch. #307 is in the same state for the same reason.

jetm:
Correction to my comment above: **a rebase will not fix this, and I should not have suggested it.**

The branch's age is not why CodeQL is missing. This PR is opened from a fork (`hiagofranco/meta-avocado`), and CodeQL default setup does not run on fork PRs. Rebasing onto `scarthgap` leaves it a fork PR, so CodeQL still would not run and the `code_scanning` rule would still hold it.

This is already established here rather than new: #294 was merged by admin override in August for exactly this reason, and the note recorded at the time reads "CodeQL never runs on a fork PR into this repo".

So the real options are:

1. **Admin override**, as #294 took.
2. **Move the branch into this repository** and open the PR from a local branch, which is what makes CodeQL run - #323 is from `avocado-linux/meta-avocado` and gets CodeQL plus the three Analyze jobs.

#307 is in the same position.

Everything else in my comment above stands: approved, its own three checks green, `MERGEABLE`, no unresolved threads, and repo admin does not clear the ruleset - the only bypass actor is org admin.

nicksinas (approved):

PR #307, opened 2026-08-25
ENG-2200 + ENG-2203: record which surface every finding occupies

hiagofranco:
The CVE report was feed-scoped — 6,227 packages on qemuarm64 — with no way to tell which of them ship on a device. Every recipe entry now carries a required `scope`:

| `scope` | Means |
| -- | -- |
| `boot-chain` | on the device, installed by no package manager |
| `base-runtime` | ships in the image as an installed package |
| `feed` | built and installable, not in this image |
| `build-only` | never reaches the device as itself |

**A label on the finding, never a filter on the report.** Narrowing to the image manifest is the obvious alternative and is wrong three ways: a feed package is installable, a statically-linked binary carries code from a toolchain recipe that is not installed, and the boot chain is in no manifest.

One taxonomy closes ENG-2200 and ENG-2203, whose carve-outs are the same rule from opposite directions.

## Worth reviewing carefully

`build-only` is the only value asserting a recipe is *absent* from the device, so it is earned: a package-less recipe is host tooling only if named like host tooling. Anything else package-less is a deploy-only target recipe (`imx-atf`, `firmware-*`, a `u-boot` with `PACKAGES = ""`) and scopes `boot-chain`.

The two mechanisms cover different cases, and both are needed. `AVOCADO_CVE_BOOT_CHAIN` matches exact recipe names and is what labels a *packaged* boot-chain recipe; it is meant to be overridden per machine. The fallback handles the package-less ones a vendor layer adds that no default list has heard of. Neither subsumes the other.

The naming rule is empirical. A first version scoped `gcc-source-13.4.0` and `libgcc-initial` as `boot-chain` on a real build; it was then checked against all 94 package-less recipes across the three cve-check statuses, which added the `-initial` and `gcc-source-<PV>` families.

## Contract

Stays version `1` — the compatibility rules do not name a new *required* field on a recipe entry, and the README records why it was made at 1 anyway. Scope and `packaged` are independent, so **scope does not partition `packaged_cves`**. Per-scope totals are deliberately not counters; only `manifests_read` is added.

## Verification

109 tests pass including the jsonschema-backed schema tests; `bitbake -c cve_report` succeeds; verified against the real qemuarm64 tree at all three statuses:

| status | base-runtime | boot-chain | feed | build-only |
| -- | -- | -- | -- | -- |
| Unpatched | 10 | 1 | 14 | 26 |
| Patched | 42 | 3 | 72 | 94 |
| Ignored | 11 | 0 | 12 | 21 |

`unscanned_recipes: 11` / `no_cve_record_recipes: 97` identical across four runs — the coverage denominator does not move.

## Known gaps, deliberate

- ENG-2200 criterion 3 is answered in the layer README, not in the report document.
- ENG-2200 criterion 2 (statically-linked runtimes attributed to the dependent package) ships as README principle, not code — nothing built here ships such a runtime yet. `scope` is per-recipe, so encoding it later is a **schema change**.
- Verified since on i.MX95 (a second BSP, 93 build-only recipes): nothing is mislabelled `build-only`, and the one genuinely package-less firmware recipe, `imx-boot-firmware-files`, scopes `boot-chain` through the fallback. Note that `u-boot-imx`, `imx-atf`, `imx-boot` and `imx-system-manager` all produce `-dbg`/`-dev` packages, so they scope `feed` and need `AVOCADO_CVE_BOOT_CHAIN` set in the i.MX machine config — the fallback is not the mechanism for them.

### DIFF
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -597,12 +597,24 @@ def default_paths(tmpdir, machine=None):
 def default_manifests(tmpdir, machine=None):
     """Image manifests under a build's TMPDIR.
 
+    IMAGE_LINK_NAME for the images that ship, so a machine's flashing images
+    and the stale timestamped manifests beside each symlink stay out - both
+    would scope base-runtime.
+
     Kept out of default_paths so its return shape stays frozen: an empty result
     is a normal outcome here, not the failure every branch of default_paths
     raises for.
     """
-    images = os.path.join(tmpdir, "deploy", "images", machine or "*")
-    return sorted(glob.glob(os.path.join(images, "*.manifest")))
+    root = os.path.join(tmpdir, "deploy", "images")
+    return sorted(
+        path
+        for images in glob.glob(os.path.join(root, machine or "*"))
+        for path in glob.glob(
+            os.path.join(
+                images, "avocado-image-*-%s.manifest" % os.path.basename(images)
+            )
+        )
+    )
 
 def main():
     import argparse
@@ -637,7 +649,8 @@ def main():
         action="append",
         default=[],
         help="image manifest naming the base runtime, repeatable; overrides "
-        "the ${DEPLOY_DIR}/images/${MACHINE}/*.manifest default",
+        "the ${DEPLOY_DIR}/images/${MACHINE}/avocado-image-*-${MACHINE}.manifest "
+        "default",
     )
     parser.add_argument(
         "--boot-chain",
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -104,7 +104,7 @@ def read_manifests(manifest_paths, stats):
     """
     installed = set()
 
-    for path in sorted(manifest_paths):
+    for path in manifest_paths:
         try:
             with open(path, errors="ignore") as f:
                 lines = f.read().splitlines()
@@ -131,11 +131,9 @@ SCOPES = ("boot-chain", "base-runtime", "feed", "build-only")
 # and libgcc-initial matched none of the obvious suffixes.
 def _is_host_tooling(recipe):
     return (
-        recipe.startswith("nativesdk-")
-        or recipe.endswith(("-native", "-cross", "-crosssdk", "-initial"))
+        recipe.endswith(("-native", "-cross", "-crosssdk", "-initial", "-source"))
         or "-cross-" in recipe
         or "-source-" in recipe
-        or recipe.endswith("-source")
     )
 
 def _scope(recipe, package_names, installed, boot_chain):
@@ -152,6 +150,11 @@ def _scope(recipe, package_names, installed, boot_chain):
     """
     if recipe in boot_chain:
         return "boot-chain"
+    # Before package membership, unlike the rest of host tooling: PKGDATA_DIR_SDK
+    # packages nativesdk recipes, so they are packaged and would read "feed" -
+    # installable on a device, which is what the prefix rules out.
+    if recipe.startswith("nativesdk-"):
+        return "build-only"
     if any(name in installed for name in package_names):
         return "base-runtime"
     if package_names:
@@ -647,8 +650,8 @@ def main():
         action="append",
         default=[],
         help="image manifest naming the base runtime, repeatable; overrides "
-        "the ${DEPLOY_DIR}/images/${MACHINE}/avocado-image-{rootfs,initramfs}-"
-        "*.manifest default",
+        "the avocado-image-rootfs-*.manifest and avocado-image-initramfs-*"
+        ".manifest defaults under ${DEPLOY_DIR}/images/${MACHINE}",
     )
     parser.add_argument(
         "--boot-chain",
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -24,6 +24,7 @@ class Stats(dict):
         "ignored_cves",
         "patched_cves",
         "unknown_status_cves",
+        "manifests_read",
     )
 
     def __init__(self):
@@ -94,6 +95,73 @@ def read_pkgdata(pkgdata_dirs, stats):
 
     stats.packages = len(packages)
     return packages, recipe_versions
+
+def read_manifests(manifest_paths, stats):
+    """Package names installed by the given image manifests.
+
+    A manifest line is "PKG ARCH VERSION". An absent one is not an error - see
+    _scope for why the report degrades rather than fails.
+    """
+    installed = set()
+
+    for path in sorted(manifest_paths):
+        try:
+            with open(path, errors="ignore") as f:
+                lines = f.read().splitlines()
+        except OSError:
+            continue
+
+        names = {fields[0] for fields in (line.split() for line in lines) if fields}
+        if not names:
+            # Truncated: counting it would report the image as scoped by it.
+            continue
+        installed.update(names)
+        stats.manifests_read += 1
+
+    return installed
+
+# Most privileged first. A label on every finding, never a filter on the
+# report - the README's "Scope" section has why.
+SCOPES = ("boot-chain", "base-runtime", "feed", "build-only")
+
+# Naming is the only signal the report's inputs carry; the real one is whether
+# the recipe inherits deploy, which needs a per-recipe datastore the standalone
+# path lacks. Checked against all 94 package-less recipes on a qemuarm64 world
+# build - the -source and -initial families are there because gcc-source-<PV>
+# and libgcc-initial matched none of the obvious suffixes.
+def _is_host_tooling(recipe):
+    return (
+        recipe.startswith("nativesdk-")
+        or recipe.endswith(("-native", "-cross", "-crosssdk", "-initial"))
+        or "-cross-" in recipe
+        or "-source-" in recipe
+        or recipe.endswith("-source")
+    )
+
+def _scope(recipe, package_names, installed, boot_chain):
+    """Which surface a recipe occupies.
+
+    Boot chain is checked first and by name, not package membership: u-boot and
+    trusted-firmware-a reach the device through avocado-img-bootfiles, which
+    scrapes DEPLOY_DIR_IMAGE, so no manifest names them.
+
+    "build-only" is the only value asserting a recipe is NOT on the device, so
+    it is earned rather than defaulted to. Read no manifest and everything
+    packaged reads "feed" - over-reporting the device, which is the safe way to
+    be wrong.
+    """
+    if recipe in boot_chain:
+        return "boot-chain"
+    if any(name in installed for name in package_names):
+        return "base-runtime"
+    if package_names:
+        return "feed"
+    if _is_host_tooling(recipe):
+        return "build-only"
+    # Deploy-only target recipe - imx-atf, imx-boot, firmware-*, a u-boot with
+    # PACKAGES = "". The README's "packaged" section says their CVEs are real
+    # and on the device, so build-only would hide them.
+    return "boot-chain"
 
 CVE_FIELDS = (
     "id",
@@ -327,7 +395,15 @@ def packages_digest(packages):
 
 REPORT_VERSION = "1"
 
-def build_report(cve_dir, pkgdata_dirs, machine=None, status="Unpatched", summary=True):
+def build_report(
+    cve_dir,
+    pkgdata_dirs,
+    machine=None,
+    status="Unpatched",
+    summary=True,
+    manifest_paths=(),
+    boot_chain=(),
+):
     if status not in CVE_STATUSES:
         raise ValueError(
             "unknown cve-check status %r; expected one of %s"
@@ -340,6 +416,23 @@ def build_report(cve_dir, pkgdata_dirs, machine=None, status="Unpatched", summar
         cve_dir, recipe_versions, stats, status=status, summary=summary
     )
 
+    installed = read_manifests(manifest_paths, stats)
+    declared_boot_chain = set(boot_chain)
+    recipe_packages = {}
+    for name, pkg in packages.items():
+        recipe_packages.setdefault(pkg["recipe"], []).append(name)
+
+    # Scope totals are deliberately not counters: every entry carries its own,
+    # so an aggregate cannot drift from what the entries say.
+    for name, entry in recipes.items():
+        entry["scope"] = _scope(
+            name, recipe_packages.get(name, ()), installed, declared_boot_chain
+        )
+
+    # Scope is a field, not a filter, so this denominator does not move: both
+    # lists stay derived from pkgdata, and reading a manifest leaves them
+    # identical.
+    #
     # Both lists are scoped to recipes that shipped a package, so together with
     # the recipes scanned and found in the NVD they partition that set: a
     # consumer can account for every shipped recipe exactly once.
@@ -374,6 +467,7 @@ def build_report(cve_dir, pkgdata_dirs, machine=None, status="Unpatched", summar
             "ignored_cves": stats.ignored_cves,
             "patched_cves": stats.patched_cves,
             "unknown_status_cves": stats.unknown_status_cves,
+            "manifests_read": stats.manifests_read,
         },
         "recipes": recipes,
         "packages": packages,
@@ -500,6 +594,16 @@ def default_paths(tmpdir, machine=None):
     pkgdata_dirs.sort(key=lambda d: "-linux" in os.path.basename(d))
     return cve_dir, pkgdata_dirs
 
+def default_manifests(tmpdir, machine=None):
+    """Image manifests under a build's TMPDIR.
+
+    Kept out of default_paths so its return shape stays frozen: an empty result
+    is a normal outcome here, not the failure every branch of default_paths
+    raises for.
+    """
+    images = os.path.join(tmpdir, "deploy", "images", machine or "*")
+    return sorted(glob.glob(os.path.join(images, "*.manifest")))
+
 def main():
     import argparse
     import sys
@@ -528,6 +632,19 @@ def main():
     parser.add_argument(
         "--no-summary", action="store_true", help="drop CVE description text"
     )
+    parser.add_argument(
+        "--manifest",
+        action="append",
+        default=[],
+        help="image manifest naming the base runtime, repeatable; overrides "
+        "the ${DEPLOY_DIR}/images/${MACHINE}/*.manifest default",
+    )
+    parser.add_argument(
+        "--boot-chain",
+        default="",
+        help="space-separated recipes that run on the device but are "
+        "installed by no package manager, e.g. 'u-boot trusted-firmware-a'",
+    )
     parser.add_argument("-o", "--output", required=True)
     args = parser.parse_args()
 
@@ -539,6 +656,9 @@ def main():
             parser.error(str(e))
     cve_dir = args.cve_dir or cve_dir
     pkgdata_dirs = args.pkgdata_dir or pkgdata_dirs
+    manifests = args.manifest or (
+        default_manifests(args.tmpdir, args.machine) if args.tmpdir else []
+    )
 
     if not cve_dir or not pkgdata_dirs:
         parser.error("need --tmpdir, or both --cve-dir and --pkgdata-dir")
@@ -555,6 +675,8 @@ def main():
         machine=args.machine,
         status=args.status,
         summary=not args.no_summary,
+        manifest_paths=manifests,
+        boot_chain=args.boot_chain.split(),
     )
 
     if not stats.cve_files:
@@ -605,7 +727,16 @@ def main():
         ),
         file=sys.stderr,
     )
+    if not stats.manifests_read:
+        print(
+            "  no image manifest read: every packaged recipe scopes 'feed'. "
+            "Pass --manifest, or run after the image tasks have written "
+            "${DEPLOY_DIR}/images/%s/*.manifest." % (args.machine or "<machine>"),
+            file=sys.stderr,
+        )
+
     for key in (
+        "manifests_read",
         "unscanned_recipes",
         "no_cve_record_recipes",
         "stale_dropped",
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -595,11 +595,12 @@ def default_paths(tmpdir, machine=None):
     return cve_dir, pkgdata_dirs
 
 def default_manifests(tmpdir, machine=None):
-    """Image manifests under a build's TMPDIR.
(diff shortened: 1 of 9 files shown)


meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
```diff
@@ -31,6 +31,15 @@ AVOCADO_CVE_REPORT_STRICT ?= "1"
 # results by default, because the two are read together and a build that moves
 # one moves the other.
 AVOCADO_CVE_REPORT_OPTOUT_DIR ?= "${CVE_CHECK_DIR}"
+# Image manifests naming the base runtime. Matching nothing is not an error:
+# this task is ordered only before do_build, so the image tasks need not have
+# run. Everything packaged then scopes "feed", over-reporting the device rather
+# than under-reporting it.
+AVOCADO_CVE_REPORT_MANIFESTS ?= "${DEPLOY_DIR_IMAGE}/*.manifest"
```
jetm: **`*.manifest` matches more than the runtime images, so non-runtime packages scope `base-runtime`.** The Jetson Orin Nano devkit build has six manifests in `DEPLOY_DIR_IMAGE`, not the two this default assumes:

```
avocado-image-rootfs-jetson-orin-nano-devkit.manifest              136 pkgs
avocado-image-initramfs-jetson-orin-nano-devkit.manifest
tegra-initrd-flash-initramfs-avocado-jetson-orin-nano-devkit.manifest  99 pkgs
tegra-espimage-avocado-jetson-orin-nano-devkit.manifest             1 pkg
tegra-initrd-flash-initramfs-...-20260815021433.manifest
tegra-espimage-...-20260814192228.manifest
```

`read_manifests` unions all six into `installed`, so any recipe packaged into the flashing initramfs or the ESP image scopes `base-runtime`. Packages that are in the flash initramfs and *not* in the rootfs include `gptfdisk`, `e2fsprogs-mke2fs`, and a dozen `kernel-module-*` (`nvme`, `pcie-tegra194`, `pwm-tegra`, `ina3221`, ...). Their CVEs would be labelled as shipping on the device when they only ever run on the host during provisioning.

That is the direction the design says it does not want to be wrong in. The README argues the degradation is safe because it falls back to `feed`, which over-reports; this fails the other way, marking things `base-runtime` that are not in the runtime image at all - and `base-runtime` is the value a consumer will filter on to decide what is actually fielded.

It also inflates the counter: `manifests_read` reads 6 here rather than 2, and the schema describes it as "image manifests that named at least one package" - which the single-package ESP manifest technically satisfies.

qemuarm64 and imx95 both produce exactly the two expected manifests, which is why the verification runs on those machines did not surface it.

Secondary, same default, mechanism rather than measurement: the timestamped names are distinct inodes from the unstamped ones, and nothing cleans them, so repeated builds leave older stamped manifests matching the glob alongside the current pair.

Narrowing the default to the images the report means - something like `${DEPLOY_DIR_IMAGE}/avocado-image-*.manifest` - would fix both, and keeps the per-machine override for boards whose runtime image is named differently.

jetm (commented):
Reviewed the scope function, the manifest reader, the schema change and the suite. 109 tests pass at `7fc7f92`. One blocking comment inline.

**The blocking one in short:** the `AVOCADO_CVE_REPORT_MANIFESTS` default globs every `*.manifest` in `DEPLOY_DIR_IMAGE`, and on Jetson that is six files rather than two - the flashing initramfs and the ESP image among them. Packages that exist only in the flashing initramfs (`gptfdisk`, `e2fsprogs-mke2fs`, a dozen `kernel-module-*`) then scope `base-runtime`. qemuarm64 and imx95 each produce exactly two manifests, which is why neither verification run showed it.

**What held up well.** `build-only` being earned rather than defaulted is the right shape, and the fallback to `boot-chain` for package-less recipes is what keeps `imx-atf` and `imx-boot-firmware-files` visible - a real hazard, given `AVOCADO_CVE_BOOT_CHAIN` matches exact recipe names and NXP's bootloader is `u-boot-imx`. Checking the boot chain by name before package membership is also what makes the label survive a `PACKAGES = ""` bootloader.

`test_the_coverage_denominator_does_not_move` is the test I would have asked for if it were not already there: it builds the same tree twice and asserts both coverage lists, the findings set, and `packages_digest` are identical, which is what makes "a field, not a filter" checkable rather than asserted.

**On keeping the contract at version 1** - the reasoning convinced me, and it is worth the paragraph it got. An optional `scope` really is indistinguishable from a producer that read no manifest, since the safe default for an absent value (`feed`) is exactly what that producer would emit. Requiring it forces the ambiguity into `manifests_read` where a consumer can see it. Consumers of version 1 read strictly less than producers must now write, so nothing downstream breaks.

**Not raised as findings**, having checked them: `verify.py` spelling `SCOPES` out rather than importing `report.SCOPES` is deliberate and correct for a checker that must not take the producer's word for the vocabulary; and the `-source`/`-initial` families in `_is_host_tooling` look arbitrary until you see they were derived from the 94 package-less recipes on a real build rather than guessed.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
hiagofranco: Fixed in 20520919. Default is now `${DEPLOY_DIR_IMAGE}/avocado-image-*-${MACHINE}.manifest`, in the recipe and in `default_manifests()` for the standalone CLI.

`avocado-image-*` drops the flashing and ESP manifests. The `-${MACHINE}` suffix is `IMAGE_LINK_NAME`, this build's symlink — the stale timestamped files end in `-<timestamp>.manifest` and no longer match, so the secondary point is covered by the same glob. `manifests_read` goes back to counting the shipping images only.

`default_manifests()` had no test; added `DefaultManifestsTest`, which asserts both the tegra pair and an older build's timestamped manifests stay out. 111 tests pass.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
```diff
@@ -31,6 +31,19 @@ AVOCADO_CVE_REPORT_STRICT ?= "1"
...
 AVOCADO_CVE_REPORT_OPTOUT_DIR ?= "${CVE_CHECK_DIR}"
+# Image manifests naming the base runtime. Matching nothing is not an error:
+# this task is ordered only before do_build, so the image tasks need not have
+# run. Everything packaged then scopes "feed", over-reporting the device rather
+# than under-reporting it.
+#
+# Narrow, because anything extra scopes base-runtime: *.manifest also matches
+# a machine's flashing images (Jetson's tegra-initrd-flash-initramfs) and, via
+# the timestamped names beside each IMAGE_LINK_NAME symlink, every older build.
+AVOCADO_CVE_REPORT_MANIFESTS ?= "${DEPLOY_DIR_IMAGE}/avocado-image-*-${MACHINE}.manifest"
```
jetm: **This glob matches nothing on any avocado machine, so `manifests_read` is always 0 and no recipe ever scopes `base-runtime`.** The manifests are named with `MACHINE_SHORT_NAME`, not `MACHINE`.

`meta-avocado/conf/distro/avocado.inc:20` sets it distro-wide:

```
MACHINE_SHORT_NAME = "${@d.getVar('MACHINE').replace('avocado-','')}"
```

and `meta-avocado/recipes-avocado/images/avocado-image-rootfs.bb:11` (same in `avocado-image-initramfs.bb:13`) names the image with it:

```
IMAGE_MACHINE_SUFFIX = "-${MACHINE_SHORT_NAME}"
IMAGE_NAME = "${IMAGE_BASENAME}${IMAGE_MACHINE_SUFFIX}"
```

So on a real build the deploy dir is `avocado-qemuarm64` while the manifests inside it are `avocado-image-rootfs-qemuarm64.manifest`. Checked against a completed qemuarm64 build:

```
$ ls .../images/avocado-qemuarm64/avocado-image-*-avocado-qemuarm64.manifest
no matches found
$ ls .../images/avocado-qemuarm64/avocado-image-*-qemuarm64.manifest
avocado-image-initramfs-qemuarm64.manifest
avocado-image-rootfs-qemuarm64.manifest
```

Same on `avocado-raspberrypi5` (`avocado-image-rootfs-raspberrypi5.manifest`) and on the jetson build this default was written for.

The jetson listing also shows the `-${MACHINE}` suffix is doing the opposite of its intent, because the naming is inverted from what it assumes:

```
avocado-image-rootfs-jetson-orin-nano-devkit.manifest          <- want, SHORT name
avocado-image-initramfs-jetson-orin-nano-devkit.manifest       <- want, SHORT name
tegra-espimage-avocado-jetson-orin-nano-devkit.manifest        <- exclude, full MACHINE
tegra-initrd-flash-initramfs-avocado-jetson-orin-nano-devkit.manifest
```

The four tegra manifests carry the full MACHINE; the two you want carry the short name. `avocado-image-*` alone already excludes the tegra pair, so `-${MACHINE}` adds no filtering and removes the two real matches.

The stated reason for the suffix does not apply here either. `verify.py` says "The timestamped file beside each IMAGE_LINK_NAME symlink is never pruned ... hence -${MACHINE}". But `avocado-image-rootfs.bb:10` and `avocado-image-initramfs.bb:11` both set `IMAGE_LINK_NAME = ""`, and `IMAGE_NAME` is overridden to `${IMAGE_BASENAME}${IMAGE_MACHINE_SUFFIX}` with `IMAGE_NAME_SUFFIX = ""` - there is no symlink and no timestamped `avocado-image-*` manifest to guard against. The timestamped files in that listing are the tegra ones, already excluded by the prefix.

The failure is quiet in the safe direction - the recipe warns and everything degrades to `feed` - but the feature is inert on every machine rather than on none.

`${MACHINE_SHORT_NAME}` looks like the fix. Worth confirming it holds for a machine whose name does not start with `avocado-`, where the two are equal and the glob works either way.

meta-avocado-sbom/lib/avocado_sbom/report.py
```diff
@@ -500,6 +594,28 @@ def default_paths(tmpdir, machine=None):
...
+    is a normal outcome here, not the failure every branch of default_paths
+    raises for.
+    """
+    root = os.path.join(tmpdir, "deploy", "images")
+    return sorted(
+        path
+        for images in glob.glob(os.path.join(root, machine or "*"))
+        for path in glob.glob(
+            os.path.join(
+                images, "avocado-image-*-%s.manifest" % os.path.basename(images)
```
jetm: **The standalone CLI has the same defect as the recipe, from the same assumption.** `os.path.basename(images)` is the deploy directory name, which is `MACHINE` - so this builds `avocado-image-*-avocado-qemuarm64.manifest` and matches nothing, exactly as the recipe's `${MACHINE}` glob does.

Ran this function from the PR head against the real jetson layout:

```
machine dir (MACHINE):   avocado-jetson-orin-nano-devkit
manifest suffix (SHORT): jetson-orin-nano-devkit
glob built:              avocado-image-*-avocado-jetson-orin-nano-devkit.manifest

default_manifests() returned 0 path(s)
expected 2: avocado-image-{rootfs,initramfs}-jetson-orin-nano-devkit.manifest
```

Worth noting both sites need the fix, since the CLI cannot read `MACHINE_SHORT_NAME` from a datastore the way the recipe can - it only has the directory name. Deriving it by stripping a leading `avocado-` would mirror `avocado.inc:20`, but it hardcodes the distro prefix into the library; taking the short name as an argument from the recipe, or globbing `avocado-image-*.manifest` within the already-machine-scoped directory, both avoid that.

meta-avocado-sbom/tests/test_report.py
```diff
@@ -347,5 +532,44 @@ def test_a_bad_marker_does_not_hide_a_good_one(self):
         self.assertEqual(list(declared), ["good"])
         self.assertEqual(unreadable, 1)
 
+class DefaultManifestsTest(unittest.TestCase):
+    """Anything the default glob over-reads scopes base-runtime."""
+
+    def setUp(self):
+        self.tmp = tempfile.mkdtemp()
+        self.addCleanup(shutil.rmtree, self.tmp)
+        self.images = os.path.join(self.tmp, "deploy", "images", "jetson")
```
jetm: **This fixture cannot catch the glob defect, because it makes `MACHINE` and the manifest suffix the same string.** The machine directory is `jetson` and the manifests are `avocado-image-rootfs-jetson.manifest`, so `avocado-image-*-jetson.manifest` matches. In a real build they are never equal - the directory is `avocado-jetson-orin-nano-devkit` and the suffix is `jetson-orin-nano-devkit`.

The fixture already encodes the real asymmetry for the tegra files, which it names `tegra-...-avocado-jetson.manifest` with the prefix, while naming the avocado images without it. That is the mismatch the test would need to reproduce on both.

Renaming the directory to `avocado-jetson` and leaving the manifests as `-jetson` reproduces production naming, and the test fails - which is the behaviour to assert once the glob is fixed. `test_without_a_machine_every_machine_dir_is_globbed` has the same shape and would need the same treatment.

jetm (commented):
Re-reviewed after 20520919. The narrowing from `*.manifest` to `avocado-image-*` does fix what I raised - the tegra flashing and ESP manifests are excluded, and `DefaultManifestsTest` is the right thing to have added.

The `-${MACHINE}` half of it does not work, and it takes the rest of the feature with it. Avocado images are named with `MACHINE_SHORT_NAME` (`avocado.inc:20` strips the `avocado-` prefix distro-wide; `avocado-image-rootfs.bb:11` sets `IMAGE_MACHINE_SUFFIX` from it), so the glob resolves to `avocado-image-*-avocado-<machine>.manifest` while the files on disk are `avocado-image-rootfs-<machine>.manifest`. It matches nothing on qemuarm64, raspberrypi5 or the jetson build the default was written for, which means `manifests_read` is 0 everywhere and no recipe ever scopes `base-runtime`.

The jetson listing shows the naming is inverted from what the suffix assumes: the two shipping manifests carry the short name, and it is the four tegra ones that carry the full MACHINE. `avocado-image-*` alone already excludes those, so the suffix filters nothing and drops both real matches. The stated reason for it - the stale timestamped file beside the `IMAGE_LINK_NAME` symlink - does not arise for these images either, since both recipes set `IMAGE_LINK_NAME = ""` and override `IMAGE_NAME` to carry no timestamp.

I ran `default_manifests()` from this head against the real jetson layout and it returned 0 of 2; details are in the inline comments, along with the same defect in the standalone CLI path and why the test fixture cannot currently catch either.

Everything else held up. The scope fall-through reads correctly to me - deploy-only target recipes landing on `boot-chain` rather than `build-only`, and `build-only` being earned rather than defaulted. Degrading toward `feed` rather than `base-runtime` is the right direction for a CVE report, and I reached the same conclusion independently while triaging this week, so the reasoning in the README is corroborated rather than just asserted. `read_manifests` skipping an empty manifest without counting it is right, and the schema's `scope` enum matches `SCOPES`.

Not raising the contract-version question as blocking: adding a required field while `version` stays at 1 is called out in the PR itself, and it is a project decision rather than a defect in this change.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
hiagofranco: Confirmed and fixed in 1548485a. `MACHINE_SHORT_NAME` is right, and the glob was inert everywhere.

Went one step further than `avocado-image-*.manifest` within the machine dir: `avocado-image-container.bb` is `inherit image` with no naming override, so it deploys `avocado-image-container-${MACHINE}-<timestamp>.manifest` plus the `IMAGE_LINK_NAME` symlink into the same directory. That is the SDK dev container, not on the device — a prefix-only glob scopes its packages `base-runtime`. It is also the only `avocado-image-*` here that produces the timestamped duplicate the old suffix was meant to guard against.

So the default names the two images that ship and globs only the machine suffix:

```
AVOCADO_CVE_REPORT_MANIFESTS ?= "${DEPLOY_DIR_IMAGE}/avocado-image-rootfs-*.manifest ${DEPLOY_DIR_IMAGE}/avocado-image-initramfs-*.manifest"
```

The suffix glob answers your last question too — a machine whose name does not start with `avocado-` has `MACHINE_SHORT_NAME == MACHINE` and matches either way, without the library or the recipe having to know which.

meta-avocado-sbom/lib/avocado_sbom/report.py
hiagofranco: Fixed in the same commit, same shape as the recipe — the CLI needs neither `MACHINE_SHORT_NAME` nor the distro prefix:

```python
for image in ("avocado-image-rootfs", "avocado-image-initramfs")
for path in glob.glob(os.path.join(images, "%s-*.manifest" % image))
```

`os.path.basename(images)` is gone. The directory is already machine-scoped, so globbing the suffix costs nothing, and naming the two images keeps `avocado-image-container` out — which is the case that made `avocado-image-*.manifest` in the machine dir not quite enough on its own.

meta-avocado-sbom/tests/test_report.py
hiagofranco: Right — the fixture made the two strings equal, so it proved nothing. Directory is now `avocado-jetson` with the manifests at `-jetson`, which is production naming, and the old glob fails it: it returns `avocado-image-container-avocado-jetson.manifest` and neither shipping image.

The container pair is in the fixture now as well, covering both the full-MACHINE name and the timestamped file beside its symlink. `test_without_a_machine_every_machine_dir_is_globbed` got the same directory rename; its last assertion went away, since it was checking that the suffix is the dir's own MACHINE and there is no suffix anchor any more.

111 tests pass.

jetm (commented):
Re-reviewed at 1548485a. All three findings are fixed, and nothing blocking came back.

Verified rather than read:

- `default_manifests()` from this head, against the real jetson layout plus the container pair, returns exactly the two shipping manifests and excludes both the tegra flashing images and the container symlink/timestamped file. The same function at 20520919 returned 0 of 2 on that layout.
- The recipe glob resolves correctly on three real deploy directories - qemuarm64, raspberrypi5 and jetson-orin-nano-devkit - matching `avocado-image-{rootfs,initramfs}-<short>.manifest` and nothing else.
- 111 tests pass at this head.
- The new fixture is a genuine regression guard, not just a rename. Running the previous head's `default_manifests()` against it returns `avocado-image-container-avocado-jetson.manifest` and neither shipping image, so the old code fails the new test on exactly the worst case: the SDK container scoped `base-runtime` while both real images are missed.

The container catch is a better fix than what I suggested. I offered globbing `avocado-image-*.manifest` inside the machine directory as one option and that would have been wrong - `avocado-image-container.bb` is `inherit image` with no naming override, so it deploys under the default `IMAGE_NAME` with the full MACHINE and a timestamp, straight into the same directory. Confirmed in the recipe. Naming the two images avoids it and drops the `MACHINE_SHORT_NAME` dependency at the same time, which is a cleaner shape than the fix I proposed.

Checked that naming exactly two images loses nothing today: `avocado-image-var` is the only other avocado image that reaches a device, and it deploys a `.btrfs` with no manifest at all, so there is nothing for a glob to miss. A machine that later ships a third on-device image would need the override the variable already documents.

The comments no longer claim the `IMAGE_LINK_NAME` symlink is what the suffix guards against, which is right - `avocado-image-rootfs.bb` and `avocado-image-initramfs.bb` both set `IMAGE_LINK_NAME = ""`, and the container is the only `avocado-image-*` here that produces that pair.

jetm (approved):

hiagofranco:
Pushed `651aeab8`. Self-review of the branch against ENG-2200/ENG-2203 and the layer README, after the glob fix — one real defect, one test gap, the rest documentation catching up with the code. 113 tests pass.

**`nativesdk-*` was scoping `feed`, not `build-only`.**

`_scope()` only reached `_is_host_tooling()` when a recipe had no packages, but `do_cve_report` passes `PKGDATA_DIR_SDK` alongside `PKGDATA_DIR`, and nativesdk recipes have `runtime-reverse` entries there. So they came out `packaged: true` → `feed` — "built and installable" host tooling, against ENG-2203's table, which names `nativesdk` under `build-only`.

Fixed by deciding the prefix before package membership:

```python
if recipe in boot_chain:
    return "boot-chain"
if recipe.startswith("nativesdk-"):
    return "build-only"
if any(name in installed for name in package_names):
    return "base-runtime"
```

Only this family needs the earlier check — `-native` and `-cross` produce no package in any pkgdata dir, so they still fall through to `_is_host_tooling()` where the package-less rule keeps `build-only` earned. `nativesdk-` came out of `_is_host_tooling()`, where it is now unreachable. `test_a_packaged_nativesdk_recipe_is_still_build_only` gives the recipe a package and asserts the label; against the old ordering it returns `feed`.

Worth a second opinion: this is the one change here that moves a label, and it moves it toward the narrow direction — `build-only` asserts absence. The argument is that `nativesdk-` is reserved by OE for the SDK host, so the prefix is a stronger statement than package membership, not a heuristic like the suffix families.

**The scope vocabulary was declared three times and bound nowhere.** `report.SCOPES`, `verify.SCOPES` and the schema `enum` — the README's compatibility policy says a test asserts the schema and the checker agree, and `test_verify.py` already does that for statuses and counters, but not for `scope`. Added `test_schema_scopes_match_the_generator_and_the_checker`. `verify.SCOPES` stays spelled out for the reason its comment gives; the test is what makes that safe rather than hopeful.

**Documentation that had drifted from the code:**

- The default was written `avocado-image-{rootfs,initramfs}-*.manifest` in the README and in `--manifest`'s help. Python's `glob` has no brace expansion, so copied into `local.conf` verbatim it matches nothing — and matching nothing is not an error, so it would degrade silently to all-`feed`. My wording from `1548485a`; both now give the two patterns literally.
- `manifests_read` joined `CARRIED` in this PR, but the README still said "Five counters" and "the five underivable counters".
- `_is_host_tooling` matches a `-source` suffix and `-cross-`/`-source-` infixes that the README's list did not mention. Documented rather than narrowed — the patterns came from the 94 package-less recipes on a real build, and cutting them would put `gcc-source-<PV>` and the `-cross-` families back on `boot-chain`.
- ENG-2200's "state which denominator moves" was answered in the README, a comment and a test, but not in the document. The `unscanned_recipes` and `no_cve_record_recipes` schema descriptions now say the denominator is pkgdata and that scope does not move it, so a consumer reading only the report can see it.

**Two findings deliberately left:**

- `read_manifests` swallows an unreadable manifest with `except OSError: continue`, while everything around it counts unreadable inputs (`cve_files_unreadable`, `pkgdata_unreadable`). One unreadable manifest of two leaves `manifests_read: 1`, so the recipe's warning never fires and half the base runtime scopes `feed` invisibly. The consistent fix is another counter, which reopens the version-1 contract question — happy to add it if you want it in this PR rather than after.
- `manifests_read` names a count of "manifests that named at least one package". Accurate, but it is frozen contract now; the schema description carries the precision instead.

jetm:
@nicksinas @mobileoverlord Need review or approval on this. I have approved it, but I cannot merge it due to permissions

jetm (approved):

mobileoverlord (approved):

PR #308, opened 2026-08-25
nxp: fix the two things that stop i.MX machines building

hiagofranco:
Two independent regressions, both blocking i.MX builds. Found while trying to build `imx95-frdm`; neither is caught by CI because machine builds are opt-in via the `build-<machine>` label.

## 1. `imx95-frdm` cannot be checked out

`bde07247` migrated `kas/extra/deepx.yml` into `kas/feature/ai.yml` and deleted all 16 `kas/extra/` files. It updated `grinn-astra-1680-sbc.yml`, the other DeepX board, but missed `imx95-frdm.yml` — which was left as the only reference to `kas/extra/` in the tree:

```
ERROR - Configuration file not found: .../meta-avocado/kas/extra/deepx.yml
```

## 2. Variscite bbappends dangle on every other NXP board

```
ERROR: No recipes in default available for:
  .../meta-avocado-nxp/recipes-bsp/u-boot/u-boot-variscite.bbappend
  .../meta-avocado-nxp/recipes-kernel/linux/linux-variscite-imx_%.bbappend
```

`u-boot-variscite` and `linux-variscite-imx` live in `meta-variscite-bsp-imx` (collection `variscite-bsp`), which only `kas/vendor/variscite.yml` pulls in — so both appends dangle on imx95-frdm, imx8mp-evk, imx93-frdm, imx91-frdm, imx93-evk and ucm-imx8m-plus.

This is a regression of a bug already fixed once. `7ab6b7ee` moved the CompuLab appends under `dynamic-layers/` for exactly this reason and documented it in `meta-avocado-nxp/conf/layer.conf`:

> Their bbappends live under `dynamic-layers/` (NOT matched by the `recipes-*/` BBFILES glob above) and are loaded only when that collection is present, so they don't dangle ("No recipes available") on the other meta-avocado-nxp boards that share this layer (e.g. imx8mp-evk).

Variscite support landed three weeks later without following it.

The kernel bbappend shares its config fragments with the in-tree `linux-imx` append, so it reaches them via `AVOCADO_NXP_KERNEL_FILESDIR` instead of `${THISDIR}/files` — the same treatment `linux-compulab` already gets.

## Verification

On `imx95-frdm` with `BB_DANGLINGAPPENDS_WARNONLY` unset: **3653 recipes parse, no dangling variscite appends**, and `meta-deepx-m1` checks out via `kas/feature/ai.yml`. Every move is a pure rename except the `FILESEXTRAPATHS` line.

## Still broken after this, deliberately out of scope

Two appends still dangle on mx9x, both pre-existing and neither fixable in this repo:

- `meta-imx/meta-imx-bsp/.../openssl_3.2.%.bbappend` against oe-core's `openssl_3.5.7`. Worth attention separately: it silently drops i.MX cryptodev and kTLS offload (`PACKAGECONFIG:append:imx-nxp-bsp = " cryptodev-linux"`, `EXTRA_OECONF:append = " enable-ktls"`).
- `gstreamer1.0-plugins-imx_%.bbappend` — the meta-freescale recipe is `COMPATIBLE_MACHINE = mx6/7/8`, so it is skipped on mx9x, while the append is still needed on mx8 boards. There is no clean bitbake idiom for a machine-conditional bbappend, so this needs a decision rather than a mechanical fix.

Until those are resolved, mx9x boards need `BB_DANGLINGAPPENDS_WARNONLY = "1"`.


hiagofranco:
Hi @jetm, found this while trying to bitbake the imx95 on my local build. Thanks.

mobileoverlord (approved):

PR #313, opened 2026-08-25
Add a qemux86-64 kas wrapper that inherits SBOM and cve-check

jetm:
## Problem

Composing features on the command line (`kas build machine.yml:feature/sbom.yml`)
means bakar derives its build directory from the first YAML's stem, so the
invocation lands in `build-qemux86-64` alongside runs that inherit neither
feature. Sharing one build directory means the SBOM/CVE end-to-end either reuses
artifacts produced without cve-check, or forces everything else to rebuild.

## Solution

A wrapper YAML composing `machine/qemux86-64.yml`, `feature/sbom.yml` and
`feature/cve-check.yml`, giving that combination its own build directory.

## Key changes

- `kas/qemux86-64-sbom-cve.yml`

## Reviewer notes

Both features must be inherited from the start rather than added to an existing
build: cve-check writes its `*_cve.json` during the package tasks, so a build
that gains the feature afterwards leaves `avocado-cve-report` with nothing to
read and it produces an empty report instead of failing.

`SPDX_INCLUDE_SOURCES` is deliberately left unset, and the file says why. It does
make sbom-cve-check's compiled-source kernel filter fire, so the obvious reading
is that it should be on. Measured 2026-08-20, the filter removed 76 of 2,094
kernel findings (3.6%) while costing 19x SPDX storage (0.276 -> 5.36 GB), 333x on
the per-image document (7.5 MB -> 2.50 GB), 3.5x build time, and ~50 GB of RAM to
parse that document.

### DIFF
--- a/kas/qemux86-64-sbom-cve.yml
+++ b/kas/qemux86-64-sbom-cve.yml
@@ -1,0 +1,25 @@
+# qemux86-64 with SPDX 3.0 SBOM and cve-check both inherited from the start.
+#
+# The repo convention is colon composition (kas build machine.yml:ci/mirrors.yml),
+# but bakar derives its build directory from the first YAML's stem, so composing
+# on the command line would land in build-qemux86-64 alongside unrelated runs.
+# This wrapper exists to give the SBOM/CVE end-to-end its own build directory.
+#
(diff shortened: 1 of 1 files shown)


mobileoverlord (approved):

PR #314, opened 2026-08-26
imx93: AHAB signed boot, end to end on hardware

jetm:
Supersedes #279, rebased onto `wrynose`. Same work, 108 commits down to 18: 86 of
the originals were already upstream by patch-id, and `git cherry` confirms it.
The old branch is untouched at `imx93-ahab-secure-boot` if anything needs
recovering from it.

Takes AHAB from "designed" to a board that boots a fully signed chain: NXP-signed
ELE container, OEM-signed `imx-boot`, OEM-signed OS container carrying the
initramfs, Linux, LUKS2 `/var` underneath it.

```
Authenticate OS container at 0xa8000000
Starting kernel ...
[  OK  ] Found device /dev/mapper/var.

# findmnt -no SOURCE,FSTYPE /var
/dev/mapper/var btrfs
```

There is no `Loading Ramdisk` line, and that absence is the result: the initrd
that derives the `/var` key is inside the signature rather than beside it.

## What was actually wrong

Three variables look like the AHAB switch and none is. `UBOOT_SIGN_ENABLE` is the
FIT/HABv4 path, referenced only in `compile_mx8m()`; `SECO_FIRMWARE_NAME` is the
NXP-signed ELE blob needed to boot at all; `SECOEXT_FIRMWARE_NAME` is defined
upstream only for mx8ulp/mx95/mx943 and its sole consumer is
`COMPATIBLE_MACHINE = "avocado-imx95-frdm"`, so the `"none"` shim is inert here.

Signing is SPSDK, not CST. `cst_signer`'s source refuses AHAB without SPSDK and
its README puts i.MX 8x/8ULP/9 on that path, so `nxpimage ahab sign` takes the
finished container and finds the header and signature block itself, with no
offsets scraped from a build log and no `var-cst-signer` to port.

**`CONFIG_AHAB_BOOT` changes what `booti` means.** Its first argument becomes a
container: `cmd/booti.c` authenticates it, then calls `container_get_image_dst()`
on the same address to find the kernel. Authentication has already copied each
image to the destination recorded inside the container, so staging the container
at the kernel's own destination means authentication destroys what the next line
parses. That hides behind `auth_cntr`, which calls the same function and passes
because nothing reads the container afterwards.

**A second `--data` payload does not carry the initramfs.**
`container_get_image_dst()` returns the first exec image and the first data image
and stops, so a second data payload is authenticated, placed, and then never
located; `booti` still reads `argv[1]`. Keeping the separate load only for its
size is worse: `${filesize}` comes from the unauthenticated copy on disk, so a
larger file leaves unverified bytes past the authenticated region, and appending
a cpio archive is the documented way to overlay files in an initrd. Bundling into
the kernel removes the argument instead of guarding it.

## Contents

- `avocado-ahab-sign.bbclass` - shared signing and one fail-closed check. Two
  copies of that logic could drift, and a drift there is a security bug.
- `avocado-os-container.bb` - packs the initramfs-bundled Image + dtb at this
  machine's addresses and signs it. Asserts at build time that the measured
  kernel reaches neither the device tree nor the container.
- `ahab.cfg` + `u-boot-imx` bbappend - `CONFIG_AHAB_BOOT=y`, verified against the
  produced `.config`, gated on the `ahab` DISTRO_FEATURE rather than on the
  machine alone. The bbappend also rewrites the boot flow, `fdt_addr` and
  `bootm_size`, for the signed path only.
- `INITRAMFS_IMAGE_BUNDLE` - the initramfs rides inside the signed exec payload,
  so `booti` takes no ramdisk argument.
- Boot partitions 128 -> 256 MiB under AHAB, `recovery` included since it is
  `bootpart 5` and holds a boot filesystem too. The signed container is ~200 MB
  once the initramfs is in it, because oe-core's `copy_initramfs` unpacks the
  `.cpio.zst` and the kernel embeds what it is given.
- `kas/feature/ahab.yml` - opt-in, no working default. Asking for `ahab` without
  keys halts the build rather than shipping an unsigned image that looks signed.
- `scripts/avocado-flash` - the bash tool replaced by the Python port, plus `ums`
  (writes the board's own card in place) and `fuse-srk` verbs.

## What is opt-in, and what is not

Behind the `ahab` feature: the signing class, the OS container recipe, the
imx-boot signing include, the stone boot-file list and partition sizes, the
U-Boot environment rewrite, `ahab.cfg`, and `INITRAMFS_IMAGE_BUNDLE`.

Unconditional, deliberately: `AVOCADO_AHAB_FAMILY` / `AVOCADO_KERNEL_DTB` (read
only by gated code) and `scripts/avocado-flash`.

An unsigned build of these boards behaves as it did before this branch. The
signed path's `fdt_addr` and `bootm_size` are applied by the AHAB rewrite rather
than carried in the shared environment, so they do not reach it.

## The memory map had to move

The addresses were chosen for a ~33 MB kernel. At ~200 MB the kernel unpacks
across the device tree, and U-Boot stops with `FDT image overlaps OS image`.
Moving the container then hit a second wall: DRAM on this board is not
contiguous - `bdinfo` reports `0x80000000+0x16000000` and `0x98000000+0x68000000`
with 32 MiB missing between them - and a load spanning the hole is refused with
`Reading file would overwrite reserved memory`, which names neither the hole nor
the size. The container stages at `0xa8000000`, in the second bank and above the
`no-map` carve-out Linux later claims as `ele-reserved`. NXP's 256 MiB bootm
window moves with it.

The build-time check compares ranges rather than addresses. The previous check
asked whether `cntr_addr` equalled `image_addr`, which is true of any layout that
fails this way and false of every layout that did fail here.

## Fusing

`fuse-srk` is a sibling of `sd`/`emmc` and never reachable from them: flashing is
repeatable, fuses are not. `ahab_close` is absent rather than gated. Dry run is
the default and it refuses on a non-blank fuse field, a non-open lifecycle, an
unsigned bootloader, or a read-back mismatch.

**The lab board has been fused, and wrongly.** `SRK_HASH` was burned
byte-swapped, so it is spent and `ahab_status` reports `BAD_KEY_HASH` on that
unit permanently. The board still boots because it is `OEM Open`, and
**`ahab_close` must never be run on it.** The tool now reads the fuse file
little-endian and asserts the ELE read-back after burning. Neither has been
proven against a passing burn, which needs a second board.

## What the rebase dropped, and why

Three of the 21 unique commits were left out. Each was checked against `wrynose`
rather than assumed:

- The `#267` front-port. Not AHAB work, and replaying it would have inserted a
  second, identical `PACKAGECONFIG:remove:class-nativesdk = "dbus"` stanza into
  `ganesha_6.5.bb` directly below the one already there. Its still-unlanded
  parts (the qemu-user-static static-library bbappends, `USE_FSAL_GLUSTER=OFF`,
  and `persist_ssh_host_keys`, which belongs to #275) are not in this PR.
- `stop building EFI capsule authentication nothing uses`. `wrynose` already
  carries both `SRC_URI` lines and the fragment, with a comment that is more
  accurate now that the board boots a FIT.
- `bring up the OP-TEE fTPM behind an opt-in feature`. `wrynose`'s version is a
  corrected descendant of this one and says so in its own comments, so
  reapplying it would undo the fix.

## Known gaps

- **The signed boot flow has not been re-derived against the FIT migration, and
  this is the blocker.** `wrynose` moved imx93 to a single signed FIT booted with
  `bootm`; this branch's env work predates that and assumes the three-file
  `booti` layout. The AHAB rewrite in `u-boot-imx_%.bbappend` is `sed -i`, which
  does not error on a pattern that matches nothing, so three of its five
  expressions are now silent no-ops: the `^fdt_addr=` rewrite (no such line in
  the FIT environment), the `^bootm_size=` rewrite (likewise, leaving the
  vendor's `0x10000000` where this branch's own comment says the bundled kernel
  needs `0x40000000`), and the `^bootcmd=` rewrite, which is harmless only by
  luck because the FIT `bootcmd` already equals the target form. Nothing here
  reaches an unsigned build, but the `ahab` feature on imx93 needs this settled
  before it can be trusted again.
- `ahab_status` reports `BAD_KEY_HASH` on the lab board because the burned hash
  is wrong, not because none is burned. A board with a correct burn is needed to
  see it clear.
- Signing is not reproducible: ECDSA draws a fresh nonce, so a signed bootloader
  has a new checksum every build. Matters for sstate reuse and OTA deltas.
- The initramfs is embedded uncompressed, because the kernel's compression choice
  depends on `INITRAMFS_SOURCE` and that variable only reaches the kernel as a
  make argument during `do_bundle_initramfs`. Compressing it would need an
  oe-core class override; the boot partition was widened instead.
- The machine configuration's comment block still reads "deliberately does NOT
  override `INITRAMFS_IMAGE_BUNDLE`" while the AHAB-gated override sits further
  down the same file. Behaviour is correct, the prose is not.

## Verification status

Verified: no conflict markers, no duplicated stanzas, no repeated `SRC_URI`
lines, and `parse-container-sdk` passes on both arm64 and x86_64, so the layer
set parses as rebased.

Not verified: nothing has been built or booted since the rebase, and a parse
check would not catch any of the three silent `sed` no-ops above, because a
pattern that matches nothing is not a parse error. The boot evidence quoted at
the top is from the pre-rebase branch.


mobileoverlord:

PR #319, opened 2026-08-26
imx93-frdm: close the offline U-Boot environment bypass

jetm:
Closes the offline environment bypass on `avocado-imx93-frdm`: with FIT
signature verification on, a saved U-Boot environment could still redirect the
boot, so the signature guarded the payload while the thing choosing the payload
stayed writable by anyone who could reach the card.

## The bypass, and why the obvious fix was not enough

`CONFIG_ENV_WRITEABLE_LIST` plus a permit list is the mechanism. But turning it
on alone would have bricked the board, and that is worth stating because it was
found by reading rather than by flashing.

This machine's entire boot flow reached U-Boot **only** as the saved
environment - `u-boot-env.inc` runs mkenvimage over `env/${MACHINE}.txt` and fwup
writes the result. `CONFIG_ENV_WRITEABLE_LIST` rejects every `H_EXTERNAL`
variable missing from the list (`env/flags.c:561-563`), and the saved copy is
imported `H_EXTERNAL` (`env/mmc.c:547`). All 13 flow variables would have been
dropped while the compiled-in default supplied 2, leaving the board on NXP's
stock `bootcmd`.

So the flow is now compiled in via `CONFIG_ENV_SOURCE_FILE`. Compiled-in
variables arrive `H_DEFAULT`, which `env_flags_validate` returns early for, so
the boot flow is immune **by construction** rather than by being listed - listing
`bootcmd` would have reopened the hole it closes.

`env/${MACHINE}.txt` deliberately keeps its own full copy. `env_load()` seeds the
built-in default before the storage driver only under
`CONFIG_ENV_WRITEABLE_LIST`; with the feature off the table is built from the
saved environment alone, so a reduced `.txt` would leave those builds with no
`bootcmd`. There is a debt marker on the duplication.

## The permit list

```
avocado_boot_slot, avocado_device_id, avocado_device_cert, avocado_device_key,
peridio_vsn_current, a.avocado_platform, a.avocado_architecture,
b.avocado_platform, b.avocado_architecture   (all :sw)
devnum, mmcblk                               (:dw)
```

`devnum` and `mmcblk` are `:dw`, not `:sw`, and that distinction is the security
content. `_env_flags_validate_type`'s string case is a bare `break`, so `:sw`
accepts anything: `mmcblk="1p6 init=/bin/sh x"` flows into `rootdev`, then into
`bootargs`, which the FIT signature does not cover - a root shell on a verified
kernel. `devnum` the same way, into `load` argument injection. Decimal rejects
trailing non-digits and still admits 0 and 1.

## u-boot-imx stays on 2026.04 here

093f504a pinned `PREFERRED_VERSION_u-boot-imx = "2025.04"` in `avocado-imx.inc`
for every i.MX machine, because 2026.04's stock defconfigs enable EFI capsule
authentication and USB DFU and `do_compile` fails on boards carrying neither
fix. Right default; wrong answer for this board, which turns both off at source
already. Overridden back per-machine, after the `require` so it actually wins.

2026.04 is a requirement rather than a preference here:
`CONFIG_ENV_SOURCE_FILE` resolves against `board/$(SYS_VENDOR)/$(SYS_BOARD)` -
`board/nxp/imx93_frdm` on 2026.04, `board/freescale/imx93_frdm` on 2025.04. The
bbappend refuses to build on the older layout rather than emit a bootloader with
no Avocado boot flow in it. The downgrade is independently destructive:
`CONFIG_SYS_BOOTM_LEN` halves from `0x8000000` to `0x4000000`, and one flash
under the old pin produced a board that ran SPL and never reached U-Boot.

## Autoboot is interruptible by any key

The series originally added `CONFIG_AUTOBOOT_KEYED` with a stop string. It is
removed - as an **ergonomics decision, not a defect report**. An earlier
revision of this PR argued the stop string was broken; that was wrong and is
corrected here.

A day went into "the stop string does not work" - sent through a confirmed
2.001-second window at 137 chars/sec and again at 44 chars/sec, plus hand-typed
pastes, all failing. The cause was the sender, not the board. The host script
paced itself on a read timeout, and `tio`'s read returns the instant bytes are
waiting rather than sleeping out its timeout, so during a boot flood it spun and
pushed thousands of bytes per second. That overran the LPUART FIFO, and
`passwd_abort_key()` - which drains one character per `udelay(10000)` - never
saw a clean sequence. Throttled to ~8 chars/sec, an unkeyed keypress broke in on
the first try with nothing about the board or the Kconfig changed.

So the trade is convenience against unattended safety. Convenience wins for now
because every documented recovery for this hardware goes through the U-Boot
prompt (`ums 0 mmc 1` is the flashing path), and reaching it should not depend
on getting a host-side send rate right.

The cost is accepted knowingly and carries a debt marker: unkeyed, this board's
serial line can halt an unattended boot on stray bytes, which is fine for a
bench unit with a human watching and is not fine for a HIL runner. Re-enabling
the keying then needs no new investigation - the string works, and what defeated
it is now understood.

## Verified on hardware

Flashed and booted on FRDM-IMX93. Confirmed on the board:

- Saved `bootcmd` is refused: `env set bootcmd ...` + `saveenv` accepted and
  written, and after reset the compiled-in command still runs.
- Saved `avocado_boot_slot` is still honoured, so A/B switching survives.
- FIT signature enforcement intact: `Verifying Hash Integrity ... sha256,rsa2048:FIT+ OK`.
- `booti` is present but unreachable from the saved environment.
- Autoboot now prints `Hit any key to stop autoboot: 2` and any key reaches the
  prompt.

`do_configure` selects 2026.04 under the layer-wide 2025.04 pin, confirmed by
which work directory it touched.

## A/B upgrade verified

The `-t upgrade` cycle has now been run on hardware, over UMS from the U-Boot
prompt:

- `fwup: Upgrading boot slot B` - correctly selected `upgrade.b` from the
  running slot A.
- Slot stamps landed, read back from **both** halves of the redundant
  environment (`CONFIG_ENV_OFFSET=0x400000`, `CONFIG_ENV_OFFSET_REDUND=0x440000`),
  with the live copy identified by the higher flags byte (primary 15 >
  redundant 14).

| | primary (live) | redundant (stale) |
|---|---|---|
| `avocado_boot_slot` | `b` | `b` |
| `b.avocado_platform` | present | present |
| `b.avocado_architecture` | present | **absent** |

Both `b.` stamps were undefined before the upgrade. The asymmetry is why both
halves have to be read: the stale copy alone reads as a partial write.

- Rebooted into `Booting B`, `Verifying Hash Integrity ... sha256,rsa2048:FIT+ OK`
  three times, and reached userspace - so the slot switch, the signature
  enforcement and the upgraded slot all hold together.

### DIFF
--- a/meta-avocado-nxp/conf/machine/avocado-imx93-frdm.conf
+++ b/meta-avocado-nxp/conf/machine/avocado-imx93-frdm.conf
@@ -13,6 +13,37 @@ require conf/machine/include/avocado-imx-frdm.inc
 require conf/machine/imx93-11x11-lpddr4x-frdm.conf
 require conf/machine/include/fsl-imx-preferred-env.inc
 
+# Override the layer-wide u-boot-imx pin back to 2026.04 for this machine only.
+#
+# avocado-imx.inc:10 sets PREFERRED_VERSION_u-boot-imx = "2025.04" for every
+# i.MX machine, because 2026.04's stock defconfigs enable
+# CONFIG_EFI_CAPSULE_AUTHENTICATE (whose ESL step needs cert-to-efi-sig-list,
+# which no recipe in this layer set provides) and USB DFU, and do_compile fails
+# on boards that carry neither fix. That is the right default for a board nobody
+# has fixed; it is the wrong answer here, where both are already turned off at
+# source by no-efi-capsule-auth.cfg and disable-unused-vendor-features.cfg.
+#
+# This is a REQUIREMENT, not a preference. The compiled-in environment selects
+# its source with CONFIG_ENV_SOURCE_FILE, which resolves against
+# board/$(SYS_VENDOR)/$(SYS_BOARD) - board/nxp/imx93_frdm on 2026.04, but
+# board/freescale/imx93_frdm on 2025.04. u-boot-imx_%.bbappend refuses to build
+# rather than silently produce a bootloader with no Avocado boot flow in it, so
+# on 2025.04 this machine does not build at all.
+#
+# A downgrade is also destructive in its own right: 2025.04 carries
+# CONFIG_SYS_BOOTM_LEN=0x4000000 against 2026.04's 0x8000000, halving the bootm
+# budget for the FIT. One bootloader flash under the old pin produced a board
+# that ran SPL and never reached U-Boot, recoverable only over uuu serial
+# download (2026-08-25).
+#
+# Placed after the requires above on purpose - avocado-imx.inc is pulled in at
+# line 10, so an assignment before it would be overwritten by the layer-wide
+# value rather than overriding it.
+#
+# EFI capsule update is wanted on other hardware later. Re-enabling it is a
+# per-machine decision that needs efitools-native available first
+# (meta-imx/meta-imx-sdk carries it, and is not currently in bblayers).
+PREFERRED_VERSION_u-boot-imx = "2026.04"
 
 # This machine deliberately does NOT override INITRAMFS_IMAGE_BUNDLE, so the
 # distro default in conf/distro/avocado.conf ("0") applies and the initramfs
--- a/meta-avocado-nxp/conf/machine/avocado-imx93-frdm.conf
+++ b/meta-avocado-nxp/conf/machine/avocado-imx93-frdm.conf
@@ -147,10 +147,49 @@ MACHINE_FEATURES:append = " optee"
 # worth catching. Two claims, one fact: ftpm names the implementation, tpm2 the
 # interface it presents, exactly as avocado-qemuarm64.conf explains for itself.
 #
+# verified-boot is unconditional, like encrypted-var, and deliberately NOT
+# derived from DISTRO_FEATURES. The token states what this MACHINE can deliver,
+# not what a given build asked for: the FIT signing wiring lives in this file
+# (UBOOT_SIGN_*, FIT_KERNEL_SIGN_*, FIT_CONF_DEFAULT_DTB) and in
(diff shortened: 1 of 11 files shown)


mobileoverlord:
Stacked this locally with #316 (FIT boot on imx8mp-evk) to test the i.MX branches together; both parse and imx8mp-evk boots the combined tree. One interaction to fix here before merge:

`CONFIG_ENV_WRITEABLE_LIST=y` sits in `fit-verify.cfg`, which every `u-boot-imx` board gets under `verified-boot`, while the permit list (`env-writeable-list.patch` → `imx93_frdm.h`) and the compiled-in boot flow (`env-compiled-in.cfg`) exist for imx93-frdm only. On imx8mp-evk with `verified-boot` the symbol alone rejects every saved variable — `avocado_boot_slot` and the Avocado `bootcmd` included — and the board falls back to the stock `bootcmd`. Your own comment in `fit-verify.cfg` describes exactly this failure for the no-list case.

What I applied locally: move the symbol into `env-writeable-list.cfg` and add it with the patch under the same gate *and* the same machine scope:

```python
        d.appendVar('SRC_URI', ' file://fit-verify.cfg')
        if d.getVar('MACHINE') == 'avocado-imx93-frdm':
            d.appendVar('SRC_URI', ' file://env-writeable-list.cfg file://env-writeable-list.patch')
```

A board then adopts the lockdown together with its compiled-in env and permit list. Also `02bc0310e2` (keep u-boot-imx on 2026.04 against the layer pin) is redundant now that #320 removed the pin.

mobileoverlord:

PR #321, opened 2026-08-26
cryptsetup-var: encrypt a grown /var in place, report conversion size/progress

mobileoverlord:
Follow-up to #310/#317 for the OTA scenario: a device deployed with a plaintext `/var` receives an update that turns `var.encrypt` on.

**Before:** `encrypt_in_place` needs 32 MiB of free tail for the LUKS2 header, but on any deployed device grow-var has already resized the btrfs to fill the partition, so the guard fired (`leaves no 32 MiB for a LUKS header - cannot encrypt in place`), `cryptsetup-var.service` failed, `/var` never mounted, emergency shell. Only fresh flashes (461 M seed inside a 55 G partition, as tested on Thor/Orin) ever converted.

**Now:**
- When headroom < 32 MiB and the fs is btrfs: mount on `/run/cryptsetup-var-shrink`, `btrfs filesystem resize -<deficit>M`, unmount, then the unchanged `--reduce-device-size 32M` reencryption. A shrink refusal (tail chunk in use / full) still fails closed with a message saying so.
- Console line with the MiB being converted, that it is a one-time resumable migration, and `--progress-frequency 30` when cryptsetup ≥ 2.4 advertises it — a seeded `/var` can be tens of GiB and this step runs for minutes with nothing else on screen.

Generic (meta-avocado core); every target using `cryptsetup-var` gets it. The initramfs already has btrfs-progs and the btrfs module (it mounts `/var`), and `mount` (util-linux/busybox).

**Not in this PR:** bounding the work to *used* data (shrink to used+slack before encrypting, re-grow after). `--device-size` still covers btrfs `total_bytes`, so a 40 G fs with 5 G used encrypts 40 G. That needs a chunk-relocation pass and is a separate trade-off.

Tests: `tests/test-cryptsetup-var-inplace.sh` 11/11 (new: grown-fs shrink by exact deficit, mount→resize→umount ordering before reencrypt, resulting device-size, console size line, and that a small flashed fs is never shrunk).

### DIFF
--- a/meta-avocado-nxp/stone/stone-imx95-frdm.json
+++ b/meta-avocado-nxp/stone/stone-imx95-frdm.json
@@ -132,8 +132,6 @@
         {
           "name": "var",
           "image": "var",
-          "size": 512,
-          "size_unit": "mebibytes",
           "expand": "true"
         }
       ]
--- a/meta-avocado/recipes-core/cryptsetup-var/files/cryptsetup-var.sh
+++ b/meta-avocado/recipes-core/cryptsetup-var/files/cryptsetup-var.sh
@@ -217,13 +217,45 @@ maybe_resize() {
 # Interrupted mid-way (power cut), LUKS2 records the reencryption in its
 # metadata; open_var resumes it before opening. A filesystem whose size cannot
 # be read gets the whole-partition reencryption instead - slower, still correct.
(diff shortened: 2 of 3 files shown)


mobileoverlord:

PR #322, opened 2026-08-26
Fix two buildtools host-build failures

jetm:
## Problem

Two recipes fail when the host toolchain comes from buildtools rather than the distro. Both are in native/host tooling, so they block the build before any target work starts.

- `abseil-cpp` publishes a C++20 ABI its consumers do not use, and `protobuf-native`/`grpc` die on `'weak_ordering' has not been declared in 'std'`.
- `greengrass-bin`'s `do_merge_config` fails with `ModuleNotFoundError` for `yaml`.

## Solution

Each is corrected with a bbappend that fixes an assumption the upstream recipe makes about its environment, rather than by patching a recipe this layer does not own.

abseil's CMake selects the highest C++ standard the compiler advertises when `CMAKE_CXX_STANDARD` is unset. Under buildtools that is C++20, so abseil's install step rewrites `absl/base/options.h` with `ABSL_OPTION_USE_STD_ORDERING=1` while consumers still compile at the gcc default of C++17. Pinning abseil to 17 fixes the standard in one place and every consumer inherits a matching ABI, rather than each consumer having to be raised individually and the outcome still depending on what the host toolchain advertises.

`greengrass-bin`'s `do_merge_config` shells out to a bare `python3 -c` that imports yaml. On a host build the buildtools SDK python sits ahead of the staged native tools on PATH and ships no PyYAML. The recipe already `DEPENDS` on `python3-pyyaml-native`, so only the interpreter is wrong; `inherit python3native` prepends `python3-native` to PATH and the existing call resolves to it.

## Key changes

- Add `meta-avocado-distro/recipes-devtools/abseil-cpp/abseil-cpp_%.bbappend` pinning `-DCMAKE_CXX_STANDARD=17` for all classes.
- Add `meta-avocado-distro/recipes-iot/aws-iot-greengrass/greengrass-bin_%.bbappend` inheriting `python3native`.

## Reviewer notes

- The abseil pin applies to all classes, not only native, so the target abseil stays ABI-consistent with the native one. That makes this target-visible as well as host-visible.
- `greengrass-bin` is reached through the `cloud-aws` feature group, so it is covered by the `full` build variant and not by `minimal`.
- Neither change has been built yet. This is opened as a draft with `build-qemux86-64` and `build-qemuarm64` labelled so CI covers them; it will be marked ready once that run is green.
- `abseil-cpp` is a transitive dependency and is not named by any packagegroup, so it may not be exercised on every machine.
- `meta-avocado-distro/recipes-iot/` is a new directory in this layer.


meta-avocado-distro/dynamic-layers/meta-aws/recipes-iot/aws-iot-greengrass/greengrass-bin_%.bbappend
```diff
@@ -0,0 +1,7 @@
+# Host build: do_merge_config runs a bare `python3 -c "... import yaml ..."`.
+# On the host build PATH bare python3 resolves to the buildtools SDK python,
+# which has no PyYAML, so the merge fails with ModuleNotFoundError. The recipe
+# already DEPENDS on python3-pyyaml-native; inheriting python3native prepends
+# the staged python3-native to PATH so bare python3 picks the native python
+# that can import the staged yaml module.
+inherit python3native
```
mobileoverlord, line 7: Dangling bbappend on every config that does not pull `kas/feature/cloud-aws.yml` (all the minimal/qemu machines): `greengrass-bin` only exists in meta-aws, so bitbake parse fails with `No recipes in default available for .../greengrass-bin_%.bbappend`. parse-container-sdk did not catch it because that job only parses `meta-avocado` + `meta-avocado-shared`, not the distro layer.

Move this to `meta-avocado-distro/dynamic-layers/meta-aws/recipes-iot/aws-iot-greengrass/` — layer.conf already has the `BBFILES_DYNAMIC` entry for it, and `aws-iot-device-client_%.bbappend` lives there for the same reason.

mobileoverlord (changes requested):
abseil pin looks right. One blocker:

meta-avocado-distro/dynamic-layers/meta-aws/recipes-iot/aws-iot-greengrass/greengrass-bin_%.bbappend
jetm: Confirmed and fixed in `509ce1e`. Read `layer.conf` directly to check the mechanism: `BBFILES` (unconditional) globs `recipes-*/*/*.bbappend` on every parse regardless of active layers, while `BBFILES_DYNAMIC`'s `meta-aws` entry only fires when that layer is actually in the build. `greengrass-bin_%.bbappend` was sitting under the first, so any minimal/qemu config without `cloud-aws.yml` hit exactly the dangling-target parse failure you described.

Moved it to `meta-avocado-distro/dynamic-layers/meta-aws/recipes-iot/aws-iot-greengrass/`, matching `aws-iot-device-client_%.bbappend`'s existing placement in the same directory for the same reason. Confirmed no other file references the old path.

I didn't run a live `bitbake -p`/`bitbake -n` parse to prove it end-to-end - this scratch worktree isn't positioned as a `meta-avocado/` subdir inside the full multi-repo kas workspace, and reproducing that layout in the primary checkout risked disturbing other in-progress work sitting there. The fix rests on reading `BBFILES`/`BBFILES_DYNAMIC` directly plus matching the established sibling pattern, not a build run - flagging that in case you want to confirm with an actual parse before merging.

mobileoverlord (approved):
Move verified against the `BBFILES_DYNAMIC` glob in layer.conf. Thanks.

PR #327, opened 2026-08-27
Fix rootdev in the imx93-frdm compiled-in boot env

jetm:
## Problem

The per-slot dm-verity work renumbered `env/avocado-imx93-frdm.txt` for the new hash partitions but left the compiled-in environment untouched. The two are not interchangeable: `CONFIG_ENV_WRITEABLE_LIST` rejects every saved boot-path variable on import, so with the `verified-boot` feature on, the compiled-in copy is the one that runs.

Against the current layout (`rootfs-a=p5`, `rootfs-b=p6`, `rootfs-a-hash=p7`, `rootfs-b-hash=p8`) it is wrong three ways:

| | compiled-in today | correct |
|---|---|---|
| slot a `rootdev` | p6 - rootfs-**b** | p5 |
| slot b `rootdev` | p7 - rootfs-a-**hash** | p6 |
| recovery | `bootpart 5` - now rootfs-a | no recovery branch |

and `bootcmd` omits `avocado_verity_init` entirely, so `verified-boot` is the only mode that authenticates the FIT and the only mode that never reads a root hash back out of it.

Only `avocado-imx93-frdm` is affected - it is the one machine wired for `env-compiled-in.cfg`, and the permit list is scoped to it too.

## Solution

Mirror the saved copy: p5/p6 for the slots, p7/p8 for their hash trees, no recovery branch (an unset or invalid slot falls through to A, as the saved copy already does), and `avocado_verity_init` in `bootcmd`.

## Key changes

- Slot A -> `rootdev` p5 / `roothashdev` p7; slot B -> p6 / p8; top-level `rootdev` -> p5
- Recovery branch removed - its `bootpart 5` is rootfs-a now, so it could only stop the board at the prompt
- `avocado_verity_init` added and wired into `bootcmd`, matching the saved copy

## Reviewer notes

Verified on hardware, on an FRDM-IMX93 with `verified-boot` on and the PoC feature off. Built, flashed with `complete`, power-cycled, and booted:

```
/proc/cmdline : audit=0 console=ttyLP0,115200 root=/dev/mmcblk1p5 rootwait
findmnt /     : /dev/mmcblk1p5
p5            : rootfs-a
```

Before this change the same boot resolves `rootdev` to p6, which is rootfs-b. The on-device saved copy read back with `fw_printenv` now matches the compiled-in one branch for branch, including `roothashdev` p7/p8.

Slot B is not boot-tested: `complete` deliberately memsets rootfs-b and boot-b rather than writing them, so a freshly provisioned card cannot boot B without an A/B upgrade first. Its branch is verified in both environment texts on the device.

Cross-checked ahead of the board run by reproducing U-Boot's own environment compilation - `cpp` with the flags from the Makefile's `cmd_gen_envp`, then `scripts/env2string.awk` - and decoding `CONFIG_EXTRA_ENV_TEXT` before and after.

Nothing in the build compares the two environment files, which is the debt the `.txt` header already records. A guard for that belongs in its own change.

### DIFF
--- a/meta-avocado-nxp/recipes-bsp/u-boot/u-boot-imx/avocado-imx93-frdm.env
+++ b/meta-avocado-nxp/recipes-bsp/u-boot/u-boot-imx/avocado-imx93-frdm.env
@@ -48,7 +48,7 @@ devtype=mmc
 devnum=1
 mmcblk=1
 bootpart=3
-rootdev=/dev/mmcblk${mmcblk}p6
+rootdev=/dev/mmcblk${mmcblk}p5
 
 /*
  * Load the FIT well clear of where its own kernel decompresses to.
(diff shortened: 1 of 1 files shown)


mobileoverlord (approved):
Matches the saved copy in env/avocado-imx93-frdm.txt branch for branch (p5/p6 slots, p7/p8 hash trees, no recovery branch, avocado_verity_init with the trailing true). The compiled-in copy was our miss when the hash partitions landed - thanks for the hardware run.

PR #333, opened 2026-08-27
security: build declared capabilities into the feed, let the runtime choose

mobileoverlord:
Yocto builds what a machine can do; the user's `avocado.yaml` decides what a runtime uses. `encrypted-var` was the one place the feed still decided for the user (a `complete` build encrypted `/var` on every runtime, and a runtime asking for `var.encrypt` on a plain feed could not mount it). One mechanism now, on every platform.

## What

- `AVOCADO_SECURITY_CAPABILITIES` is the only build-side switch. The initramfs/rootfs packagegroups, the kernel `dm-crypt.cfg`, and the Jetson security packages gate on it; `DISTRO_FEATURES` no longer selects `encrypted-var` (`kas/feature/encrypted-var.yml` removed, `complete.yml` adds no security tokens, a leftover token warns).
- `cryptsetup-var.service` runs only when the runtime's initramfs carries `/etc/avocado/var-encrypt` — the marker `avocado-cli` writes for `var.encrypt`. The feed's own initramfs never has it. Same marker `avocado-tegra-init` already used.
- `/var` mounts from `/dev/disk/by-avocado/var`: `base-files` generates a udev rule pointing it at the plaintext partition (not when it carries a LUKS header), `cryptsetup-var-udev` points it at the opened mapping. One fstab serves both choices; `AVOCADO_VAR_PART_DEV` is always the plaintext partition and no machine conf toggles on a feature.
- `docs/security-capabilities.md` (model, per-capability, per-target table); `docs/adding-a-machine-target.md` gains the declaration, the var device rule and the two-runtime hardware test.

## Depends on

avocado-cli `main` ≥ 083fbbb (#216): `runtimes.<name>.var.encrypt` writes the `/etc/avocado/var-encrypt` marker into the runtime's initramfs. avocadoctl #24 for OS-bundle updates on GPT machines.

## How verified

`bitbake -p` / `-g avocado-complete` with `complete.yml` on imx8mp-evk, imx93-frdm, qemux86-64, raspberrypi0-2w: `encrypted-var` absent from `DISTRO_FEATURES`; `cryptsetup-var` in the initramfs and `cryptsetup-var-udev`/`-posture` in the rootfs from the declaration; `base-files` produces `/dev/disk/by-avocado/var /var ...` and the expected rule (`ID_PART_ENTRY_NAME=="var"` on GPT, `KERNEL=="mmcblk0p3"` on the Pi). jetson-orin-nano parse clean. Hardware pass on imx8mp-evk done (see comments): plaintext runtime, then `var.encrypt` turned on over `avocado deploy` with in-place conversion, then a clean reopen boot.

Behaviour unchanged for: Jetson (already opt-in), verified-boot/ahab/ftpm/tpm2 (still build-time requests checked against the declaration), any machine with no declaration.


mobileoverlord:
jetson-orin-nano parse (system-base + networking): `PARSE_OK`; `SECURITY_PACKAGES` resolves to the full set from the `encrypted-var`/`ftpm` declaration; `cryptsetup`/`cryptsetup-var` in the initramfs packagegroup; `AVOCADO_VAR_PART_DEV=none`, so no fstab line or udev rule is generated - Jetson keeps mounting /var from avocado-tegra-init.

jetm (approved):

meta-avocado/recipes-core/cryptsetup-var/files/cryptsetup-var.service
```diff
@@ -13,6 +13,13 @@ After=dev-disk-by\x2dpartlabel-var.device
 After=systemd-udevd.service
 Wants=systemd-udevd.service
 ConditionPathExists=/dev/disk/by-partlabel/var
+# The per-runtime opt-in. avocado-cli writes this marker into a runtime's
+# initramfs when avocado.yaml sets var.encrypt; the feed's own initramfs and a
+# runtime that did not opt in have no marker and mount the plaintext /var.
+# The tooling is in every declaring machine's feed regardless, so the switch
+# is the user's, not the build's. (avocado-tegra-init masks this unit and
+# checks the same marker itself.)
+ConditionPathExists=/etc/avocado/var-encrypt
```
jetm, line 31: **The `/etc/avocado/var-encrypt` writer this condition depends on doesn't exist in `avocado-cli` yet, on any branch.** I checked: `git grep -rn "var-encrypt\|var\.encrypt" src/` in the local `avocado-cli` clone returns nothing, and `git log --all --grep="var.encrypt\|var-encrypt"` across every branch (`main`, `container-dev-mode`, and every remote branch) also returns nothing.

Combined with this same PR deleting `kas/feature/encrypted-var.yml` - previously the only way to produce a build that actually encrypts `/var` - the net effect right now is: no build on any machine can encrypt `/var` until the CLI ships this marker writer. The unit condition-skips silently every boot (a failed `ConditionPathExists` is not an error, nothing logs it), so this isn't a build break, it's a feature going dark with no diagnostic.

If the CLI change is landing in lockstep (a branch I don't have visibility into, or queued right behind this), say so in the PR description so a reviewer isn't left to rediscover this. If it isn't queued yet, this PR probably shouldn't merge ahead of it.

meta-avocado/recipes-avocado/packagegroups/packagegroup-avocado-initramfs.bb
```diff
@@ -6,7 +6,7 @@ inherit packagegroup nospdx
 PACKAGES = "${PN}"
 
 RDEPENDS:${PN} = "\
-  ${@bb.utils.contains('DISTRO_FEATURES', 'encrypted-var', 'cryptsetup cryptsetup-var', '', d)} \
+  ${@bb.utils.contains('AVOCADO_SECURITY_CAPABILITIES', 'encrypted-var', 'cryptsetup cryptsetup-var', '', d)} \
```
jetm, line 9: **The initramfs never gets the udev rule that lets the LUKS mapping claim `/dev/disk/by-avocado/var`, so an already-encrypted device can't mount `/var` in the initrd.**

`99-zz-cryptsetup-var.rules` - the only rule that points `by-avocado/var` at the dm mapping once the container is open - ships in the `cryptsetup-var-udev` package. This RDEPENDS list pulls only `cryptsetup cryptsetup-var`, not `-udev`. `cryptsetup-var-udev` is pulled by the *rootfs* packagegroup (line 64 of `packagegroup-avocado-rootfs.bb`) and by `packagegroup-avocado-tegra-extra.bb`, never by this one.

The initrd carries its own `base-files`-installed `/etc/fstab`, and that fstab now names `/dev/disk/by-avocado/var` for `/var` (per this same PR's change to `base-files_%.bbappend`). On second-and-later boots of an encrypting device the partition is `crypto_LUKS`, so `61-avocado-var.rules` correctly declines to claim the symlink - and nothing else in the initrd creates it either, since the rule that would is not installed there. `dev-disk-by\x2davocado-var.device` never appears; the initrd's `var.mount` blocks for the full device timeout and fails, taking `local-fs.target` down with it. Before this PR the same fstab slot named `/dev/mapper/var` directly, created by the stock dm rules already present in the initrd - this worked.

Add `cryptsetup-var-udev` here, or explain why the initrd `var.mount` no longer needs it.

meta-avocado-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -2,15 +2,51 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
...
 do_install:append() {
     install -m 0644 ${UNPACKDIR}/fstab ${D}${sysconfdir}/fstab
     if [ "${AVOCADO_VAR_PART_DEV}" = "none" ]; then
-        # Var partition is mounted dynamically by init (e.g. avocado-tegra-init)
         sed -i '/@AVOCADO_VAR_PART_DEV@/d' ${D}${sysconfdir}/fstab
-    else
-        sed -i 's|@AVOCADO_VAR_PART_DEV@|${AVOCADO_VAR_PART_DEV}|g' ${D}${sysconfdir}/fstab
+        return
     fi
+    sed -i 's|@AVOCADO_VAR_PART_DEV@|/dev/disk/by-avocado/var|g' ${D}${sysconfdir}/fstab
```
jetm, line 40: **The ordering guarantee the six deleted `AVOCADO_VAR_PART_DEV` ternaries existed for isn't re-established, so first-boot encryption can race the mount instead of gating it.**

Four of those six deleted comments said essentially the same thing (e.g. the old `avocado-qemux86-64.conf`): mounting `/var` from `/dev/mapper/var` meant `var.mount` implicitly ordered after `cryptsetup-var.service`, because that device path could not exist until the service created it. That was a structural guarantee, not a policy - `var.mount` had nothing to race.

This line replaces it with `/dev/disk/by-avocado/var`, a name the *plaintext* partition can claim on its own via `61-avocado-var.rules` the moment udev sees it - no service required. `cryptsetup-var.service` itself gained no explicit `Before=var.mount` (or `Conflicts=var.mount`) to compensate; its `Requires=`/`After=` block, unchanged by this PR, only concerns the partlabel device, not the mount unit.

On a runtime's very first boot with `var.encrypt: true`, the partition is still plaintext, so `61-avocado-var.rules` claims the symlink immediately and `var.mount` can start concurrently with `cryptsetup-var.service`. If `var.mount` wins the race, `cryptsetup`'s in-place encrypt attempt hits a mounted filesystem and fails (`cryptsetup-var.sh` has no unmount handling) - and because nothing here fails loudly on that outcome, the device is left silently, permanently unencrypted: no LUKS header ever gets written, `write_posture` (step 6 of the script) never runs, and `avocado-posture-publish.sh` reports "`/var` encryption not in this image" rather than "not encrypted" - so fleet-wide posture reporting can't even see the gap. The old model made this outcome structurally unreachable; this one makes it a timing-dependent silent failure.

Needs an explicit `Before=var.mount` on `cryptsetup-var.service` (both here and in the initrd, once the previous comment's package gap is closed).

jetm (commented):
Reviewed. 7 blocking comments inline, all in the encrypted-`/var` mount/ordering mechanics this PR reworks - the declared-capability/gating migration itself is clean (traced every `DISTRO_FEATURES`+`encrypted-var` and `AVOCADO_VAR_PART_DEV == '/dev/mapper/var'` site; all migrated, no stragglers, `kas/feature/encrypted-var.yml`'s deletion has no remaining referrers). A larger set of advisory notes (stale comments now describing mechanisms this PR removes, doc-coverage gaps, a couple of narrower edge cases) is withheld per the posting bar; ask if useful.

meta-avocado-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -2,15 +2,51 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
...
+# that opted into encryption (avocado.yaml var.encrypt) and one that did not,
+# and the build never has to know which (docs/security-capabilities.md).
+# udev match for the plaintext partition, from the three forms above.
+def avocado_var_udev_match(d):
+    dev = d.getVar('AVOCADO_VAR_PART_DEV') or ''
+    if dev.startswith('/dev/disk/by-partlabel/'):
+        return 'ENV{ID_PART_ENTRY_NAME}=="%s"' % dev.rsplit('/', 1)[-1]
+    if dev.startswith('PARTUUID='):
+        return 'ENV{ID_PART_ENTRY_UUID}=="%s"' % dev.split('=', 1)[1]
+    if dev.startswith('/dev/'):
```
jetm: **This `/dev/` catch-all accepts paths it can't actually match, and produces a udev rule that silently never fires - no `bb.fatal`, no build error, `/var` just never mounts.**

Only `by-partlabel` gets a real match clause (`ENV{ID_PART_ENTRY_NAME}`). Every other `/dev/disk/by-*` form - `by-uuid`, `by-label`, `by-id`, `by-partuuid` - falls into this branch instead of the earlier ones, and `KERNEL=="%s"` (line 24) matches a bare kernel device name like `sda3` or `mmcblk0p3`, never a path containing `/`. So `AVOCADO_VAR_PART_DEV = "/dev/disk/by-uuid/<uuid>"` produces `KERNEL=="disk/by-uuid/<uuid>"`, which is syntactically valid udev and will never match anything. Same fate for anyone who carries over the old value verbatim: `/dev/mapper/var` produces `KERNEL=="mapper/var"`.

The function's own strictness principle (reject unhandled forms via `bb.fatal` at line 27) doesn't actually cover this case, because the value *does* start with `/dev/`, so it never reaches the fatal - it just silently generates a dead rule. The doc table this PR adds (`docs/adding-a-machine-target.md`) documents `PARTUUID=` and "a raw `/dev/<node>`" as accepted forms without saying a `/dev/disk/by-*` path (other than by-partlabel) is rejected, so this is an easy value for someone to reach for.

Either require `'/' not in dev[len('/dev/'):]` before taking this branch (falling through to the existing `bb.fatal` otherwise), or add explicit handling for the other `by-*` forms (`ID_FS_UUID`, `ID_PART_ENTRY_UUID`, `ID_FS_LABEL`).

meta-avocado-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -2,15 +2,51 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
...
+
+    install -d ${D}${nonarch_base_libdir}/udev/rules.d
+    cat > ${D}${nonarch_base_libdir}/udev/rules.d/61-avocado-var.rules <<EOF
+# Generated by base-files from AVOCADO_VAR_PART_DEV=${AVOCADO_VAR_PART_DEV}.
+# The plaintext /var partition claims /dev/disk/by-avocado/var; once it carries a
+# LUKS header it stops claiming it, and only the opened mapping does
+# (99-zz-cryptsetup-var.rules). Deciding on ID_FS_TYPE rather than on link
+# priority alone keeps the name from ever pointing at the raw LUKS partition
+# during coldplug, where var.mount would otherwise race the mapping and fail.
+SUBSYSTEM=="block", ${AVOCADO_VAR_UDEV_MATCH}, ENV{ID_FS_TYPE}!="crypto_LUKS", SYMLINK+="disk/by-avocado/var"
```
jetm: **Two separate issues on this line.**

**1. `ENV{ID_FS_TYPE}!="crypto_LUKS"` is satisfied when the property is simply unset, not only when it's set to something else** - that's how udev's negative match works. The comment above (lines 44-46) says this check "keeps the name from ever pointing at the raw LUKS partition during coldplug", but that only holds once `60-persistent-storage.rules`'s `blkid` import has actually populated `ID_FS_TYPE` for that event. Any coldplug/event path where it hasn't yet (this rule carries no `ACTION=="add|change"` filter itself, unlike its `99-zz-cryptsetup-var.rules` sibling, which explicitly does) makes the *raw LUKS partition* claim the symlink - the exact case this line exists to prevent. Fix: assert the property is present before trusting the negative test, e.g. `ENV{ID_FS_TYPE}=="?*", ENV{ID_FS_TYPE}!="crypto_LUKS"`, and add `ACTION=="add|change"`.

**2. The `disk/by-avocado/var` name isn't scoped to the boot disk.** The match clause (from `avocado_var_udev_match`) tests partition name/UUID/kernel-name only - any GPT partition literally named `var` on *any* attached disk matches. On an encrypting device the internal partition is excluded (crypto_LUKS), which leaves the name free for competing media to claim - e.g. a USB stick with a `var`-labeled partition, attached and powered on. This doc this PR adds (`docs/security-capabilities.md:176-178`) already states the underlying problem for Jetson ("a fixed `by-avocado/var` rule cannot express 'the var partition on the boot disk'") and works around it by opting Jetson out entirely (`AVOCADO_VAR_PART_DEV = "none"`) - every other machine still ships the unscoped rule. Worth scoping on `ENV{ID_PART_ENTRY_DISK}` (or the root device's major:minor) so only the boot disk's `var` partition can claim the name.

meta-avocado-raspberrypi/conf/machine/include/avocado-raspberrypi.inc
```diff
@@ -16,15 +16,12 @@ CMDLINE:append = " audit=0"
 # is honoured.
 AVOCADO_SECURITY_CAPABILITIES = "encrypted-var"
```
jetm: **This machine declares `encrypted-var` with no dm-crypt kernel support reachable from anywhere in its build.**

The kernel half of "declaring builds it into the feed" reaches a machine through exactly two routes: `meta-avocado/recipes-kernel/linux/linux-yocto_%.bbappend` (matched to the `linux-yocto` recipe name only) or a vendor bbappend that `require`s `recipes-kernel/linux/avocado-security-kernel.inc`. `linux-raspberrypi_%.bbappend` does neither - it only adds `avocado-core.cfg`/`avocado-extra.cfg`/`reterminal.cfg`, none of which carry DM or LUKS-cipher kernel symbols. There's also no `recipes-core/cryptsetup-var/cryptsetup-var.bbappend` for this platform, so it falls back to the generic `var-key.sh` backend.

This PR's own new `docs/security-capabilities.md` names Raspberry Pi's *partlabel* gap ("built, service gap") but the kernel gap is separate and isn't mentioned. An integrator reading the table would reasonably conclude the tooling is ready to use once the partlabel issue is worked around - it isn't. Worth either wiring the kernel fragment for this platform, or naming this gap explicitly alongside the partlabel one so nobody discovers it by way of a failed `luksFormat` on-device.

meta-avocado/recipes-avocado/packagegroups/packagegroup-avocado-rootfs.bb
```diff
@@ -61,7 +61,7 @@ RDEPENDS:${PN} = " \
   avocadoctl \
   avocado-users \
   btrfs-tools \
-  ${@bb.utils.contains('AVOCADO_VAR_PART_DEV','/dev/mapper/var','cryptsetup-var-udev cryptsetup cryptsetup-var-posture','',d)} \
+  ${@bb.utils.contains('AVOCADO_SECURITY_CAPABILITIES', 'encrypted-var', 'cryptsetup-var-udev cryptsetup cryptsetup-var-posture', '', d)} \
```
jetm, line 64: **This gate migration ships `cryptsetup-var.service` to Raspberry Pi for the first time, and every RPi boot will now stall waiting on a device that can never exist there.**

Before this PR, `cryptsetup-var` reached the rootfs only when `AVOCADO_VAR_PART_DEV == '/dev/mapper/var'` - never true for RPi, since it's MBR-partitioned with no GPT partlabel. Now the gate is the `encrypted-var` capability declaration, which RPi *does* set (`avocado-raspberrypi.inc:17`), so the package - and the statically `.wants`-installed `cryptsetup-var.service` (`cryptsetup-var.bb` hand-creates the symlink unconditionally) - now ships there for the first time.

`cryptsetup-var.service`'s unchanged `Requires=`/`After=dev-disk-by\x2dpartlabel-var.device` block still waits on a partlabel device this MBR machine structurally can never produce - this PR's own docs call that gap out explicitly ("var.encrypt does not yet work on this layout", `avocado-raspberrypi.inc:23-24`). Per `systemd.unit(5)`, ordering dependencies (`Requires=`/`After=`) are still enqueued and waited on even for a unit whose `ConditionPathExists=` will later skip it - the condition is only evaluated once the unit would actually start, not before the device wait. So every RPi initrd now enqueues a start job for a device that never appears, stalls for the full device timeout (90s by default), and logs a dependency failure - a new boot-time regression on every Raspberry Pi image, independent of whether anyone ever sets `var.encrypt`.

Jetson has an escape hatch (`avocado-tegra-init` masks the unit outright); Raspberry Pi doesn't. Worth downgrading `Requires=` to `Wants=` on the service, or gating its `.wants` install on the partlabel actually being reachable for the machine.

mobileoverlord:
Hardware pass on imx8mp-evk (OTA turning on `var.encrypt` from a plaintext runtime) found three things, all folded into the commit (`00bd7e9081`):

- `cryptsetup-var.service` needs `Before=var.mount`: `by-avocado/var` resolves to the plaintext partition immediately, so the initrd's fstab-generated `var.mount` grabbed it and the in-place reencrypt failed with "already mapped or mounted". The old `/dev/mapper/var` entry got that ordering for free.
- `OnFailure=emergency.target`: the boot then fell open to a plaintext `/var` despite the marker.
- The grown-filesystem shrink now targets used + 25% + 512 MiB instead of just the 32 MiB header, so the conversion covers the data in use (a 30 GB grown `/var` with 200 MB of data was about to convert 29 GB). growfs re-expands at the first mount of the mapping. Host test extended (13/13).

Also added a Transitions section to the doc (what can and cannot be reached by OS update).

mobileoverlord:
Hardware pass complete on imx8mp-evk, all via `avocado deploy` (OS bundle to the inactive slot, avocadoctl #24 on the device):

1. Baseline runtime without `var.encrypt`: `/var` plaintext through `/dev/disk/by-avocado/var -> mmcblk2p9`, `cryptsetup-var` not run, rootfs verity live (`root=/dev/mapper/root`).
2. Deploy of a runtime with `var.encrypt: true` (slot B): first boot shrank the grown 30 GB btrfs to allocated+1.5 GiB, reencrypted in place (3.9 GB instead of 29 GB), `/var` from `/dev/mapper/var`, LUKS2 Argon2id keyslot, posture `argon2id`, U-Boot KV `avocado_var_encrypted=yes` / `avocado_var_unlock=argon2id`, data intact, 0 failed units.
3. Next deploy (slot A) with the fixed `base-files` rule: clean reopen with no intervention — `opening LUKS2 container` -> `/dev/mapper/var ready` -> `Mounted /var`, `by-avocado/var -> dm-1` created in the initrd, 0 failed units.

Not yet observed on hardware: the `cryptsetup-var-posture` preset (`SYSTEMD_PACKAGES += "${PN}-posture"`, last fixup) — the board's feed predates it; the publisher runs correctly when started by hand.

meta-avocado-raspberrypi/conf/machine/include/avocado-raspberrypi.inc
```diff
@@ -16,15 +16,12 @@ CMDLINE:append = " audit=0"
...
-# only appears once cryptsetup-var has opened the container - so the mount waits
-# for the unlock instead of racing it and dropping to emergency mode. Unlike
-# avocado-qemuarm64/qemux86-64/imx93-frdm, the non-encrypted-var branch stays the
-# raw /dev/mmcblk0p3 rather than /dev/disk/by-partlabel/var: this machine is
-# MBR-partitioned (stone/rpi/rootdisk.conf), which carries no GPT PARTLABEL, so a
-# by-partlabel path would not resolve here.
-AVOCADO_VAR_PART_DEV = "${@bb.utils.contains('DISTRO_FEATURES', 'encrypted-var', '/dev/mapper/var', '/dev/mmcblk0p3', d)}"
+# The plaintext var partition; /var mounts /dev/disk/by-avocado/var, which
+# resolves here unless the initrd opened a LUKS mapping over it. A raw node
+# rather than /dev/disk/by-partlabel/var: this machine is MBR-partitioned
```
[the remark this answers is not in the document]
mobileoverlord: Taken option (a) in 8349e0844d: Raspberry Pi now declares `AVOCADO_SECURITY_CAPABILITIES = ""` (explicitly nothing) until a dm-crypt fragment and a GPT `var` label exist for it, with the reason in the conf and the doc. That also stops `cryptsetup-var` from being pulled into its images at all.

meta-avocado/recipes-core/cryptsetup-var/files/cryptsetup-var.service
mobileoverlord: The writer exists: avocado-cli `main` commit 083fbbb (#216, merged 2026-08-27) - `runtimes.<name>.var.encrypt` writes `/etc/avocado/var-encrypt` into the runtime's initramfs (`src/commands/initramfs/image.rs`). Your clone predates it. Exercised on imx8mp-evk with that cli: a runtime without it boots plaintext, one with it encrypted in place over `avocado deploy` (results in the PR comments). PR description updated to name the dependency. On the silent-skip point: the skip is the intended no-opt-in path, and a runtime that did opt in now fails closed (`OnFailure=emergency.target`) rather than silently mounting plaintext.

meta-avocado/recipes-avocado/packagegroups/packagegroup-avocado-initramfs.bb
mobileoverlord: You were right, and it bit on hardware exactly as described (second boot: `Can't lookup blockdev`, `by-avocado/` absent in the initrd). Fixed in 8349e0844d by moving the dm claim into the rule `base-files` generates (`61-avocado-var.rules`), which is in both the initramfs and the rootfs; `99-zz` stays for the post-switch-root re-assert. Verified: clean reopen on the next boot with no intervention, `by-avocado/var -> dm-1` created in the initrd.

meta-avocado-distro/recipes-core/base-files/base-files_%.bbappend
mobileoverlord: Confirmed on hardware (first boot: `Failed to open ... in exclusive mode (already mapped or mounted)`) and fixed in 8349e0844d: `cryptsetup-var.service` has `Before=var.mount` and `OnFailure=emergency.target`, so the race is gone and a failure can no longer fall through to a plaintext mount. Posture on the encrypted device reads `avocado_var_encrypted=yes` / `avocado_var_unlock=argon2id` in the U-Boot KV.

meta-avocado-distro/recipes-core/base-files/base-files_%.bbappend
mobileoverlord: Done in 8349e0844d: the bare-node branch now requires no `/` after `/dev/`, `/dev/disk/by-partuuid/<uuid>` is handled alongside `PARTUUID=`, and every other form reaches the `bb.fatal`. Both docs list the accepted forms and say the rest fail the build.

meta-avocado-distro/recipes-core/base-files/base-files_%.bbappend
mobileoverlord: 1. Done in 8349e0844d: `ACTION=="add|change"` and `ENV{ID_FS_TYPE}=="?*"` before the negative match, on both lines. 2. Agreed it is unscoped; a static rule cannot say "the var partition on the boot disk", which is why Jetson keeps its own init. Documented as a known follow-up in `docs/security-capabilities.md` rather than solved here - the fix needs a disk identity the build can pin (e.g. the stone manifest's GPT disk GUID as `ID_PART_TABLE_UUID`), which is a separate change.

meta-avocado-raspberrypi/conf/machine/include/avocado-raspberrypi.inc
mobileoverlord: Resolved by dropping the declaration (8349e0844d): Raspberry Pi declares `""` until both the kernel fragment and the GPT label are wired, and the doc names both prerequisites explicitly.

meta-avocado/recipes-avocado/packagegroups/packagegroup-avocado-rootfs.bb
mobileoverlord: Resolved by the same change (8349e0844d): with no declaration, `cryptsetup-var` is not pulled into the Raspberry Pi initramfs or rootfs, so the unit and its `Requires=` on the partlabel device never ship there. Kept `Requires=` for machines that do declare: on a GPT layout the device exists, and downgrading to `Wants=` would let a genuinely missing partition slip through.

jetm (approved):

PR #336, opened 2026-08-28
cryptsetup-var: --device-size is the data to convert, not data plus the header shift

mobileoverlord:
First-boot in-place encryption on a freshly provisioned imx8mp-evk failed after the filesystem shrink succeeded:

```
cryptsetup-var: first boot - encrypting existing btrfs ... in place (--device-size 352M)
Reduced data size is larger than real device size.
```

`--device-size` names the data to convert; cryptsetup adds the `--reduce-device-size 32M` shift itself and refuses when data + shift exceeds the partition. The script passed filesystem + 32 MiB, which only fit while the partition was much larger than the filesystem (every earlier test). Pass the filesystem size. Host test (14/14) updated.


mobileoverlord:
Pushed directly to wrynose as cd4652e70f (rebased onto the Jetson preset commit) at the maintainer's request; verified on a freshly provisioned imx8mp-evk: shrink -29M, --device-size 320M, /var on /dev/mapper/var, posture published.

PR #343, opened 2026-08-29
encrypted-var: operator recovery keyslot and var.hardware requirements

mobileoverlord:
Stacked on #342 (CAAM backend) — review the last commit only.

The initramfs side of the operator-held `/var` recovery key (avocado-cli `avocado var-key enroll`, avocadoctl `var-key`):

- **Volume key linked on every open** (`--link-vk-to-keyring @u::%user:cryptsetup:var`). No keyslot passphrase is reachable from the running system — the hardware backends and the derived key live only in this initramfs — so this is what lets avocadoctl add/remove the recovery keyslot later via `luksAddKey --volume-key-keyring`. A `user` key because cryptsetup must read it back; root on the device could already reach the volume key through the live dm-crypt mapping.
- **Derived slot retired once a recovery token exists.** With an operator key as the recovery path, the SoC-UID-derived Argon2id slot is the one key anyone who can read the UID could derive. Kept when it is what opened `/var` this boot (engine failure), so a device is never stranded.
- **`var.hardware` honoured** via `/etc/avocado/var-hardware` (written by the cli only when not `auto`): `caam`/`tpm2` fail closed *before* the container is touched when that engine is unusable (→ emergency target); `none` enrols no hardware keyslot. `auto` is today's behaviour.
- Posture: `avocado_var_recovery = key | soc-uid`; `VAR_HARDWARE` in the /run file. Docs updated.

**Tests:** host test 35/35 — keyring link asserted on every open path; retire / keep-when-it-opened / no-retire cases; caam-required-and-down refuses with no cryptsetup call; `none` enrols nothing.

Companions: avocado-linux/avocadoctl `jschneck/var-key`, avocado-cli `jschneck/var-recovery-key`.

### DIFF
--- a/meta-avocado-nxp/recipes-core/cryptsetup-var/tests/test-var-hwkey.sh
+++ b/meta-avocado-nxp/recipes-core/cryptsetup-var/tests/test-var-hwkey.sh
@@ -45,7 +45,7 @@ sh "$backend" new "$work/blob" || bad "new failed"
 [ "$(wc -l < "$work/blob")" = 1 ] && ok "blob is one text line" || bad "blob lines: $(wc -l < "$work/blob")"
 [ "$(openssl base64 -d -A -in "$work/blob")" = "BLOB-OF-avocado-var-kek" ] && ok "blob is the base64 of caam-keygen's black blob" || bad "blob content: $(cat "$work/blob")"
 grep -q "^caam-keygen create avocado-var-kek ccm -s 32$" "$LOG" && ok "key is a random 256-bit CCM black key" || bad "unexpected caam-keygen argv: $(grep caam-keygen "$LOG")"
-[ -e "$work/caam/avocado-var-kek" ] || [ -e "$work/caam/avocado-var-kek.bb" ] && bad "black key material left in the working dir" || ok "working dir scrubbed after new"
+[ ! -e "$work/caam/avocado-var-kek" ] && [ ! -e "$work/caam/avocado-var-kek.bb" ] && ok "working dir scrubbed after new" || bad "black key material left in the working dir"
 
 # --- derive ---
 sh "$backend" derive "$work/blob" > "$work/pass" || bad "derive failed"
--- a/meta-avocado-nxp/recipes-core/cryptsetup-var/tests/test-var-hwkey.sh
+++ b/meta-avocado-nxp/recipes-core/cryptsetup-var/tests/test-var-hwkey.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+# Host test for the i.MX 8M CAAM /var key backend (files/var-hwkey.sh). No CAAM
+# here: caam-keygen and caam-crypt are stubs that record their argv and write
+# fixed bytes, so what is asserted is the contract the backend keeps with
+# cryptsetup-var.sh and with NXP's tools - the blob round-trips through the
+# token as one text line, derive is a pure function of the blob with a fixed
+# input and IV, and every failure is a refusal rather than a wrong passphrase.
+set -euo pipefail
+here="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+backend="$here/../files/var-hwkey.sh"
+pass=0; fail=0
+ok()  { printf '  ok   - %s\n' "$1"; pass=$((pass+1)); }
+bad() { printf '  FAIL - %s\n' "$1"; fail=$((fail+1)); }
+command -v openssl >/dev/null 2>&1 || { echo "SKIP: host missing openssl"; exit 0; }
+
+work="$(mktemp -d)"; trap 'rm -rf "$work"' EXIT
+mkdir -p "$work/bin" "$work/caam" "$work/run"
+export PATH="$work/bin:$PATH" LOG="$work/log" AVOCADO_CAAM_DIR="$work/caam" AVOCADO_PROC_CRYPTO="$work/proc-crypto" TMPDIR="$work/run"
+: > "$LOG"
+# caam-keygen create <name> ccm -s 32 -> <dir>/<name> (black key) + <name>.bb (blob)
+cat > "$work/bin/caam-keygen" <<S
+#!/bin/sh
+echo "caam-keygen \$*" >> "\$LOG"
+[ "\$1" = create ] || exit 1
+printf 'BLACKKEY' > "$work/caam/\$2"; printf 'BLOB-OF-%s' "\$2" > "$work/caam/\$2.bb"
+S
+# caam-crypt enc AES-256-CBC -k <blob> -in <f> -out <f> -iv <hex>: 80 bytes keyed on the blob content
+cat > "$work/bin/caam-crypt" <<'S'
+#!/bin/sh
+echo "caam-crypt $*" >> "$LOG"
+[ -s "$4" ] || exit 1
+grep -q '^BLOB-OF-' "$4" || { echo "import failed"; exit 1; }
+{ head -c 64 /dev/zero | tr '\0' 'P'; head -c 16 /dev/zero | tr '\0' 'Q'; } > "$8"
+echo "chatter"
+S
+chmod +x "$work/bin/"*
+printf 'name         : tk(cbc(aes))\n' > "$work/proc-crypto"
+
+# --- probe ---
+sh "$backend" probe >/dev/null 2>&1 && bad "probe passed without /dev/caam-keygen" || ok "probe refuses without the caam-keygen device"
+[ "$(sh "$backend" name)" = caam ] && ok "backend name is caam" || bad "name: $(sh "$backend" name)"
+
+# --- new ---
+sh "$backend" new "$work/blob" || bad "new failed"
+[ "$(wc -l < "$work/blob")" = 1 ] && ok "blob is one text line" || bad "blob lines: $(wc -l < "$work/blob")"
+[ "$(openssl base64 -d -A -in "$work/blob")" = "BLOB-OF-avocado-var-kek" ] && ok "blob is the base64 of caam-keygen's black blob" || bad "blob content: $(cat "$work/blob")"
+grep -q "^caam-keygen create avocado-var-kek ccm -s 32$" "$LOG" && ok "key is a random 256-bit CCM black key" || bad "unexpected caam-keygen argv: $(grep caam-keygen "$LOG")"
+[ -e "$work/caam/avocado-var-kek" ] || [ -e "$work/caam/avocado-var-kek.bb" ] && bad "black key material left in the working dir" || ok "working dir scrubbed after new"
+
+# --- derive ---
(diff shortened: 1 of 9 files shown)


meta-avocado-nxp/recipes-core/cryptsetup-var/tests/test-var-hwkey.sh
```diff
@@ -0,0 +1,63 @@
...
+# --- probe ---
+sh "$backend" probe >/dev/null 2>&1 && bad "probe passed without /dev/caam-keygen" || ok "probe refuses without the caam-keygen device"
+[ "$(sh "$backend" name)" = caam ] && ok "backend name is caam" || bad "name: $(sh "$backend" name)"
+
+# --- new ---
+sh "$backend" new "$work/blob" || bad "new failed"
+[ "$(wc -l < "$work/blob")" = 1 ] && ok "blob is one text line" || bad "blob lines: $(wc -l < "$work/blob")"
+[ "$(openssl base64 -d -A -in "$work/blob")" = "BLOB-OF-avocado-var-kek" ] && ok "blob is the base64 of caam-keygen's black blob" || bad "blob content: $(cat "$work/blob")"
+grep -q "^caam-keygen create avocado-var-kek ccm -s 32$" "$LOG" && ok "key is a random 256-bit CCM black key" || bad "unexpected caam-keygen argv: $(grep caam-keygen "$LOG")"
+[ -e "$work/caam/avocado-var-kek" ] || [ -e "$work/caam/avocado-var-kek.bb" ] && bad "black key material left in the working dir" || ok "working dir scrubbed after new"
```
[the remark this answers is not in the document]
mobileoverlord: Rewritten in 9d6dfbcfe7, though not for the stated reason. `||` and `&&` have equal precedence and are left-associative in POSIX sh, so the line parsed as `(([ -e kek ] || [ -e kek.bb ]) && bad) || ok` — with either file present `bad` did run. Verified all three cases against the real helpers before changing anything.

It was still worth changing: it was the one assertion in the file written inside out rather than the house `cond && ok || bad` idiom, and its correctness quietly depended on `bad()` returning 0 — if `bad` ever gained an `exit 1` the same case would print both FAIL and ok. It now reads:

    [ ! -e "$work/caam/avocado-var-kek" ] && [ ! -e "$work/caam/avocado-var-kek.bb" ] && ok "..." || bad "..."

Checked in both directions: passes as-is, and fails with "black key material left in the working dir" when the stub is made to leave the key behind.

meta-avocado/recipes-core/cryptsetup-var/tests/test-cryptsetup-var-inplace.sh
```diff
@@ -130,4 +157,72 @@ s="$work/c7"; mkdir -p "$s"; echo btrfs > "$s/fstype"; echo 2147483648 > "$s/fsb
...
+run "$s" || bad "case 11 script exit"
+[ -e "$s/token" ] && bad "token imported although the keyslot was not added" || ok "no orphan token after a failed keyslot add"
+
+# --- Case 13: recovery slot enrolled by the operator and the engine opened /var
+# -> the derived (untokened) Argon2id slot 0 is retired; volume key linked ---
+s="$work/c13"; mkdir -p "$s"; touch "$s/luks" "$s/hwkey" "$s/recovery"; cp "$work/c8/token" "$s/token"
+run "$s" || bad "case 13 script exit"
+grep -q "^cryptsetup luksKillSlot --batch-mode $blk 0$" "$s/log" && ok "derived keyslot retired once a recovery slot exists and the engine opened /var" || bad "derived slot kept: $(grep -c luksKillSlot "$s/log")"
+grep -q "^cryptsetup luksOpen --link-vk-to-keyring @u::%user:cryptsetup:var --key-file" "$s/log" && ok "volume key is linked into root's user keyring on open" || bad "no --link-vk-to-keyring: $(grep luksOpen "$s/log")"
+grep -q "^VAR_RECOVERY=key$" /dev/null 2>&1 || true
```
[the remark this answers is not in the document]
mobileoverlord: Good catch, and it was three lines rather than one — the same dead-assertion shape appears at the old lines 170 (`grep ... "$s/posture"`, a file nothing ever wrote), 199 (the `/dev/null` one you flagged) and 226 (`"$s/out"`, which never carries the posture output). All three ended in `|| true`, so none could ever fail.

Rather than delete them, 9d6dfbcfe7 makes them real. The root cause was that `cryptsetup-var.sh` writes its posture to `/run/avocado-var-posture`, and the test mounts a tmpfs on `/run` inside its own mount namespace — so the file died with the namespace before anything could read it. `run()` now clears a stale copy and lifts the file out before leaving the namespace, and the three lines assert properly: `VAR_HWKEY_TOKEN=yes` on case 8, `VAR_RECOVERY=key` on case 13, `VAR_HARDWARE=none` on case 18.

Each was checked to fail when its expected value is changed, so they are not no-ops in a new disguise. 35 -> 38 assertions, suite green.

meta-avocado/recipes-core/cryptsetup-var/files/cryptsetup-var.sh
```diff
@@ -78,6 +78,123 @@ has_tpm2_token() {
...
+# is how `avocadoctl var-key` authorises adding or removing the operator's
+# recovery keyslot later: luksAddKey --volume-key-keyring. A `user` key, not
+# `logon`, because cryptsetup has to read it back; root on the device could
+# already reach the volume key through the running dm-crypt mapping.
+VK_LINK="--link-vk-to-keyring @u::%user:cryptsetup:var"
+RECOVERY_TOKEN_TYPE="avocado-recovery"
+
+# avocado.yaml runtimes.<r>.var.hardware, written by avocado-cli next to the
+# var-encrypt marker only when it is not the default "auto": caam|tpm2 = that
+# engine must hold a keyslot, refuse to boot on the derived key if it cannot;
```
lee-reinhardt, line 107: blocker: `caam` and `tpm2` do not actually fail closed after this preflight. This only proves that the CAAM backend probes or `/dev/tpm0` exists; `open_var` still falls back to the Argon2id key when an existing hardware token cannot unlock, and both enrollment helpers remain best-effort. Preserve the derived-key open needed for first enrollment, but abort an explicit mode if enrollment fails and reject derived fallback once a hardware token exists.

lee-reinhardt (changes requested):

meta-avocado/recipes-core/cryptsetup-var/files/cryptsetup-var.sh
mobileoverlord: You are right on all three counts, and the negative check makes it concrete: pointed at the previous script, the new cases fail with

    FAIL - opened /var although the required caam keyslot failed
    FAIL - opened on the derived key: cryptsetup luksOpen ... --key-file /run/tmp.nOV6zwN1gq /dev/loop0 var
    FAIL - completed the boot although the caam keyslot could not be enrolled
    FAIL - left /var open: 0 luksClose
    FAIL - opened /var although the required TPM2 keyslot failed
    FAIL - opened on the derived key: cryptsetup luksOpen ... --key-file /run/tmp.t9NDKb8qol /dev/loop0 var
    FAIL - completed the boot although the TPM2 keyslot could not be enrolled

Fixed in a5db71ada2. Every path that could end the boot on the derived key now routes through `fail_closed()`, which names the unmet requirement, closes the mapping and fails the unit:

- `open_var`: a hardware or TPM2 token that exists but will not unlock no longer falls through to the Argon2id open under the matching explicit mode. This is the one the preflight structurally cannot catch — the engine is up, the stored blob just no longer derives — so the test grows a `hwkey_stale_blob` stub flag to reach it rather than reusing `hwkey_down`, which the preflight catches first.
- `ensure_hwkey_enroll`: no backend installed, engine unusable, blob not produced, `luksAddKey` refused, token import failed — all fatal under `caam`.
- `ensure_tpm2_enroll`: no TPM, no `systemd-cryptenroll`, refused enrolment — all fatal under `tpm2`.

The teardown is deliberate: the enrolments run after the open, so `/var` can already be open on the derived key by the time the requirement is known unmet. It is unconditional rather than guarded on `[ -b "$MAPPER" ]`, since that guard would skip the close in exactly the odd states where it matters. The preflight refusal keeps its own exit and never reaches `fail_closed`, so case 16's "nothing touched before the refusal" still holds.

First enrolment is preserved as you asked, and keys on the absence of the token rather than the mode — before a hardware token exists the derived key is the only way into the container and the only key `luksAddKey` can authenticate with. Case 24 asserts that open still happens under `tpm2`.

`var.hardware=auto` is unchanged: cases 20 and 22 assert it still degrades and reports. `TPM_DEV` is overridable purely so the host test can drive the TPM paths without a TPM; the production default is `/dev/tpm0`. 38 -> 48 assertions, all three suites green.

The docs at `docs/security-capabilities.md:148` claimed this behaviour already, so they are rewritten to state what is now true, including the first-enrolment carve-out.

meta-avocado/recipes-core/cryptsetup-var/files/cryptsetup-var.sh
```diff
@@ -142,25 +308,47 @@ write_posture() {
 # permanently.
 open_var() {
     echo "cryptsetup-var: opening LUKS2 container on $VAR_DEV"
+    if has_hwkey_token; then
```
lee-reinhardt: Blocker: `open_var` returns after any hardware-key token succeeds, before it checks TPM2. If `auto` previously enrolled both token types and the configuration later requires `tpm2`, a working CAAM slot bypasses a stale or failed TPM2 token; `ensure_tpm2_enroll` then returns because the token exists, so the boot succeeds without the required engine. I reproduced this with a mixed-token host-test case: all 48 existing assertions pass, but the script exits successfully through CAAM with `var.hardware=tpm2` and a failing TPM2 unseal. In explicit modes, attempt only the selected engine, allowing the derived path only while that selected token is absent for first enrollment.

mobileoverlord: Confirmed and fixed. You're right that the preflight only proved the engine probes, not that it is what opened the container — and `open_var` returned on the first token that worked regardless of the configured mode.

The engine is now selected up front rather than discovered by trying: `tpm2` attempts only TPM2, an explicit hardware backend only that one, `auto` tries both as before, `none` neither. I also covered the case your repro implies but does not hit: when the selected engine has no token at all, the script refuses if a token of the *other* kind proves the device is already enrolled under the wrong engine. A genuine first-enrolment boot, where neither token exists, still opens on the derived key as it has to.

Three host cases added, and the dump stub grows a `systemd-tpm2` token so a device can carry both: the mixed-token device with a failing TPM2 unseal is refused and the hardware blob is never read at all; the mirror case (`caam` on a TPM2-enrolled device) is refused; and `auto` on the same mixed device still opens with the hardware key without reaching the failing TPM2. 52 assertions pass.

lee-reinhardt (approved):

PR #358, opened 2026-09-02
ENG-2198: set CVE_PRODUCT on imx-atf

hiagofranco:
Without `CVE_PRODUCT`, cve-check falls back to `${BPN}` — `imx-atf`, which no advisory database indexes. So the secure firmware chain scans clean on every i.MX board, and a zero there currently means "never looked" rather than "nothing found".

A bbappend rather than a recipe edit, because meta-imx's and meta-freescale's `imx-atf_2.10.bb` do not set it either.

The two products are measured against cve-check's NVD snapshot, not copied from meta-arm's `trusted-firmware-a.inc` — three of the four products it names match no records. Kept vendor-qualified to exclude the `amd:` and `renesas:` forks.

Effect: imx95 (PV 2.12) gains a scan and no finding; boards on 2.10 gain one unpatched finding, CVE-2023-31339. Record counts, per-version numbers and the rest of the boot chain are on [ENG-2198](https://linear.app/peridio/issue/ENG-2198/bootloader-and-firmware-in-the-sbom).

Not verified by a build. Without this the recipe scans under a name that matches nothing, which is exactly today's behaviour.

### DIFF
--- a/meta-avocado-nxp/recipes-bsp/imx-atf/imx-atf_%.bbappend
+++ b/meta-avocado-nxp/recipes-bsp/imx-atf/imx-atf_%.bbappend
@@ -1,0 +1,4 @@
+# NVD indexes NXP's TF-A fork under these two names. The ${BPN} default matches
+# nothing, so the secure firmware chain scans clean on every i.MX board. Kept
+# vendor-qualified to exclude the amd: and renesas: forks.
+CVE_PRODUCT = "trustedfirmware:trusted_firmware-a arm_trusted_firmware_project:arm_trusted_firmware"


hiagofranco:
Generated the description with Claude Opus 5, tested and it looks good for me.

jetm (commented):
Reviewed. Verified both CPE identifiers against the live NVD CVE database (`trustedfirmware:trusted_firmware-a`, `arm_trusted_firmware_project:arm_trusted_firmware` are both real and match the recipe's CVEs), confirmed the mechanism reaches the SBOM path via `spdx-common.bbclass` and `get_cpe_ids()`, checked version normalization (`+git${SRCPV}` stripping) resolves to the correct CPE versions for both in-tree imx-atf providers, and confirmed no existing bbappend collision. No blocking issues.

jetm:
@nicksinas @mobileoverlord Need review or approval on this. I don't have permission to approve/merge it myself.

jetm (approved):

hiagofranco:
Force-pushed: signed commit.

nicksinas:
CodeQL only runs on a schedule for forks, so the code-scanning rule never gets a result on these. Bypassing and merging as deemed as compliant.

PR #359, opened 2026-09-02
ENG-2346: count the CVEs that reach the device, not the ones in pkgdata

hiagofranco:
`packaged_cves` is keyed on pkgdata membership, which is the wrong set in both directions. A deploy-only bootloader packages nothing and ships anyway, so it is missing from the count; a `nativesdk` recipe packages something and reaches no device, so it is inside it. The README tells people to read that number.

Adds sibling counters `device_recipes` / `device_cves` — every recipe whose scope is anything but `build-only` — and leaves `packaged_*` alone. Redefining a counter requires a new major under the layer's compatibility policy, and two codebases read this one. A new counter is allowed at version 1, so consumers are unaffected.

Also names each board's boot chain in `AVOCADO_CVE_BOOT_CHAIN`. The default names three generic recipes; i.MX and Jetson build none of them, and `imx-atf` and `u-boot-imx` scoped `feed` on the strength of their `-dbg` packages. Appended rather than set, or the default's `optee-os` goes with it.

## What this does not do

**It does not make the boot chain covered.** On the `avocado-imx95-frdm` tree all seven i.MX boot recipes carry zero unpatched CVEs, so none reaches `recipes` and the report scopes `boot-chain: 0`. That is ENG-2198's and ENG-2196's gap — no `CVE_PRODUCT`, no NVD record — and the list goes live once the `imx-atf` fix lands. The lists fix the label; they add no CVE to any count, since `feed` and `boot-chain` are both non-`build-only`.

## Verification

- 142 unit tests. Both directions are pinned: a deploy-only recipe is in `device_cves` and not `packaged_cves`, a `nativesdk` one the reverse.
- `test_a_freshly_built_report_passes_the_checker` runs `verify.check_report` over a report `build_report` just produced. The emitted counts block was the one file of the policy's four that nothing bound — a counter could reach Stats, the schema and the checker and still be missing from the document.
- `verify --strict` passes on real qemuarm64 and imx95 reports.
- Every boot-chain name checked against its layer: the i.MX seven against the imx95 build tree, the RPi and Tegra names against the kas-pinned vendor commits.

Not verified: no build on Raspberry Pi or Jetson, so those two lists are name-checked only. On both trees measured `device_cves` equals `packaged_cves`, so the divergence is proven in unit tests rather than on a real build.

### DIFF
--- a/meta-avocado-nvidia/conf/machine/include/avocado-jetson.inc
+++ b/meta-avocado-nvidia/conf/machine/include/avocado-jetson.inc
@@ -31,3 +31,10 @@ IMAGE_TEGRAFLASH_DATA = "${DEPLOY_DIR_IMAGE}/avocado-image-var-${MACHINE_SHORT_N
 PACKAGECONFIG:remove:pn-wpewebkit = "gbm"
 DEPENDS:append:pn-wpewebkit = " libdrm"
 DEPENDS:append:pn-chromium-ozone-wayland = " virtual/libgbm"
+# The Tegra boot chain. meta-tegra spells TF-A arm-trusted-firmware, not the
+# trusted-firmware-a the AVOCADO_CVE_BOOT_CHAIN default names, so the default
+# matches nothing here; optee-os in it does resolve, to meta-tegra's l4t recipe.
+AVOCADO_CVE_BOOT_CHAIN:append = " arm-trusted-firmware edk2-firmware-tegra \
+                                  edk2-firmware-tegra-rcmboot tos-optee \
+                                  tegra-bootfiles tegra-firmware \
+                                  standalone-mm-optee-tegra"
--- a/meta-avocado-nxp/conf/machine/include/avocado-imx.inc
+++ b/meta-avocado-nxp/conf/machine/include/avocado-imx.inc
@@ -1,1 +1,9 @@
 MACHINEOVERRIDES =. "avocado-imx:"
+
+# The i.MX boot chain under its own recipe names - AVOCADO_CVE_BOOT_CHAIN's
+# default names u-boot and trusted-firmware-a, neither of which this board
+# builds. Appended rather than set, so the default's optee-os survives.
+# imx-atf and u-boot-imx emit -dbg/-dev packages, so without this they scope
+# feed: real bootloader CVEs filed as "not in this image".
+AVOCADO_CVE_BOOT_CHAIN:append = " imx-atf imx-boot u-boot-imx imx-system-manager \
+                                  imx-oei firmware-imx imx-boot-firmware-files"
--- a/meta-avocado-sbom/README
+++ b/meta-avocado-sbom/README
@@ -121,7 +121,12 @@ Variables
                                names them. Defaults to "u-boot
                                trusted-firmware-a optee-os". Override per
                                machine: a board with a different
-                               secure-firmware stack has a different list
+                               secure-firmware stack has a different list.
+                               avocado-imx.inc, avocado-raspberrypi.inc and
+                               avocado-jetson.inc append theirs - the default
+                               names three generic recipes and i.MX builds
+                               none of them. Append rather than set, or the
+                               default's optee-os goes with it
 
 Evidence
 --------
@@ -213,9 +218,15 @@ in three separate directions:
     the device.
 
 That last point is why boot-chain is decided by recipe name, from
-AVOCADO_CVE_BOOT_CHAIN, and checked before package membership. It also makes
-the label independent of whether a given machine's u-boot sets PACKAGES = ""
-(see "packaged" below): the bootloader scopes boot-chain either way.
+AVOCADO_CVE_BOOT_CHAIN. It also makes the label independent of whether a given
+machine's u-boot sets PACKAGES = "" (see "packaged" below): the bootloader
+scopes boot-chain either way.
+
+An installed package still wins over the name. firmware-imx and tegra-firmware
+are boot chain and also ship packages the rootfs installs, and base-runtime is
+the set a consumer filters on to ask what is in the rootfs, so they must not
+leave it. Naming a recipe in AVOCADO_CVE_BOOT_CHAIN therefore only ever adds
+scope to a recipe no manifest names; it never moves one out of base-runtime.
 
 The coverage denominator does not move. unscanned_recipes and
 no_cve_record_recipes are derived from pkgdata - everything this build produced
@@ -230,6 +241,14 @@ Per-scope totals are deliberately not counters. Every recipe entry carries its
 own scope, so any aggregate a consumer wants is one pass over "recipes" and
 cannot drift from the entries it summarises - which is the failure the derived
 -count checks in verify.py exist to catch for the counters that do exist.
+
+device_recipes and device_cves are not an exception to that. They are not a
+per-scope total but the partition the report exists to answer - on the device
+or not - and they are counters for the one reason a per-scope breakdown is not:
+every consumer needs that number, so leaving it out puts the same pass in each
+of them, spelt differently. They are derived after the scope pass and
+re-derived by verify.py from the entries, so they are held to the same standard
+as the aggregates that are not counters.
 
 When no manifest is read, manifests_read is 0 and every packaged recipe scopes
 "feed". do_cve_report is ordered before do_build only, so it can legitimately
@@ -511,9 +530,10 @@ Requires a new major:
 
 Adding a counter means updating four files in one change: the Stats namespace
 and the counts block in report.py, the schema, REPORT_COUNTERS in verify.py,
-and the counts block in tests/fixtures/make-fixture.py. A test asserts the
-schema and the checker agree, so a change that touches only one of the first
-three fails before it reaches a consumer. make-fixture.py has no such test: it
+and the counts block in tests/fixtures/make-fixture.py. Tests bind the schema,
+the checker and Stats to each other, so a change that touches only some of the
+first three fails before it reaches a consumer. make-fixture.py has no such
+test - binding it would mean running it, which needs a real report - and it
 builds counts from its own key list, so a counter missing there is only found
 when someone regenerates the fixture and CI reports it missing. It does import
 HEALTH_COUNTERS from verify rather than restating it, so at least the question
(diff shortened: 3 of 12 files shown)


meta-avocado-nxp/conf/machine/include/avocado-imx.inc
```diff
@@ -1 +1,9 @@
 MACHINEOVERRIDES =. "avocado-imx:"
+
+# The i.MX boot chain under its own recipe names - AVOCADO_CVE_BOOT_CHAIN's
+# default names u-boot and trusted-firmware-a, neither of which this board
+# builds. Appended rather than set, so the default's optee-os survives.
+# imx-atf and u-boot-imx emit -dbg/-dev packages, so without this they scope
+# feed: real bootloader CVEs filed as "not in this image".
+AVOCADO_CVE_BOOT_CHAIN:append = " imx-atf imx-boot u-boot-imx imx-system-manager \
+                                  imx-oei firmware-imx imx-boot-firmware-files"
```
jetm, line 9: **`firmware-imx` mislabels an installed rootfs recipe as boot-chain-only.**

`_scope()` (`report.py:153`) tests boot-chain membership before package membership, so naming a recipe that ships installed runtime packages silently moves it out of `base-runtime` into `boot-chain` instead of adding it to `device_cves` for free. `firmware-imx` is not deploy-only: at the pinned `meta-freescale@0c27c21c`, `conf/machine/include/imx-base.inc:436` puts `firmware-imx-easrc-imx8mn firmware-imx-xcvr-imx8mp` into `MACHINE_FIRMWARE`, which `:442` folds into `MACHINE_EXTRA_RRECOMMENDS` - so on `avocado-imx8mp-evk`, `avocado-imx8mp-var-dart` and `avocado-ucm-imx8m-plus` these packages land in the rootfs manifest.

A consumer that answers "what CVEs are in the shipped rootfs" by filtering `scope == "base-runtime"` (the split the README documents) silently loses every VPU/EPDC firmware CVE on imx8mp. Nothing catches this - no counter moves, so no test and no health counter flags it - and `_scope()`'s own docstring ("reach the device ... so no manifest names them") becomes false for this entry.

The same defect exists at `meta-avocado-nvidia/conf/machine/include/avocado-jetson.inc:39` for `tegra-firmware`: `meta-tegra@d2613418`'s `conf/machine/include/tegra-common.inc:82` sets `MACHINE_ESSENTIAL_EXTRA_RDEPENDS = "tegra-firmware ..."`, installed on all six Jetson machines - same silent loss of WiFi/USB firmware CVEs from `base-runtime`.

meta-avocado-nvidia/conf/machine/include/avocado-jetson.inc
```diff
@@ -30,4 +30,11 @@ IMAGE_TEGRAFLASH_DATA = "${DEPLOY_DIR_IMAGE}/avocado-image-var-${MACHINE_SHORT_N
...
 DEPENDS:append:pn-wpewebkit = " libdrm"
-DEPENDS:append:pn-chromium-ozone-wayland = " virtual/libgbm"
\ No newline at end of file
+DEPENDS:append:pn-chromium-ozone-wayland = " virtual/libgbm"
+# The Tegra boot chain. meta-tegra spells TF-A arm-trusted-firmware, not the
+# trusted-firmware-a the AVOCADO_CVE_BOOT_CHAIN default names, so the default
+# matches nothing here; optee-os in it does resolve, to meta-tegra's l4t recipe.
+AVOCADO_CVE_BOOT_CHAIN:append = " arm-trusted-firmware edk2-firmware-tegra \
+                                  edk2-firmware-tegra-rcmboot tos-optee \
+                                  tegra-bootfiles tegra-firmware \
```
jetm, line 39: Same defect as `meta-avocado-nxp/conf/machine/include/avocado-imx.inc:9` (see that comment for the full analysis) - `tegra-firmware` is installed via `MACHINE_ESSENTIAL_EXTRA_RDEPENDS` (`meta-tegra@d2613418` `conf/machine/include/tegra-common.inc:82`) on all six Jetson machines, not deploy-only, so declaring it here silently removes its CVEs from `base-runtime` instead of adding it to `boot-chain` for free.

meta-avocado-sbom/README
```diff
@@ -456,6 +471,18 @@ Report format
...
+device_cves is the headline: CVEs over every recipe whose scope is anything but
+build-only. Read that one, not packaged_cves.
+
+packaged_cves counts pkgdata membership, which is not the same question and is
+wrong in both directions for this one. A deploy-only bootloader packages
+nothing and ships anyway, so it is missing from it; a nativesdk recipe packages
+something and reaches no device, so it is inside it. Both counters are kept:
+the compatibility policy below makes redefining one a new major, and the
+pkgdata number is still what answers "how much of this correlated to something
+installed". Where they differ, the difference is exactly the two cases above,
```
jetm: **"exactly the two cases" denies a real third case the same file documents 15 lines below.**

The sentence claims `device_cves` and `packaged_cves` differ only by (a) a deploy-only bootloader and (b) a nativesdk recipe. A stale, dropped recipe is a third case: `README:496-505` says outright "stale_dropped cannot see it either - that counter only fires when pkgdata knows the recipe at some other version."

Concretely: a recipe the build no longer produces (after a `PREFERRED_PROVIDER_virtual/kernel` switch or an `AVOCADO_FEATURE_GROUPS` drop) is absent from pkgdata, so `report.py:264`'s `known = recipe_versions.get(name)` is `None` and the `if not known:` branch keeps the entry without incrementing `stale_dropped`. It's then `packaged: false` (excluded from `packaged_cves`) and falls through `_scope()` to `boot-chain` (included in `device_cves`) at `report.py:169`. So the number this PR promotes to headline can carry CVEs from recipes the image does not contain, `stale_dropped` reads 0, and the README tells the reader that cannot happen. A security reviewer reconciling the two counters on a real board finds a delta the doc says is impossible.

jetm (commented):
Reviewed against the pinned vendor commits (meta-tegra `d2613418`, meta-freescale `0c27c21c`, meta-imx `92ad51ef`, meta-raspberrypi `2c646d29`) and against a live BitBake 2.18 for the `.inc` conf-append/default-vs-override semantics the change relies on. Full test suite passes (142 tests) and `avocado_sbom.verify` runs clean on the updated fixture. Three blocking issues below, anchored where the diff allows.

One more that can't be anchored inline because the file it's about isn't touched by this PR: `meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb`'s `do_cve_report` task (around line 173) still prints `bb.plain("(%d recipes, %d CVEs packaged)")`, unchanged by this PR, even though `README:531` now claims `device_cves` is "the one the layer's own output leads with." The build-time console output - the surface that runs on every build - still headlines the packaged count this PR argues undercounts bootloader exposure. Separately, neither of the two CLIs this PR did update actually leads with the device number either: both print the unfiltered total first and put the device pair in a parenthetical, so the README's claim is false on all three surfaces today.

Also flagging, not blocking: `verify.py`'s `on_device` predicate (line 213) has zero test coverage - two independent finders mutation-proved it can be broken without failing anything, while the producer side in `report.py` is covered. The test fixture has `device_*` equal to `packaged_*` in every case, which is why the gap is invisible to the suite. And the two new required counters (`device_recipes`, `device_cves`) break compatibility with any archived pre-PR report per the schema's own `required` list, without the "made under these rules that they would otherwise forbid" callout the README uses elsewhere for schema exceptions.

meta-avocado-nxp/conf/machine/include/avocado-imx.inc
hiagofranco: Right that the ordering is the bug, though not by that route: `packagegroup-avocado-rootfs` deliberately does not RDEPEND on `packagegroup-core-boot` and drops `MACHINE_ESSENTIAL_EXTRA_RDEPENDS`, so `MACHINE_FIRMWARE` never reaches an avocado rootfs. The imx95 manifest confirms it - 158 packages, no `firmware-imx*`.

It gets installed anyway, by our own packagegroup: `packagegroup-avocado-imx-rootfs.bb:26` has `RDEPENDS:${PN}:append:mx8mp-nxp-bsp = " firmware-imx-sdma-imx7d"`, a `firmware-imx` subpackage, on imx8mp-evk / ucm-imx8m-plus / imx8mp-var-dart. So the defect is real on the boards you named.

Fixed at the root rather than by dropping the two names: `_scope()` now matches an installed package before the boot chain name. A boot-chain entry can only add scope to a recipe no manifest names; it can never move one out of `base-runtime`. Covers future entries too, so no hand-audit of `rpi-eeprom` and friends. Test added.

meta-avocado-nvidia/conf/machine/include/avocado-jetson.inc
hiagofranco: Same answer as the imx one. `tegra-firmware` does not arrive via `MACHINE_ESSENTIAL_EXTRA_RDEPENDS` - `packagegroup-avocado-rootfs` drops that variable on purpose - but it is installed, through `packagegroup-avocado-tegra-rootfs.bb:16` (`tegra-firmware-tegra234`) and `tegra-firmware-xusb` in the initramfs packagegroup. Fixed by the `_scope()` reorder.

meta-avocado-sbom/README
hiagofranco: Correct, and it contradicts a bullet in the same section. Reworded to "the two cases above plus the stale leftover below", and the bullet list header now says three things land in `packaged: false` instead of two.

hiagofranco:
Thanks - all four hold up. Amended rather than a follow-up commit.

**Scope ordering.** `_scope()` now matches an installed package before the boot chain name, so naming a recipe can only add scope to one no manifest names. `firmware-imx` and `tegra-firmware` stay in the boot-chain lists and stay `base-runtime` where they are installed. Route is not the one you cited - `packagegroup-avocado-rootfs` drops `MACHINE_ESSENTIAL_EXTRA_RDEPENDS` and never pulls `packagegroup-core-boot` - but our own packagegroups install both, so the defect was real. Details in the two inline replies.

**Console output.** All three lines now lead with the device pair: `avocado-cve-report.bb`, the report CLI and the verify CLI. On a regenerated imx95 report:

```
26 recipes, 4255 unpatched CVEs on the device (53 recipes, 4435 CVEs scanned; 26, 4255 packaged; 7027 packages)
```

**README delta.** Reworded to name the stale-leftover case, and the bullet header says three, not two.

**`on_device` coverage.** New test flips a packaged entry to `build-only` and drops the device counts. Mutation-checked: `on_device = packaged` fails it with `counts.device_cves is 5, but the report holds 6`.

**Required counters.** Added as a third callout in the compatibility policy, same shape as the `scope` one.

144 tests pass; `verify --strict` clean on regenerated qemuarm64 and imx95 reports.

jetm (approved):

jetm:

hiagofranco:
Force-pushed: signed commit.

jetm:

hiagofranco:
Fixed

jetm (approved):

PR #360, opened 2026-09-02
ENG-2197: filter kernel CVEs on what the configuration compiles

hiagofranco:
cve-check flags a kernel CVE when the vulnerable source is *present*, not when it is compiled. The kernel CNA records which files each CVE lives in, and the kernel's SPDX document records which files we build, so the two can be joined. `do_kernel_cve_improve` does that by running oe-core's `scripts/contrib/improve_kernel_cve_report.py`.

Measured on qemuarm64, linux-yocto 6.6.142+git:

| | unpatched | critical | high |
| -- | -- | -- | -- |
| raw cve-check | 795 | 43 | 347 |
| + version analysis alone | 2,310 | — | — |
| + compiled-files filter | 700 | 13 | 123 |

Read the middle row before calling this a filter. The script is an enrichment tool first — it merges the CNA's whole vulnerable set for our version into the report, which on its own trebles the kernel's findings. The compiled-files filter is what pays for that, and it is why the task refuses to run when the SPDX document carries no compiled sources: leaving cve-check's own results alone beats publishing the enriched set unfiltered.

## This lands inert

All three settings that arm it are commented out in `kas/feature/cve-check.yml`, so stacking the fragment changes nothing today: no kernel debug rebuild, no extra SPDX, no vulns clone, and the report carries cve-check's own kernel results.

They stay off until the pinned oe-core carries the fix for `improve_kernel_cve_report.py`'s unguarded `cve_data[cve]['detail']` read — the stock script raises `KeyError` on any real cve-check input, and the task treats a failing script as fatal. Merged on oe-core master as `f5da16b0d3c8`; the scarthgap backport is [on the list](https://lore.kernel.org/openembedded-core/20260901-fix-kernel-cve-scarthgap-v1-1-405f5d7a8957@baylibre.com/) and tracked under ENG-2414. Arming them lands as a follow-up commit together with the `kas/vendor/oe.yml` repin.

They go together deliberately: the DWARF and SPDX settings on their own would cost every build a `linux-yocto` debug rebuild and +63 MB of SPDX for a filter that is switched off.

## Notes for review

- Results go to `${CVE_CHECK_DIR}/kernel-improved/` and are read as an **overlay** — a file there replaces the same-named one, and cve-check's own output is never overwritten, so the before and after both stay on disk and stay comparable.
- No report schema change: filtered CVEs become `Ignored` and drop out of an Unpatched report on their own.
- The script is run as a subprocess, never imported: it is GPL-2.0 and this layer is Apache-2.0, the same boundary the README draws for sbom-cve-check.
- `--spdx` rather than the cheaper `--debug-sources-file`, because `read_debugsources` imports `zstandard`, which oe-core packages nowhere.
- 399 of the surviving findings are new, and 395 carry no CVSS — the kernel CNA assigns none. A consumer gating on severity has to decide what that means (ENG-2379, ENG-2385).
- 140 unit tests green. The end-to-end bitbake run is blocked on the oe-core fix above.

Full measurement history, including a review correction to the numbers, is on [ENG-2197](https://linear.app/peridio/issue/ENG-2197/config-aware-kernel-cve-filtering).

### DIFF
--- a/meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-kernel-vulns-native.bb
+++ b/meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-kernel-vulns-native.bb
@@ -1,0 +1,95 @@
+SUMMARY = "Keeps the kernel CNA's vulnerability data current for do_kernel_cve_improve"
+DESCRIPTION = "Clones and refreshes git.kernel.org/pub/scm/linux/security/vulns.git, \
+which records for each kernel CVE the versions and source files it affects. \
+avocado-cve-report reads it to decide which kernel CVEs apply to what this \
+configuration actually compiles."
+# Nothing from this recipe is built into or shipped with an image: it only
+# refreshes a checkout under DL_DIR that a build-time task reads. The data
+# itself is GPL-2.0-only (the kernel's own tree) plus MITRE's CVE terms of use,
+# both recorded in the repository's LICENSES directory.
+LICENSE = "GPL-2.0-only"
+
+INHIBIT_DEFAULT_DEPS = "1"
+# Reached only through do_kernel_cve_improve's [depends]. A world build must
+# not clone half a gigabyte to serve a task that may skip.
+EXCLUDE_FROM_WORLD = "1"
+
+inherit native nopackages avocado-nospdx
+
+deltask do_unpack
+deltask do_patch
+deltask do_configure
+deltask do_compile
+deltask do_install
+deltask do_populate_sysroot
+
+AVOCADO_CVE_KERNEL_VULNS_URL ?= "https://git.kernel.org/pub/scm/linux/security/vulns.git"
+AVOCADO_CVE_KERNEL_VULNS_DIR ?= "${DL_DIR}/kernel-vulns"
+# Seconds before the checkout is refreshed. Same shape as
+# cve-update-nvd2-native's CVE_DB_UPDATE_INTERVAL: 0 forces an update, a
+# negative value skips it. A day matches the CNA's publication cadence.
+AVOCADO_CVE_KERNEL_VULNS_UPDATE_INTERVAL ?= "86400"
+
+python () {
+    # Same guard as cve-update-nvd2-native: without cve-check there is no
+    # kernel scan to improve, so there is nothing for this data to serve.
+    if not bb.data.inherits_class("cve-check", d):
+        raise bb.parse.SkipRecipe("Skip recipe when cve-check class is not loaded.")
+}
+
+python do_fetch() {
+    import subprocess
+    import time
+
+    repo = d.getVar("AVOCADO_CVE_KERNEL_VULNS_DIR")
+    url = d.getVar("AVOCADO_CVE_KERNEL_VULNS_URL")
+
+    try:
+        interval = int(d.getVar("AVOCADO_CVE_KERNEL_VULNS_UPDATE_INTERVAL"))
+    except ValueError:
+        bb.fatal("AVOCADO_CVE_KERNEL_VULNS_UPDATE_INTERVAL is %r; expected a "
+                 "number of seconds."
+                 % d.getVar("AVOCADO_CVE_KERNEL_VULNS_UPDATE_INTERVAL"))
+
+    def run(*args, cwd=None):
+        try:
+            subprocess.run(args, cwd=cwd, check=True, capture_output=True, text=True)
+        except subprocess.CalledProcessError as e:
+            bb.fatal("avocado-cve-kernel-vulns: %s failed (%d):\n%s"
+                     % (" ".join(args), e.returncode,
+                        (e.stderr or "").strip()[-2000:]))
+
+    bb.utils.mkdirhier(os.path.dirname(repo))
+    # Two builds sharing a DL_DIR must not fetch into one checkout at once,
+    # and the report task must not read it mid-fetch.
+    lock = bb.utils.lockfile(repo + ".lock")
+    try:
+        head = os.path.join(repo, ".git", "FETCH_HEAD")
+        if not os.path.isdir(os.path.join(repo, ".git")):
+            bb.note("avocado-cve-kernel-vulns: cloning %s into %s (~500 MB, "
+                    "once)" % (url, repo))
+            # Shallow: only the current records are read, never their history.
+            run("git", "clone", "--depth", "1", url, repo)
+            return
+        if interval < 0:
+            bb.note("avocado-cve-kernel-vulns: update disabled, keeping %s" % repo)
+            return
+        age = time.time() - os.path.getmtime(head if os.path.exists(head) else repo)
+        if interval and age < interval:
+            bb.note("avocado-cve-kernel-vulns: %s is %d h old, not refreshing"
+                    % (repo, age // 3600))
+            return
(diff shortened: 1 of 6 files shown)


meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
```diff
@@ -59,6 +64,129 @@ def _warn_if_shared_cve_dir(d, cve_dir):
...
+    #
+    # The improved copy goes beside the scan rather than over it, so the raw
+    # cve-check output stays readable and the two stay comparable.
+    import glob
+    import json
+    import subprocess
+    import sys
+    import avocado_sbom.report as report
+
+    overlay_dir = os.path.join(d.getVar("CVE_CHECK_DIR"), report.OVERLAY_DIRNAME)
```
jetm, line 148: **The `kernel-improved/` overlay directory, written inside `CVE_CHECK_DIR`, breaks `default_paths`'s machine-scope detection.**

`report.py:533-537` (`default_paths`) detects a machine-scoped layout as exactly this: any subdirectory of `${TMPDIR}/deploy/cve` containing `*_cve.json` files. Once this overlay is armed on a build that has NOT machine-scoped `CVE_CHECK_DIR`, the new `kernel-improved/` directory is misread as a machine. `_warn_if_shared_cve_dir` only warns about that layout - it's a supported configuration, not a misuse - so this isn't a corner case.

Repro against this PR's head, flat `CVE_CHECK_DIR`, standalone reporter with no `--machine` (a supported CLI shape with its own `elif flat:` branch): before the overlay exists it resolves `deploy/cve` correctly; after `do_kernel_cve_improve` writes `kernel-improved/linux-yocto_cve.json`, the same call raises `ValueError: .../deploy/cve holds both the flat cve-check layout and machine-scoped subdirectories (kernel-improved); pass --machine, or --cve-dir to read the flat one deliberately`. `main()` routes that to `parser.error` - a hard stop, exit 2, on a build that worked a moment earlier, with an error that names an internal directory as if it were a machine.

Machine-scoped builds are unaffected (the overlay sits one level deeper there). Either exclude `OVERLAY_DIRNAME` from the `scoped` comprehension in `default_paths`, or write the overlay outside `CVE_CHECK_DIR`.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-kernel-vulns-native.bb
```diff
@@ -0,0 +1,95 @@
...
+        try:
+            subprocess.run(args, cwd=cwd, check=True, capture_output=True, text=True)
+        except subprocess.CalledProcessError as e:
+            bb.fatal("avocado-cve-kernel-vulns: %s failed (%d):\n%s"
+                     % (" ".join(args), e.returncode,
+                        (e.stderr or "").strip()[-2000:]))
+
+    bb.utils.mkdirhier(os.path.dirname(repo))
+    # Two builds sharing a DL_DIR must not fetch into one checkout at once,
+    # and the report task must not read it mid-fetch.
```
jetm, line 64: **The lock comment claims a reader guard that isn't implemented.**

The comment states the lock exists so "two builds sharing a DL_DIR must not fetch into one checkout at once, and the report task must not read it mid-fetch." The lock is taken only inside `do_fetch`. `do_kernel_cve_improve` never acquires it - no `[lockfiles]` flag, no `bb.utils.lockfile` call. Its `[depends]` on `avocado-cve-kernel-vulns-native:do_fetch` orders it only within its own build, which is not the cross-build hazard the comment names. The second half of the claim is false as written.

`do_kernel_cve_improve[lockfiles] += "${AVOCADO_CVE_KERNEL_VULNS_DIR}.lock"` would make it true, and matches how oe-core's `cve-update-nvd2-native` does its own locking (via the flag, not by hand) - this recipe otherwise models that one closely.

Failure scenario: two builds share a DL_DIR (this project's normal setup). Build B's `do_fetch` runs `git reset --hard FETCH_HEAD` over `kernel-vulns` while build A's `do_kernel_cve_improve` script is reading it. The script does `glob(datadir/**/CVE-*.json, recursive=True)` then opens each file: one removed between glob and open raises `FileNotFoundError`, non-zero exit, `bb.fatal` - failing build A visibly. The worse case is quieter: a truncated file hits `if len(cve_info) == 0: logging.error(...); break`, so the script exits 0 having filtered against a partial CNA set, and build A publishes a silently wrong kernel CVE report with no warning at all.

jetm (commented):
Reviewed against the PR head, the pinned oe-core (`18263a00da05`), and the layer's own test suite (140 tests pass). Verified the SPDX document path/subdir naming and `MACHINE_ARCH` dash-mangling against a real deploy tree, confirmed the stderr substrings the improve script greps for actually occur in its INFO output, and wrote a repro for the first blocking finding below (confirmed the exception it describes). Two blocking issues, both anchored.

A handful of advisories not worth blocking on: an unguarded `CVE_CHECK_DIR` lookup that pre-empts the recipe's own better error message when `cve-check` isn't inherited; a no-sources warning that recommends a kernel-yocto-only mechanism (`KERNEL_EXTRA_FEATURES:append`) even for vendor kernels that don't have it; an overstated "replaces rather than adds" comment in `report.py`'s overlay-merge logic; the overlay-write task clears and rewrites `*_cve.json` in a way that two machines could clobber if `CVE_CHECK_DIR` isn't machine-scoped (the SPDX-input side of this hazard is already handled, the overlay-output side isn't); and the hardcoded `recipes/` SPDX subdir will silently degrade under the in-flight wrynose migration, which renames that directory.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
hiagofranco: Confirmed and fixed in 07369b80. `default_paths` now skips `OVERLAY_DIRNAME` when detecting machine subdirectories, so a flat `CVE_CHECK_DIR` stays flat once the overlay exists. Kept the overlay inside `CVE_CHECK_DIR` — moving it out would lose the beside-the-scan property. Test added; it fails without the fix.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-kernel-vulns-native.bb
hiagofranco: Right, the comment described a guard that was not there. Added `do_kernel_cve_improve[lockfiles]` in 07369b80, conditional on `AVOCADO_CVE_KERNEL_VULNS_DIR` like the `[depends]` beside it — the default empty value would otherwise lock a relative `.lock` in the cwd. The comment in the vulns recipe is now true as written.

jetm (approved):
Both fixes check out, verified against the branch's tests. One nit not worth blocking on: kas/feature/cve-check.yml:18 has a bare ENG-2414 in a comment; worth swapping for a self-contained description of the trigger condition since tracker IDs aren't resolvable outside the workspace.

jetm:

hiagofranco:
Force-pushed: signed commit.

jetm (approved):

jetm (approved):

jetm:

hiagofranco:
Fixed, probably will need another rebase once other PRs are merged, let me know.

jetm (approved):

PR #361, opened 2026-09-02
ENG-2281: emit build evidence — versions, patched CVEs, and which patches are ours

hiagofranco:
`do_cve_report` filtered to one cve-check status per run, so the patched half — what our
patches and `CVE_STATUS` declarations already fix — was never emitted. That half is what
the correlation service (ENG-2367) subtracts, and no server can reconstruct it from an
SBOM plus an advisory feed.

## What changed

**Both statuses, one document each.** `AVOCADO_CVE_REPORT_STATUS` now takes a list and
defaults to `"Unpatched Patched"`. The first status keeps the existing filename and each
further one is suffixed, so the frozen contract avocado-cli and Peridio fetch keeps
holding the unpatched half rather than being renamed out from under them.
`obsolete_paths()` removes documents the configured statuses will not write, so switching
in either direction cleans up after itself.

The v1 envelope ENG-2347 froze is unchanged — each document still carries one top-level
`status`, and the new per-issue `evidence` key is an addition `verify.additions()`
accepts.

**A new `evidence` field, and a new bbclass to make it honest.** cve-check marks a CVE
`Patched` by three routes and its JSON distinguishes only one:

| route | `detail` in the JSON |
| -- | -- |
| a `CVE_STATUS` declaration | present |
| an applied patch names the CVE | absent |
| shipped version outside every affected range | absent |

The last two are indistinguishable after the build, and only the middle one is a fix we
carry. `avocado-cve-backports.bbclass` splits them during the build instead:
`get_patched_cves()` is exactly patch-file CVEs plus `CVE_STATUS` ones, so minus the
`CVE_STATUS` varflags what remains is the backports. It rescans on every call, so the
class appends to `do_cve_check` rather than prepending, where a bare `return` would abort
the whole spliced task. Written per recipe beside the cve-check results, joined by
`read_backports()`.

Without the class inherited nothing reports `patch-file` and those issues stay
`cve-check` — a build that says less rather than one that guesses.

## Verified on real trees

qemuarm64 and RPi4, both on scarthgap, `verify --strict` clean and matching
`packages_digest` across each pair. RPi4 with the class inherited:

```
patched evidence: {'cve-check': 17290, 'patch-file': 649, 'cve-status': 33}
patch-file by scope: {'build-only': 286, 'feed': 248, 'base-runtime': 115}
```

649 CVEs fixed by patches we carry, against 17,290 cve-check called patched on version
alone. Before this they were one bucket of 17,939 — a 27x overstatement of our backports
if a consumer subtracted the lot.

Cross-checked against an independent derivation: oe-core builds its SPDX
`security_VexFixedVulnAssessmentRelationship` entries from the same `get_patched_cves()`
call. Of 651 marker pairs vs 365 SPDX pairs, exactly one is in SPDX and not in the
markers (`binutils CVE-2025-1180`, a `CVE_STATUS` declaration the class correctly
excludes); the 287 the other way are `-native`/`-cross` recipes SPDX does not document.

170 unit tests, no Yocto required.

## Known gap, not fixed here

`machine/raspberrypi4.yml` stacks the multi-kernel overlay, and
`avocado-multikernel.bbclass` merges only `rpm/` and `pulp-uploads/`. So the alt
multiconfig's `cve/` and `spdx/` stay in `tmp-raspberrypi-6_6/` and the RPi4 report covers
the 6.12.25 kernel alone — the 6.6.63 kernel's 3,723 unpatched CVEs are invisible to it,
and its SBOM is missing a shipped kernel. That is a bbclass gap rather than a
meta-avocado-sbom one; filed separately.

Closes ENG-2281.

### DIFF
--- a/meta-avocado-sbom/classes/avocado-cve-backports.bbclass
+++ b/meta-avocado-sbom/classes/avocado-cve-backports.bbclass
@@ -1,0 +1,100 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Records which CVEs this recipe's own applied patches fix, next to the
+# cve-check results.
+#
+# cve-check marks a CVE "Patched" by three routes and its output distinguishes
+# only one of them. A CVE_STATUS declaration leaves a "detail" in the JSON, so
+# it is recognisable. The other two are not: an applied patch whose filename or
+# "CVE:" line names the CVE, and the shipped version simply falling outside
+# every affected range in the CVE database. Both arrive with no detail at all.
+#
+# Only the first is a fix we carry, and it is the one fact a correlation
+# service cannot reconstruct - it holds the same SBOM and the same advisory
+# feed, and neither says this recipe backported anything. On a qemuarm64 world
+# build the two indistinguishable routes are ~10k of ~22k patched entries, so
+# without this the backport set ships as a superset the consumer cannot
+# isolate.
+#
+# The split is not derivable after the build, but it is trivial during one.
+# get_patched_cves() is exactly patch-file CVEs plus CVE_STATUS ones, and the
+# CVE_STATUS half is already recognisable, so subtracting it leaves the
+# backports. It rescans on every call, so this is unaffected by check_cves()
+# adding its version-comparison verdicts to cve-check's own copy.
+#
+# This is how oe/spdx30_tasks.py derives "backported-patch" for its VEX output.
+# Read from there rather than joined against it: the SPDX documents are written
+# by a different feature (kas/feature/sbom.yml), and the CVE report should not
+# require SBOM generation to be enabled to say what this build patched.
+
+AVOCADO_CVE_BACKPORTS_FILE ?= "${CVE_CHECK_DIR}/${PN}_backports.json"
+AVOCADO_CVE_BACKPORTS_VERSION = "1"
+
+def avocado_cve_backported_cves(d):
+    """CVEs fixed by a patch this recipe applies, and by nothing else.
+
+    get_patched_cves() scans SRC_URI patches for a CVE id in the filename or a
+    "CVE:" line in the body, then adds every CVE_STATUS varflag that maps to
+    Patched. Discarding the varflags leaves the patch-derived set.
+    """
+    from oe.cve_check import get_patched_cves
+
+    patched = set(get_patched_cves(d))
+    # Not "every CVE_STATUS entry mapping to Patched": a varflag naming a CVE a
+    # patch also fixes should still count as declared, the same precedence
+    # cve-check's own JSON has, where a detail wins over silence.
+    patched.difference_update(d.getVarFlags("CVE_STATUS") or {})
+    return sorted(patched)
+
+python do_cve_check:append() {
+    # Append, not prepend: bitbake splices both into one function, so a bare
+    # return here would return out of do_cve_check itself.
+    import json
+
+    path = d.getVar("AVOCADO_CVE_BACKPORTS_FILE")
+
+    try:
+        cves = avocado_cve_backported_cves(d)
+    except FileNotFoundError:
+        # Same failure get_patched_cves() raises into cve-check's own fatal. A
+        # marker is not worth failing a build cve-check is about to fail
+        # anyway, and a missing one reads as "no backports", not as a lie.
+        bb.warn("avocado-cve-backports: could not scan %s's patches; its "
+                "backported CVEs will be absent from the report"
+                % d.getVar("PN"))
+        cves = []
+
+    if not cves:
+        # A recipe that used to carry a CVE patch and no longer does.
+        # CVE_CHECK_DIR is never pruned, so a stale marker would keep claiming
+        # a backport this build does not contain.
+        bb.utils.remove(path)
+    else:
+        bb.utils.mkdirhier(os.path.dirname(path))
+        tmp_path = path + ".tmp"
+        try:
+            with open(tmp_path, "w") as f:
+                json.dump(
+                    {
+                        "version": d.getVar("AVOCADO_CVE_BACKPORTS_VERSION"),
+                        "name": d.getVar("PN"),
+                        "cves": cves,
+                    },
+                    f,
+                    indent=2,
+                    sort_keys=True,
+                )
+                f.write("\n")
+            os.replace(tmp_path, path)
+        except BaseException:
+            bb.utils.remove(tmp_path)
+            raise
+}
+
+# do_cve_check is nostamp upstream, so the markers are rewritten every build and
+# cannot describe a patch set that has since changed. That is what keeps them
+# fresh - not the vardeps below, which never gates a rerun on a nostamp task.
+# It is declared anyway because the task hash should still name the variables
+# the marker is derived from, which is why cve-check.bbclass carries its own
+# vardeps on this same task.
+do_cve_check[vardeps] += "CVE_STATUS SRC_URI"
--- a/meta-avocado-sbom/lib/avocado_sbom/report.py
+++ b/meta-avocado-sbom/lib/avocado_sbom/report.py
@@ -186,11 +186,60 @@ CVE_SUMMARY_FIELDS = ("summary", "description")
 # match no issue at all and produce an empty, CVE-free-looking report.
 CVE_STATUSES = ("Unpatched", "Patched", "Ignored")
 
+# Kept in step with the recipe's AVOCADO_CVE_REPORT_STATUS: a run prunes the
+# documents its statuses do not cover, so two defaults that disagreed would
+# delete each other's output.
+DEFAULT_STATUSES = ("Unpatched", "Patched")
+
+# The whole vocabulary of the "evidence" field. Mirrored by the schema's enum,
+# which test_verify pins to this tuple - the two drifting apart is how a
+# consumer ends up filtering on a value nothing emits.
+CVE_EVIDENCE = ("cve-status", "cve-check")
+
 _FILTERED_COUNTER = {
     "Unpatched": "unpatched_cves",
     "Patched": "patched_cves",
     "Ignored": "ignored_cves",
 }
+
+def _evidence(issue, status):
+    """Who decided this issue's status, which the fields cve-check supplies do
+    not say outright.
+
+    "cve-status" - a recipe declared it. A "detail" is the CVE_STATUS keyword
+    decode_cve_status() read off the recipe (oe/cve_check.py), so its presence
+    means someone wrote this verdict down; "detail" itself says which keyword.
+    A claim inherited from layer metadata, true of the version rather than of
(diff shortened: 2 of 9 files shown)


meta-avocado-sbom/classes/avocado-cve-backports.bbclass
```diff
@@ -0,0 +1,106 @@
...
+                % d.getVar("PN"))
+        bb.utils.remove(path)
+        return
+
+    if not cves:
+        # A recipe that used to carry a CVE patch and no longer does.
+        # CVE_CHECK_DIR is never pruned, so a stale marker would keep claiming
+        # a backport this build does not contain.
+        bb.utils.remove(path)
+        return
```
jetm: **`return` in `do_cve_check:prepend` aborts the whole task, not just the prepended body (also line 70).**

BitBake splices a `:prepend` body into the SAME generated function as the base task (`bb/build.py`: `text = "def %s(d):\n%s" % (func, d.getVar(func, False))`), so the bare `return` at line 77 (`if not cves`) and line 70 (`except FileNotFoundError`) returns out of `do_cve_check` itself - skipping cve-check.bbclass's own body AND the `do_cve_check:append()` in the sibling `avocado-cve-optout.bbclass`. That sibling class uses `:append`, where a terminal `return` is harmless; the shape was copied into a `:prepend` where it is not.

Any recipe with no CVE-naming patch and no Patched `CVE_STATUS` - the overwhelming majority in a world build - takes the `if not cves` path. `check_cves()`/`cve_write_data_json()` never run, so no `<PN>_cve.json` is written; the optout append never runs either, so no `<PN>_optout.json`. `bitbake world && bitbake avocado-cve-report` then reads a `CVE_CHECK_DIR` holding only the handful of backporting recipes, and under the default `AVOCADO_CVE_REPORT_STRICT ?= "1"` `_publishable()` fails the build with "no cve-check opt-out markers ... Add INHERIT += \"avocado-cve-optout\"" - naming a class that is already inherited.

Fix: guard the write (`if cves: ... else: bb.utils.remove(path)`) with no `return` on any path, or move to `:append`.

Verified two ways: read `exec_func_python` in the local bitbake clone, then reproduced with bitbake's own `DataSmart` - with the prepend present, neither the base body nor the append executed.

meta-avocado-sbom/lib/avocado_sbom/report.py
```diff
@@ -639,9 +778,19 @@ def main():
     )
     parser.add_argument(
         "--status",
-        default="Unpatched",
+        default=["Unpatched"],
```
jetm: **CLI `--status` default deletes the build's two documents.**

The CLI's `--status` default is `["Unpatched"]` (single status -> unsuffixed path) while the recipe's `AVOCADO_CVE_REPORT_STATUS` default is now `"Unpatched Patched"` (suffixed). `obsolete_paths()` (lines 608-622, applied here) unconditionally unlinks every other suffix on the stem, before any validation and before anything is written. The two surfaces disagree on naming, so the debugging tool destroys the build's output.

Reproduced: with `avocado-cve-report-qemuarm64-patched.json` and `-unpatched.json` on disk, running `python3 -m avocado_sbom.report --cve-dir ... --pkgdata-dir ... -o <dir>/avocado-cve-report-qemuarm64.json` (no `--status`, i.e. the exact invocation the README's "the same code can run standalone" section invites) left only `avocado-cve-report-qemuarm64.json`. Both build documents gone, including the Patched half this change exists to produce, with no warning. Worse if the run then hits `return 1` on missing pkgdata: the deletion has already happened and nothing is written.

Fix: make the CLI default match the recipe, and/or restrict pruning to suffixes the run was actually asked about.

jetm (commented):
Reviewed at head `e10279fc` (base confirmed as `origin/scarthgap`'s tip). Verified the central premise this PR relies on - that a patch-derived Patched CVE genuinely carries no `detail` on scarthgap's cve-check - directly against oe-core source. Ran the full test suite (170 tests, all pass) and confirmed both blocking issues below are invisible to it. Both reproduced against synthetic build trees, not just traced.

The first is the significant one: `:prepend`'s bare `return` aborts the whole spliced task function, not just the prepended body, which is BitBake's own function-concatenation semantics rather than a guess - verified by reading `exec_func_python`'s `text = "def %s(d):\n%s"` construction and reproducing the abort with BitBake's `DataSmart`. On a world build this means the overwhelming majority of recipes (no CVE patch, no Patched status) never write their `_cve.json` or `_optout.json` at all, and the strict-mode publish gate then fails citing a class that's already inherited. The second is a straightforward but destructive default mismatch between the CLI and the recipe that deletes prior build output with no warning - reproduced end to end.

Advisories not blocking: an `--output`-is-a-directory guard that's now bypassed in multi-status mode (reproduced, raises an unhandled `IsADirectoryError`); a backport-marker diagnostic that names the wrong directory when `--cve-dir`/`--backports-dir` differ; a stated rationale for using `:prepend` over `:append` that doesn't hold (the value is freshly computed, so ordering doesn't matter, and `:append` would sidestep the blocking issue entirely); a duplicate per-recipe read of the patched-CVE set on a `nostamp` task; a README claim ("every issue carries evidence") contradicted 17 lines later; a kas/README drift where the documented "outside kas" equivalent is missing two of the three newly-required `INHERIT` classes; and a stale test fixture that never exercises the new Patched-document shape through CI.

meta-avocado-sbom/classes/avocado-cve-backports.bbclass
hiagofranco: Fixed in 4a706185: moved to `do_cve_check:append`, and both bare `return`s are gone - the write is an if/else now.

You were right about the rationale too. `get_patched_cves()` builds a fresh set on every call (scarthgap `meta/lib/oe/cve_check.py`), so `check_cves()` mutating cve-check's own copy never reaches this one and the ordering bought nothing. The class comment, the README and the commit message all said otherwise; all three corrected.

meta-avocado-sbom/lib/avocado_sbom/report.py
hiagofranco: Fixed in 1c2e5905: `DEFAULT_STATUSES = ("Unpatched", "Patched")` is now the CLI default, and `test_report.py` pins it against the recipe's `AVOCADO_CVE_REPORT_STATUS` so the two cannot drift again.

The second half of your note is fixed separately in 80d8cac4: the pruning now happens just before the first `write_report()` instead of before the checks, so a run that bails on missing pkgdata leaves the tree as it found it. Reproduced both ways against a synthetic tree.

hiagofranco:
Both blocking issues fixed, and six of the seven advisories. Amended into the commits they belong to rather than added on top; head is now `80d8cac4` (was `e10279fc`). 173 tests pass.

Advisories:

- `--output` is a directory: the guard now runs over every path about to be unlinked, not just the ones written, so the `obsolete_paths()` entry for `--output` itself is covered.
- Backport-marker diagnostic: hoisted `backports_dir`, and the message names it instead of `cve_dir`.
- `:prepend` rationale: wrong, dropped. Covered in the thread above.
- README "every issue carries evidence": reworded to "when something decided its status", and the line 17 below it to "absent only on".
- kas/README drift: the outside-kas block now lists all three `INHERIT` lines.
- Stale fixture: `load_patched()` in `test_verify.py`, with the Patched document going through both `check_report()` and the schema.

Not fixed: the duplicate per-recipe read. `get_patched_cves()` runs twice per recipe by construction - once in cve-check, once here - and the only way around it is patching `cve-check.bbclass` to publish its own set. Not worth it for a scan already bounded by the patch files on disk.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
```diff
@@ -64,9 +80,20 @@ python do_cve_report() {
...
+
+    out_file = d.getVar("AVOCADO_CVE_REPORT_FILE")
+    out_paths = report.status_paths(out_file, statuses)
+    # Every path removed before anything can fail, so a run that dies partway
+    # leaves no stale document from a previous build beside a fresh one - the
+    # pair a consumer would read as one build's evidence. Documents this run
+    # will not write go too - see report.obsolete_paths().
+    for path in (list(out_paths.values())
+                 + report.obsolete_paths(out_file, out_paths)):
+        bb.utils.remove(path)
```
jetm: This loop (and the comment above it) still runs before every `bb.fatal()` check below - `cve_dir` unset, no `*_cve.json` files, no pkgdata. `main()` in `report.py` got the deferred-prune fix moved to just before the first `write_report()` call, but this task, the actual `bitbake avocado-cve-report` path the README documents, didn't get the same change. A run that dies on "No pkgdata found" or "No CVE data" still deletes the previous build's report documents first, which is the bug the earlier reply on this thread said was fixed.

Can you move this prune down to just before the first `write_report()` call here too, the same way `main()` handles it?

jetm (commented):
Both threads' first fixes look solid. The second thread's deferred-pruning fix landed in the standalone `report.py` `main()` but not in `do_cve_report()` itself - see the inline comment for where.

jetm:
Not approving this one yet - my review left an open finding (do_cve_report() still prunes before its bb.fatal() checks, see the inline comment). Flagging you two for visibility since this repo needs your sign-off regardless; will approve once Hiago addresses it.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
hiagofranco: Fixed in e8927429 (amended): `doomed` is built at the top but removed inside the status loop, just before the first `write_report()` - same shape as `main()`. Reproduced against a synthetic tree: a run dying on "No pkgdata found" now leaves both documents in place.

jetm (commented):
Re-reviewed the new push. Confirmed the prune-ordering fix from my earlier comment landed: `do_cve_report()` now defers the `bb.utils.remove()` pass until after every `bb.fatal()` validation check, matching `report.main()`'s already-reviewed pattern line-for-line. No blocking findings.

jetm (approved):

jetm:

hiagofranco:
Superseded by #362.

That PR was branched off this one (`hfranco-eng-2281`) and squash-merged as `7dc5ce49`, carrying all three ENG-2281 commits with it — which is why this one now conflicts: upstream already has the content, in the newer form #362 refactored it into.

Rebasing onto `scarthgap` confirms it. git drops `ENG-2281: remove the documents a run will not write` on its own ("patch contents already upstream"), and the remaining two reduce to a blank line added then removed in `avocado-cve-report.bb`. The rebased tree is identical to `scarthgap` — the diff is empty.

Closing as merged-by-proxy; no work is lost.

PR #362, opened 2026-09-02
ENG-2564: carry the alt multiconfig's CVE and SPDX evidence

hiagofranco:
Raspberry Pi and Jetson build two kernels in two multiconfigs. Both ship in the
same feed, but `do_multikernel_merge` carried only `rpm/` and `pulp-uploads/`,
so only one of them appeared in the CVE report or the SBOM.

Two commits: the merge, then the report shape it needs.

**Merge.** Each alt-mc recipe's cve-check results and SPDX documents go into a
`<mc>/` subdirectory of `CVE_CHECK_DIR` and `DEPLOY_DIR_SPDX`. Both mcs write
the same recipe's result and document under one filename, so no additive copy
keeps both, and a subdirectory is a path no task installs - which is what keeps
`do_create_spdx`'s shared-area guard out of it.

**Report.** A recipe held at a second version in that subdirectory becomes an
`alt_versions` record on the entry rather than being merged into it by CVE id.
Additive at v1: `alt_recipes` and `alt_cves` are siblings, not additions to
`cves` and `packaged_cves`, so a consumer that ignores the key reads counters
describing what it read.

Measured on `build-raspberrypi4-sbom-cve` (scarthgap), running the merge and the
report against the real tree:

```
entry   6.12.25+git  base-runtime  3140 CVEs
alt     6.6.63+git   feed          3723 CVEs
counts  alt_recipes 1, alt_cves 3723, stale_dropped 0
verify --strict: ok, health counters zero
```

The 6.6 kernel is in the feed, not the rootfs manifest, so `feed` is its correct
scope. `counts.packages` and `packages_digest` move on the first build with
this, because the alt kernel's ~1,800 feed packages now appear.

Not verified: a full `bitbake avocado-distro` has not been run, so the merge is
proven standalone against the real trees rather than inside bitbake. Jetson is
name-checked against `jetson-l4t.conf` only - no tree exists here. It needs no
`alt_versions` for its kernel, which is a different recipe, but its shared
helpers (`kernel-devsrc`, `nvidia-kernel-oot`) keep one PN across both mcs and
do.

Known ceiling: opt-out and backport markers stay keyed by recipe name, so a
backport carried by the second kernel alone reports `cve-check` evidence rather
than `patch-file` - understating what we patched. Neither kernel emits one
today.

Stacked on #361; the last three commits are that PR's.


meta-avocado/classes/avocado-multikernel.bbclass
```diff
@@ -46,6 +54,52 @@ do_multikernel_merge() {
...
+        #
+        # What to copy is decided from the alt mc's own output alone, like the
+        # rpm filter below and for the same reason: this task runs before
+        # do_compile, and do_compile is what depends on the rest of the distro,
+        # so the default mc's cve/ and spdx/ are mostly absent here. Comparing
+        # against them would copy a different set depending on when it ran.
+        recipe="${pair#*:}"
+        deploy_dir="${DEPLOY_DIR}"
+
+        cve_rel="${CVE_CHECK_DIR#$deploy_dir/}"
```
jetm: **Prefix strip reads unset shell vars; both merges break (also line 80).**

`cve_rel="${CVE_CHECK_DIR#$deploy_dir/}"` and `spdx_rel="${DEPLOY_DIR_SPDX#$deploy_dir/}"` need `CVE_CHECK_DIR`/`DEPLOY_DIR_SPDX` as *shell* variables, but BitBake's `emit_func` only emits `export`/`unexport`-flagged vars into the run script, and neither is exported anywhere in oe-core or this tree. The `#...` form escapes BitBake's own inline-expansion regex (`bb/data_smart.py:35`, `__expand_var_regexp__`), so it falls through to the shell where the name is unset and both `*_rel` resolve to empty strings.

On any RPi/Jetson multi-kernel build, `src` becomes `tmp-<mc>/deploy//<recipe>_cve.json`, which never exists, so `[ -f ] || continue` skips every file and `${CVE_CHECK_DIR}/<mc>/` is never created - the PR's headline `alt_versions` feature never fires. The SPDX branch fails *open* instead: `spdx_src` is `tmp-<mc>/deploy/`, a directory, so `[ -d ]` passes and `grep -rl` recurses the whole alt deploy tree, landing hits two levels deeper than `docs/multi-kernel.md` documents.

Verified by simulating the run script through BitBake's own `DataSmart`+`emit_func` and executing it: `cve_rel=[]`, guard passes, every `src` misses. Cross-checked against a real `run.do_compile` in an existing build tree (85 `export` lines, zero bare assignments, no `DEPLOY_DIR`).

meta-avocado-sbom/lib/avocado_sbom/report.py
```diff
@@ -301,6 +394,29 @@ def read_cve_data(cve_dir, recipe_versions, stats, status="Unpatched", summary=T
 
             packaged = name in recipe_versions
             existing = recipes.get(name)
+            if existing is not None and is_alt and version != existing["version"]:
```
jetm: **Alt kernel becomes the entry when the primary has no issues at the requested status.**

The alt branch is gated on `existing is not None`, but `if not issues: continue` runs earlier, so no entry exists for a recipe with zero issues at the requested status. The alt file then takes the primary-entry path.

Reproduced: shipped `linux-raspberrypi` 6.12.25 carries only `Patched` records, alt 6.6.63 carries one `Unpatched`. The `Unpatched` document emits `version: 6.6.63`, `scope: feed`, no `alt_versions`, `alt_recipes: 0, alt_cves: 0`. The report names the kernel the image does NOT install as the shipped one, its CVEs land in `cves`/`packaged_cves`, and every contract and health check passes. Also reachable when the default mc's `_cve.json` is truncated or stale-dropped. `AltVersionsTests` always populates issues on both sides, so no existing test covers this.

jetm (commented):
Reviewed the full range this PR sits on top of (5 commits, `66ef4966..721b51a7`), since the payload touches BitBake multiconfig internals shared with the sibling PRs in this batch. Verified the shell-expansion claim against BitBake source (`data_smart.py`'s inline-expansion regex, `emit_func`'s export-only emission) and confirmed it empirically by simulating the run script through BitBake's own `DataSmart`+`emit_func` and executing it, then cross-checked against a real build tree's `run.do_compile`. Reproduced findings 2, 3 and 5 directly. Full test suite passes (191 tests) - none of the six issues below are visible to it. Six blocking issues, all anchored.

The first finding is the significant one: the multiconfig merge class silently no-ops via an unset shell variable, so the CVE side of this PR's whole `alt_versions` feature never activates on a real build, while the SPDX side fails open and copies far more than intended. Findings 2 and 4 compound it - a missing or empty alt-mc directory (which is what finding 1 produces) is treated identically to "nothing built at more than one version," so the failure is invisible in the report's own diagnostics.

Advisories not blocking: a docstring placeholder (`ponytail:` where the README says `Known ceiling:`), a percentage that contradicts a sibling file's own math (45% called "the majority"), the Jetson kas comment claiming alt-mc CVEs land as `alt_versions` when the two kernels are actually different recipes there (so no record is ever created - the PR's own test documents the opposite), a log-vs-counter file-count mismatch on alt-mc builds, an ambiguous `origin` field that can't distinguish default-mc from alt-mc packages, unpruned versioned SPDX documents left behind on an alt-kernel bump, two of three copied per-mc files (`_optout.json`, `_backports.json`) that get written but never read back, and a status-loop/obsolete-path implementation now duplicated between `report.py` and the recipe with an already-observed drift (`main()` never prints the new alt counters).

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
```diff
@@ -21,8 +22,19 @@ deltask do_cve_check
...
+# consumer needs both to answer either. Unpatched is what a user can act upon.
+# Patched is what our own patches and CVE_STATUS declarations already fix -
+# the half no server can reconstruct from an SBOM, because "openssl 3.0.12" and
+# an advisory feed together cannot know this recipe carries a backport. Without
+# it every backport reads as a live finding, and they outnumber the real ones
+# by more than an order of magnitude.
+#
+# More than one status suffixes every output file with its status; a single one
+# keeps AVOCADO_CVE_REPORT_FILE verbatim. See report.status_paths().
+AVOCADO_CVE_REPORT_STATUS ?= "Unpatched Patched"
```
jetm, line 37: **Default status change renames and deletes the old report, with no deprecation window.**

`AVOCADO_CVE_REPORT_STATUS` defaults to `"Unpatched Patched"`, so `status_paths()` (report.py:673) suffixes both documents and no default build ever writes `avocado-cve-report-${MACHINE}.json` again. `obsolete_paths()` additionally *deletes* that historical filename.

The README's own "The frozen contract" section names avocado-cli and Peridio as two out-of-tree codebases building against this output. Their fetch of `${DEPLOY_DIR}/avocado-cve/avocado-cve-report-${MACHINE}.json` gets ENOENT on the first build after this lands - no deprecation window, no note in the README, the recipe, or the frozen-contract table.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
```diff
@@ -48,6 +64,47 @@ AVOCADO_CVE_REPORT_MANIFESTS ?= "${DEPLOY_DIR_IMAGE}/avocado-image-rootfs-*.mani
...
+    cve_dir = d.getVar("CVE_CHECK_DIR") or ""
+    topdir = d.getVar("TOPDIR") or ""
+    machine = d.getVar("MACHINE") or ""
+    pairs = (d.getVar("AVOCADO_MULTIKERNEL_MC_RECIPES") or "").split()
+
+    cve_dirs = []
+    pkgdata_dirs = []
+    for mc in dict.fromkeys(p.partition(":")[0] for p in pairs if ":" in p):
+        path = os.path.join(cve_dir, mc)
+        if not os.path.isdir(path):
```
jetm, line 99: **Missing alt-mc CVE dir skipped with no diagnostic, and this is exactly what lets the multikernel.bbclass bug above ship unnoticed.**

`_alt_mc_sources()` does a bare `if not os.path.isdir(path): continue` when `${CVE_CHECK_DIR}/<mc>` is absent, while the sibling pkgdata branch three lines later `bb.warn`s. `do_cve_report` also has no ordering dependency against `do_multikernel_merge`, which lives on a different recipe (`avocado-distro`, `after do_configure before do_compile`).

A non-empty `AVOCADO_MULTIKERNEL_MC_RECIPES` already proves the directory should exist. When it doesn't - build ordering, or the shell-var bug above - the report ships `alt_recipes: 0, alt_cves: 0`, identical in shape to a single-kernel machine, and the build log states "0 recipe(s) built at more than one version" as fact.

meta-avocado-sbom/lib/avocado_sbom/report.py
```diff
@@ -678,70 +906,115 @@ def main():
...
-            file=sys.stderr,
+    for path in out_paths.values():
+        if os.path.isdir(path):
+            parser.error("--output %s is a directory" % path)
+
+    # Every path removed before any is written: a run that fails partway leaves
+    # no stale document from a previous run beside a fresh one, which is the
+    # pair a consumer would read as one build. Documents this run will not
+    # write go too - see obsolete_paths().
+    for path in list(out_paths.values()) + obsolete_paths(args.output, out_paths):
```
jetm: **`--output` directory guard bypassed in multi-status mode.**

The old code checked `args.output` itself; the new loop checks only `out_paths.values()` (the suffixed names), while `args.output` is still reached unguarded by the `obsolete_paths` unlink.

Reproduced: `report.py --status Unpatched Patched -o <existing-dir>` raises an unhandled `IsADirectoryError` out of `os.unlink`, where it previously exited with a clean `parser.error("--output ... is a directory")`.

meta-avocado-sbom/lib/avocado_sbom/report.py
```diff
@@ -753,19 +1026,22 @@ def main():
...
-        "unknown_status_cves",
     ):
         if stats.get(key):
             print("  %s: %d" % (key, stats[key]), file=sys.stderr)
 
+    if backports_unreadable:
+        print(
+            "  %d backport marker(s) in %s would not parse; the CVEs they name "
+            "report 'cve-check' evidence rather than 'patch-file'"
+            % (backports_unreadable, cve_dir),
```
jetm: **Backport-marker warning names the wrong directory.**

`read_backports(args.backports_dir or cve_dir)` at line 916, but the unreadable-marker message here interpolates `cve_dir`. The recipe path (avocado-cve-report.bb:401) gets this right and uses `backports_dir`.

With `--cve-dir A --backports-dir B` and a corrupt marker in B, the message prints "N backport marker(s) in **A** would not parse", sending the operator to a directory that holds no markers.

meta-avocado/classes/avocado-multikernel.bbclass
hiagofranco: Confirmed and fixed in ccd5b02e. `CVE_CHECK_DIR` and `DEPLOY_DIR_SPDX` are read into shell variables first, and the prefix is stripped off those, so both `*_rel` are the real relative paths.

Checked by running the emitted body against a fake alt-mc tree: the cve result lands in `cve/<mc>/`, the SPDX document in `spdx/3.0.1/<mc>/recipes/`, and the `by-hash` symlink tree stays out.

meta-avocado-sbom/lib/avocado_sbom/report.py
hiagofranco: Confirmed and fixed in 2b189016. `read_cve_data` now records the default multiconfig's version for every recipe it scanned, issues or not, and the alt branch keys on that instead of on an existing entry. With no entry it creates one at the shipped version with an empty `cves`, and the alt hangs off it.

Added `test_the_alt_never_becomes_the_entry_when_the_first_has_no_issues`, which fails on the old code with `6.6.63 != 6.12.25`.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
hiagofranco: Confirmed. Fixed on #361, where `status_paths` lives: the first status now keeps `AVOCADO_CVE_REPORT_FILE` and only further ones are suffixed. With the default `"Unpatched Patched"` the historical filename keeps holding the unpatched half, so avocado-cli and Peridio see no change. `obsolete_paths` drops the `-unpatched` name the earlier code would have written.

meta-avocado-sbom/recipes-avocado/avocado-cve-report/avocado-cve-report.bb
hiagofranco: Fixed the silent skip in 2b189016 - a missing `${CVE_CHECK_DIR}/<mc>` now warns and names the directory and the task that writes it.

Left the ordering dependency out on purpose. `do_cve_report` deliberately depends on no image or distro task (see the `AVOCADO_CVE_REPORT_MANIFESTS` comment); a `[depends]` on `avocado-distro:do_multikernel_merge` would pull the whole distro into `bitbake avocado-cve-report`. The warning is what covers the case now.

One correction: the `"0 recipe(s) built at more than one version"` line is guarded by `if alt_cve_dirs`, so a missing directory printed nothing at all rather than a false zero. Same problem, quieter.

meta-avocado-sbom/lib/avocado_sbom/report.py
hiagofranco: Already fixed on #361 - the directory check runs over `doomed`, which includes `--output`. This branch was stacked on the older commits; rebased onto the current #361.

meta-avocado-sbom/lib/avocado_sbom/report.py
hiagofranco: Already fixed on #361 - the message uses `backports_dir`. Same stale-stack reason as above, now rebased.

jetm (approved):

jetm:

hiagofranco:
Rebased only.

meta-avocado-sbom/lib/avocado_sbom/report.py
```diff
@@ -428,9 +615,17 @@ def build_report(
...
     # so an aggregate cannot drift from what the entries say.
     for name, entry in recipes.items():
-        entry["scope"] = _scope(
-            name, recipe_packages.get(name, ()), installed, declared_boot_chain
-        )
+        # Per version: a multi-kernel machine ships one kernel in the rootfs
+        # and the other in the feed. The fallback covers a version pkgdata
+        # knows no packages under, where both sets are empty anyway.
+        by_version = recipe_version_packages.get(name, {})
+        for record in [entry] + list(entry.get("alt_versions", ())):
```
jetm, line 628: This call scores both the entry version AND every `alt_versions` record with the same `_scope()`, which decides base-runtime vs feed by checking package names against `installed` - and `read_manifests()` stores names unversioned (`fields[0]` only, no version). This PR's own README states the norm this feature is built for: "a name both kernels package - kernel, kernel-dbg, kernel-dev - goes to the first pkgdata directory" and that distinguishing the shipped kernel's CVEs from the feed-only kernel's is "exactly the distinction scope exists to make."

Reproduced against this commit: with both multiconfigs packaging a recipe under the same unversioned name (the README's own documented shape), both the entry version AND the alt version score `base-runtime`:
```
entry version: 6.12.25 scope: base-runtime
alt version: 6.6.63 scope: base-runtime
```
Only 6.12.25 is actually on the device. `test_each_version_is_scoped_on_its_own_packages` passes only because it uses synthetic version-suffixed names (`kernel-6.12.25`/`kernel-6.6.63`) rather than the shared unversioned names the README documents as reality, so it never exercises this collision.

A consumer reading `alt_versions[].scope` to judge whether a feed-only kernel's CVE matters for the running device will see `base-runtime` on it too - over-stating device exposure and defeating the one distinction this feature exists to provide.

jetm (commented):
Re-reviewed against the current head (the branch was rebased since my earlier approval, which is no longer reachable). The #361 prune-ordering fix carried over correctly - `do_cve_report()` now defers its remove-before-fatal-checks the same way `report.main()` does.

One blocking finding on the new alt_versions scoping itself: it doesn't actually discriminate by version, which defeats the feature's stated purpose. See the inline comment.

meta-avocado-sbom/lib/avocado_sbom/report.py
hiagofranco: Confirmed and fixed in 3b79c40c. `read_pkgdata` now also returns which version won each package name - the first pkgdata directory, which is the image's - and a record counts an installed name only if that name was won by its own version. `names` stays the full list, so "packages something at all" and the feed scope are unchanged.

Verified against the real `build-raspberrypi4-sbom-cve` trees, both pkgdata directories (4,853 shared package names) and the alt CVE directory the merge writes. With the shipped manifests nothing moves: 44 scope records, 0 differences, counters and `packages_digest` identical, entry 6.12.25+git `base-runtime` and alt 6.6.63+git `feed`. Adding `kernel` and `kernel-dev` to that manifest is where the two part:

```
old: 6.12.25+git base-runtime | alt 6.6.63+git base-runtime
new: 6.12.25+git base-runtime | alt 6.6.63+git feed
```

One narrowing of the premise: that image installs only version-suffixed kernel package names, so the collision does not fire on raspberrypi4 today. It fires the moment an image pulls `kernel`, `kernel-dbg` or `kernel-dev`, which is a package selection away and not something the report should depend on.

Added `test_a_name_both_kernels_package_scopes_only_the_shipped_one`, which packages both versions under the bare `kernel` name and fails on the old code with `base-runtime != feed`. It also turned up a fixture landmine: `build()` used `kw.setdefault("manifest_paths", [self.manifest(...)])` and `manifest()` writes the file, so the default clobbered any manifest a test passed in. The README's alt-pkgdata paragraph now records that scope leans on the same directory order the package map does.

jetm (approved):
Re-reviewed against the two commits added since the last approval (`98383edf`, `3b79c40c`). Approving — the version-scoping is right and the reasoning is documented where it matters. Three things I checked specifically:

**The per-version scope computation is correct**, including the subtle part. `package_versions[name]` is only recorded for the first pkgdata directory that claims a name, so a shared alias (`kernel`, `kernel-dbg`, `kernel-dev`) resolves to the default multiconfig's version, and the alt record's `mine` set correctly excludes it. That's what makes the feed-only kernel score `feed` rather than `base-runtime`, and it depends on `pkgdata_dirs += alt_pkgdata_dirs` appending last. Both halves are commented; good.

**Ordering is safe.** `default_versions` is populated before any alt path is read, because `paths` is built default-first, and it is set before the issues filter so a recipe with no CVE at this status still records the shipping version. The `existing is None` branch that creates the entry at `default_version` is the case that needs it, and it's handled.

**The counter isolation is self-checking.** `alt_cves` staying out of `cves`/`packaged_cves` is asserted by `_check_derived` rather than only documented, so a consumer that ignores `alt_versions` under-reports by exactly `alt_cves` and cannot drift.

---

**One finding, and it is the reason I am flagging rather than just merging.**

When `AVOCADO_MULTIKERNEL_MC_RECIPES` is set but the merge has not run, `_alt_mc_sources` emits `bb.warn` and returns empty, and `do_cve_report` then succeeds. The resulting report has `alt_recipes: 0`.

That value is indistinguishable from a single-kernel machine's correct `0`. The report carries no field saying "this machine declares two kernels", so nothing in the artifact records that half the evidence is missing — only a warning in a build log, and only for whoever was watching that build.

This is the shape the project has been trying to design out: a security artifact that reports a clean number it cannot substantiate. A downstream consumer reading `alt_recipes: 0` has no way to tell an accurate zero from an unmerged one.

The recipe already fatals for the analogous case one screen up (`No pkgdata found, run a world build first`), and the guard is cheap: `pairs` is empty on every single-kernel machine, so failing hard when it is non-empty and no alt directory resolved cannot affect them. The alternative, if the warn is deliberate, is a counter or a top-level field naming the declared multiconfigs, so the gap is visible in the artifact rather than only in the log.

Not blocking this merge — the behaviour is documented and the failure is loud in the build. Worth a follow-up either way.

Separately, the `ponytail:` note about opt-out and backport markers landing in the alt subdirectory records the ceiling ("neither kernel emits one today") and the fix, but not the condition that should trigger it. Worth stating as "when either kernel gains a backport or opt-out marker", so it fires on its own rather than being rediscovered.

hiagofranco:
Both taken, in `e55639a1`.

**The unsubstantiated zero.** Went with the fatal rather than a new field. `_alt_mc_sources` now fatals in both branches — the missing `${CVE_CHECK_DIR}/<mc>` and the missing `tmp-<mc>/pkgdata/${MACHINE}` — instead of warning and continuing. Your read of the blast radius holds: `AVOCADO_MULTIKERNEL_MC_RECIPES` is empty on every single-kernel machine, so the loop body never executes there and nothing existing changes.

Picked it over the counter/top-level field because the field only makes the gap *legible* — a consumer still has to know to look at it, and `alt_recipes: 0` beside `alt_mcs: 1` is a document that says "I am incomplete" and gets shipped anyway. Refusing to write it is the smaller change and leaves nothing to read wrong.

The ordering dependency stays out for the reason from the earlier pass: `do_cve_report` still `[depends]` on no distro task, so `bitbake avocado-cve-report` pulls nothing new. It just refuses to produce a document on a multi-kernel machine when what `do_multikernel_merge` writes isn't there. The message names the task to run.

Second branch got a better instruction while I was in it — if the merged results are present but the alt tmp tree is gone (cleaned, or that kernel stopped shipping), "delete the merged results" is the other valid resolution and it wasn't stated.

Documented in the README beside the counters, since "both are zero on every single-kernel machine" was the sentence a consumer would have read as covering this: a zero there now means one kernel and never a merge that didn't run.

**The marker ceiling.** Fair — the note recorded the ceiling and the fix but not the trigger. Both the `ponytail:` comment in the bbclass and the README's "Known ceiling" paragraph now say to key by `(name, version)` when either kernel gains a backport or opt-out marker, rather than leaving "neither emits one today" as the whole story.

196 tests pass. No lib change, so nothing in the suite moves — the change is entirely in the recipe's guard.

jetm (commented):
Confirmed the alt_versions scope bug is fixed. `read_pkgdata()` now tracks which version won each package name, and `build_report()` filters `installed` down to `mine` (only names whose winning version matches the record's own version) before calling `_scope()`. Reproduced against `e717bf52`: the new `test_a_name_both_kernels_package_scopes_only_the_shipped_one` test exercises exactly the shared-unversioned-name collision, and ran the full `test_report` suite (81/81 pass). Entry (6.12.25) now scores `base-runtime`, alt (6.6.63) scores `feed` - the correct split.

No blocking findings. 2 advisories withheld: `_scope()`'s third parameter is still named/documented as `installed` even though the only call site now passes the filtered `mine` subset (harmless, just a stale name); and the `avocado-cve-report.bb` warn-and-continue → fatal change for a missing alt-multiconfig source is a real, deliberate behavior change worth being aware of (unreachable on single-kernel machines, but a hard fail for anyone relying on the old graceful skip during partial builds).

jetm (approved):

nicksinas:
CodeQL only runs on a schedule for forks, so the code-scanning rule never gets a result on these. Bypassing and merging as deemed as compliant.

PR #364, opened 2026-09-03
jetson: rename Orin Nano / AGX Orin to match the 2026 feed, and unbreak nvidia-kernel-oot

mobileoverlord:
Three independent commits, landing together because `jetson-orin-nano` needs all three to build end to end.

## 1. Drop the `-devkit` suffix

The 2026 feed already calls these targets `jetson-orin-nano` and `jetson-agx-orin`; only 2024 still carries `-devkit`. That split is why the BSP repos have to name a different target per feed leg, and why their 2026 leg published into `2026/next/target/jetson-orin-nano-devkit-ext` — a target the 2026 feed does not have — leaving the real `-ext` repos **empty**.

Renames machine confs, kas machine files, stone manifests, the SDK build/provision scripts, and the machine list in `scripts/lib/common.sh`.

`MACHINEOVERRIDES` moves too. Those strings collide with real upstream meta-tegra machine names, but nothing here consumes `:jetson-*-devkit` as an override suffix — our confs require `orin-nano.inc` directly rather than the upstream machine conf — and **wrynose already made this exact change**, including the `recomputer-j301x` chain that inherits it. The one comment in `tegra-flashvars_%.bbappend` that genuinely names upstream confs is left alone.

## 2. Bump meta-tegra to `053a4e97` (L4T R36.5.2)

The pin was 17 commits behind upstream OE4T scarthgap; our fork had **no divergent commits**, so it was fast-forwarded and the pin follows. `meta-tegra-community` was already at its upstream head and is unchanged.

The vendored `tegra-flash-helper.sh` was checked for drift, since it shadows meta-tegra's and does not move on a repin — byte-identical at both pins, nothing to reconcile.

## 3. Fix nvidia-kernel-oot against linux-yocto 6.6.147

```
nv-pci.c:219:9: error: too few arguments to function 'pci_resize_resource'
```

The 6.6 stable series backported a fourth `exclude_bars` argument. **This breaks scarthgap today, independent of any renaming** — a control build of unmodified scarthgap fails on the same line, R36.5.2 does not fix it, and no fix exists in meta-tegra scarthgap or master as of `053a4e97`.

`exclude_bars` is a bitmask of BARs the kernel must not release while resizing, so the three-argument behaviour is exactly `exclude_bars == 0` — which is what we pass. `nv_resize_pcie_bars()` wants BAR1 and BAR3 both released, a subset of what a zero mask covers, so on-device behaviour is unchanged.

Probed with a conftest rather than `LINUX_VERSION_CODE`: this arrived as a **stable backport**, present in 6.6.147 and absent from earlier 6.6.y, so a version check would be wrong on exactly the kernels that matter. Follows the existing `NV_VFIO_NOTIFIER_ARGUMENT_COUNT` idiom, so it is upstreamable as-is (`Upstream-Status: Pending`).

## Verification

- All three affected machines (`jetson-orin-nano`, `jetson-agx-orin`, `recomputer-j301x`) parse and dependency-graph clean: 6992 recipes, 0 errors. The upstream `tegra234:orin-nano:` chain stays intact.
- `avocado-complete` for `avocado-jetson-orin-nano`: **17,372 of 17,373 tasks**. The one failure was nvidia-kernel-oot, which commit 3 fixes.
- **Control build** of unmodified scarthgap under the old machine name reproduced that failure identically — the rename contributes no failures of its own.
- Commit 3 verified with a `cleansstate` rebuild: `nv-pci.o` compiles, links, and the 36.5.2 RPMs are produced.

## Follow-up

Once this lands, the 2024 leg of `bsp-jetson-orin-nano` / `bsp-jetson-agx-orin` should move to the new name and their `supported_targets` collapse back to a single entry. Existing 2024 projects pinning the old `default_target` will need updating — the old feed path goes stale rather than redirecting.

### DIFF
--- a/docs/adding-a-machine-target.md
+++ b/docs/adding-a-machine-target.md
@@ -679,7 +679,7 @@ To add a new machine called `acme-widget`:
 | Machine | Layer | Bootloader | Provisioning | Notes |
 |---------|-------|-----------|-------------|-------|
 | `raspberrypi5` | `meta-avocado-raspberrypi` | U-Boot | img, sd, usb | Reference ARM target |
-| `jetson-orin-nano-devkit` | `meta-avocado-nvidia` | U-Boot (CBoot) | tegraflash | NVIDIA Jetson, uses swupdate |
+| `jetson-orin-nano` | `meta-avocado-nvidia` | U-Boot (CBoot) | tegraflash | NVIDIA Jetson, uses swupdate |
 | `intel-x86-64-v2` | `meta-avocado-x86-64` | systemd-boot | img, usb | x86-64-v2 EFI target (SSE4.2, Atom-class) |
 | `intel-x86-64-v3` | `meta-avocado-x86-64` | systemd-boot | img, usb | x86-64-v3 EFI target (AVX2, Core-class) |
 | `intel-x86-64-v4` | `meta-avocado-x86-64` | systemd-boot | img, usb | x86-64-v4 EFI target (AVX-512) |
--- a/kas/feature/multi-kernel-jetson.yml
+++ b/kas/feature/multi-kernel-jetson.yml
@@ -9,7 +9,7 @@ header:
 # the per-kernel packagegroup-avocado-{rootfs,initramfs}-modules-${KERNEL_VERSION}).
 #
 # This overlay is included from every Jetson machine yml
-# (kas/machine/{jetson-orin-nano-devkit,jetson-agx-orin-devkit,
+# (kas/machine/{jetson-orin-nano,jetson-agx-orin,
 # recomputer-j301x,recomputer-j401x,icam-540}.yml). Callers do not compose
 # it onto the kas command line; multi-kernel is the standard build for the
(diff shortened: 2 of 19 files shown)


jetm (commented):
Reviewed the rename and the nvidia-kernel-oot fix. Confirmed no dangling old machine-name references anywhere in the tree (kas, machine confs, stone manifests, SDK hook scripts, docs), the dependent `avocado-recomputer-j301x.conf` machine got updated in step, and the kernel-oot conftest probe direction/ifdef gate/Kbuild wiring are correct for both the linux-yocto and linux-jammy-nvidia-tegra build paths. No blocking findings.

3 advisories withheld (not worth blocking on): the new patch's header format doesn't match sibling patches' git-format-patch convention, a stale machine-name reference in an unrelated historical Wrynose-migration doc, and the commit message's "no fix exists upstream" claim being unverifiable from this tree.

jetm (approved):

PR #366, opened 2026-09-03
ENG-2578: add advanced CodeQL setup that scans fork PRs

jasontwong:
Code scanning's default setup never analyzes `pull_request` events from forks (no repo secrets / write token for untrusted head branches), so a fork PR's head commit never gets a CodeQL result and the `code_scanning` branch ruleset on `scarthgap` blocks the merge forever waiting on results that never come. Found via #358, opened from a contributor fork.

## What this does

- `codeql-analyze.yml`: runs on `pull_request` (fork or same-repo alike), `push`, and a weekly `schedule`. For `pull_request` it only produces a local SARIF artifact (`upload: false`) plus the PR number/head SHA as a separate artifact, since GitHub caps the token to read-only for fork PRs regardless of declared permissions, and `github.event.workflow_run.pull_requests` is empty for fork-originated runs. For `push`/`schedule` it uploads directly, same as default setup did.
- `codeql-upload.yml`: triggered by `workflow_run` watching the analyze workflow, runs with the base repo's real permissions. Downloads the SARIF + PR metadata artifacts and uploads via `codeql-action/upload-sarif`, keyed explicitly to the PR's actual head commit (`sha`/`ref`).
- Languages are `python` + `actions` only — `c-cpp` dropped, since nothing in this repo is actually built in CI, so that coverage was never real.

The ruleset's `code_scanning` rule needs no changes — it just checks for a CodeQL result at the head commit, not which workflow produced it.

## Sequencing

Default setup must stay **on** until this is confirmed working (a repo can't run both at once, and disabling it first leaves a coverage gap). Once a same-repo PR and a real fork PR both get correct CodeQL results through this path, default setup gets turned off separately (Settings → Code security → CodeQL analysis).

Linear: https://linear.app/peridio/issue/ENG-2578/codeql-default-setup-never-scans-fork-prs-permanently-blocking-merges

## Test plan

- [ ] Open this PR (same-repo) and confirm `codeql-analyze.yml` produces SARIF + `codeql-upload.yml` uploads it, keyed to this PR's head commit, satisfying the branch ruleset's `code_scanning` check
- [ ] Open a PR from an actual fork and confirm the same end-to-end path produces a CodeQL result for its head commit
- [ ] Confirm `push` to `scarthgap` (post-merge) uploads directly without needing the second workflow
- [ ] Disable default setup only after both PR paths are confirmed working

### DIFF
--- a/.github/workflows/codeql-analyze.yml
+++ b/.github/workflows/codeql-analyze.yml
@@ -6,17 +6,24 @@ name: CodeQL Analyze
 # and merging stayed BLOCKED forever waiting on results that would never come.
 #
 # On push/schedule (trusted, same-repo) we upload directly. On pull_request
-# (fork or same-repo alike) we only ever produce a local SARIF file and hand
-# it to codeql-upload.yml via artifact, since the analyze/upload split is what
-# lets the actual upload happen from a context with security-events: write --
-# see codeql-upload.yml for why.
+# (fork or same-repo alike) we only ever produce a local SARIF artifact, since
+# GitHub caps the token to read-only for fork PRs regardless of what's
+# requested below -- codeql-upload.yml, running with the base repo's real
+# permissions, does the actual upload. It derives the commit to attribute
+# results to from the workflow_run event itself (GitHub-verified) rather than
+# from anything this job writes, since this job's steps run on the fork
+# author's own branch content and are not a trustworthy source for that.
+#
+# Checked out explicitly at the PR's head commit (not the default merge
+# commit) so the tree CodeQL analyzes is the exact commit codeql-upload.yml
+# attributes results to -- see that file for the ref/sha pairing.
 
 on:
   pull_request:
-    branches: [scarthgap]
+    branches: [scarthgap, wrynose]
     types: [opened, synchronize, reopened]
   push:
-    branches: [scarthgap]
+    branches: [scarthgap, wrynose]
   schedule:
     # Catches drift push/PR triggers structurally can't: newly published
     # CodeQL queries matching code that hasn't changed, and any PR analysis
@@ -30,8 +37,12 @@ permissions:
   security-events: write
 
 concurrency:
-  group: codeql-analyze-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+  group: codeql-analyze-${{ github.event_name }}-${{ github.event.pull_request.number || github.ref }}
+  # Only supersede in-flight PR runs. push and schedule share no group key
+  # with each other or with pull_request runs, so a weekly cron tick can't
+  # cancel a just-landed push's baseline scan (or vice versa) and silently
+  # leave scarthgap's default-branch coverage skipping a commit.
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   analyze:
@@ -43,6 +54,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
 
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v3
@@ -64,22 +77,4 @@ jobs:
         with:
           name: sarif-${{ matrix.language }}
           path: sarif-results
-          retention-days: 1
-
-      # PR number/sha aren't available to the workflow_run watcher below,
-      # since it runs against the default branch's own workflow file rather
-      # than this PR's - save them once so it can find the right commit.
-      - name: Save PR metadata
-        if: github.event_name == 'pull_request' && matrix.language == 'python'
-        run: |
-          mkdir -p pr
-          echo "${{ github.event.pull_request.number }}" > pr/number
-          echo "${{ github.event.pull_request.head.sha }}" > pr/sha
-
-      - name: Upload PR metadata
-        if: github.event_name == 'pull_request' && matrix.language == 'python'
-        uses: actions/upload-artifact@v4
-        with:
-          name: pr-metadata
-          path: pr/
           retention-days: 1
--- a/.github/workflows/codeql-analyze.yml
+++ b/.github/workflows/codeql-analyze.yml
@@ -1,0 +1,85 @@
+name: CodeQL Analyze
+
+# Replaces code scanning's "default setup", which never analyzes pull_request
+# events from forks (no repo secrets/write token for untrusted head branches),
+# so a PR's head commit could never satisfy the code_scanning branch ruleset
+# and merging stayed BLOCKED forever waiting on results that would never come.
+#
+# On push/schedule (trusted, same-repo) we upload directly. On pull_request
+# (fork or same-repo alike) we only ever produce a local SARIF file and hand
+# it to codeql-upload.yml via artifact, since the analyze/upload split is what
+# lets the actual upload happen from a context with security-events: write --
+# see codeql-upload.yml for why.
+
+on:
+  pull_request:
+    branches: [scarthgap]
+    types: [opened, synchronize, reopened]
+  push:
+    branches: [scarthgap]
+  schedule:
+    # Catches drift push/PR triggers structurally can't: newly published
+    # CodeQL queries matching code that hasn't changed, and any PR analysis
+    # that silently failed or never ran.
+    - cron: "0 6 * * 1"
+
+permissions:
+  contents: read
+  # Only takes effect for push/schedule -- GitHub caps the token to read-only
+  # for pull_request events from forks regardless of what's requested here.
+  security-events: write
+
+concurrency:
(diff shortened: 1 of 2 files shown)


jetm:
@jasontwong Thanks for the PR. Please include `wrynose` branch too

.github/workflows/codeql-upload.yml
```diff
@@ -0,0 +1,67 @@
...
+      # This is the step that actually needs the base repo's real
+      # permissions - forks never get security-events: write, which is
+      # why the analyze workflow above only ever produces the SARIF and
+      # never uploads it itself for pull_request events.
+      - name: Upload SARIF
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: sarif-results
+          category: "/language:${{ matrix.language }}"
+          ref: refs/pull/${{ steps.pr.outputs.number }}/merge
```
jetm: **The upload's target commit/PR comes from data the fork PR author fully controls, so a malicious contributor can attribute a fabricated SARIF result to a commit they never scanned.**

`pull_request` runs execute `codeql-analyze.yml` from the PR's own branch, so a fork author controls the "Save PR metadata" step (`codeql-analyze.yml:72-77`) that writes `pr/number` and `pr/sha`, and the SARIF content itself. `codeql-upload.yml` reads those bytes back verbatim (lines 43-47) and passes them straight to `upload-sarif` as `ref`/`sha` (lines 66-67) with no validation, running with real `security-events: write` from the default branch. `github/codeql-action/upload-sarif@v3` honors both inputs verbatim on this trigger path (no fork check applies to `workflow_run`, no server-side ref/sha consistency check). Concretely: a forked PR can post `results: []` attributed to any existing commit SHA - including another open PR's head, or the base branch's current tip - satisfying the `code_scanning` branch ruleset for a commit that was never actually scanned; inverted, it can inject fabricated alerts to block a specific PR.

`github.event.workflow_run.head_sha` is a GitHub-verified (non-forgeable) value already available in this job and should replace the artifact-sourced `sha`. The PR number can be resolved server-side via `GET /repos/{owner}/{repo}/commits/{head_sha}/pulls` instead of trusting `pr/number`.

.github/workflows/codeql-analyze.yml
```diff
@@ -0,0 +1,85 @@
...
+
+jobs:
+  analyze:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [python, actions]
+    steps:
+      - name: Checkout
```
jetm, line 55: **The recorded `pr/sha` (PR head) gets paired downstream with `refs/pull/N/merge`, but this checkout produces the merge commit's tree, not the head's - the SARIF and the SHA it's attributed to describe two different commits.**

`actions/checkout@v4` here takes no `ref:`, so on a `pull_request` event it checks out the merge commit (`GITHUB_SHA`), and that's the tree CodeQL actually analyzed. But `pr/sha` (line 77) is `pull_request.head.sha`, and `codeql-upload.yml:66-67` pairs that head sha with `ref: refs/pull/N/merge`. The accepted pairings are `refs/pull/N/head` + head sha, or `refs/pull/N/merge` + the merge sha (`github.sha`, never saved to the artifact here). A ref/sha mismatch is a documented cause of ingestion failures in codeql-action (upstream issues #796, #2623) and, if it doesn't hard-fail, may just silently misattribute the analysis. Worth an end-to-end test against a real fork PR before merge - the whole point of this change is to make the branch ruleset satisfiable, and this could quietly not do that.

.github/workflows/codeql-upload.yml
```diff
@@ -0,0 +1,67 @@
...
+  security-events: write
+  actions: read
+  contents: read
+
+jobs:
+  upload:
+    runs-on: ubuntu-latest
+    # push/schedule runs of CodeQL Analyze already uploaded directly and
+    # never produced a pr-metadata artifact -- only act on PR-triggered runs.
+    if: |
```
jetm, line 47: **A single failing matrix language blocks the upload of the language that succeeded, reproducing the exact "stuck forever" failure this PR exists to fix.**

`workflow_run.conclusion` is run-level, not per-matrix-leg: if `analyze (python)` fails while `analyze (actions)` succeeds, `conclusion` is `failure`, so this gate skips the upload job entirely - the healthy `sarif-actions` artifact is produced and never consumed, then expires in a day (`retention-days: 1`). Relaxing this gate alone isn't enough either: PR-metadata production is itself gated on `matrix.language == 'python'` succeeding (`codeql-analyze.yml:73`), so if the python leg is the one that fails, the `pr-metadata` artifact never gets created and the upload job would just fail at the download step instead. And this workflow's own matrix (lines 32-34) has no `fail-fast: false` to match the analyze workflow's, so one language's download/upload failure here cancels the other mid-flight.

Net effect: any PR where CodeQL fails for one language (or hits a transient toolchain hiccup) gets zero results and stays blocked on the `code_scanning` ruleset with no fix short of a full clean re-run.

.github/workflows/codeql-analyze.yml
```diff
@@ -0,0 +1,85 @@
...
+    - cron: "0 6 * * 1"
+
+permissions:
+  contents: read
+  # Only takes effect for push/schedule -- GitHub caps the token to read-only
+  # for pull_request events from forks regardless of what's requested here.
+  security-events: write
+
+concurrency:
+  group: codeql-analyze-${{ github.event.pull_request.number || github.ref }}
```
jetm: **`push` and `schedule` share one concurrency group, so a push to `scarthgap` and the weekly cron can silently cancel each other.**

`github.event.pull_request.number || github.ref` resolves to the same string for both event types when the ref is the default branch: a `schedule` run has no `pull_request` context (falls through to `github.ref`), and a `push` to `scarthgap` also sets `github.ref` to `refs/heads/scarthgap` - same group, `cancel-in-progress: true`. If the Monday 06:00 cron fires while a just-landed push's analyze run is still in flight (or vice versa), one of them gets cancelled outright, and a cancelled push run never uploads its direct SARIF - the default-branch code-scanning baseline silently skips that commit with nothing surfacing an error.

Include `github.event_name` in the group key, and consider only cancelling in-progress runs for `pull_request` events.

jetm (commented):
Reviewed the two-workflow CodeQL split. The approach (analyze produces SARIF as an artifact for `pull_request`, a `workflow_run`-triggered job uploads it with the base repo's real permissions) is the right shape for scanning fork PRs, but as written the upload side trusts data the fork PR author controls, and a few of the gating conditions defeat the change's own goal for partial failures. Left four inline comments; details below.

**Blocking:**
- The `ref`/`sha` passed to `upload-sarif` are read back from an artifact whose contents a fork PR author fully controls (via `codeql-analyze.yml`'s own workflow file, which runs from their branch) - this lets a malicious PR attribute a fabricated SARIF result to an arbitrary existing commit or PR.
- Separately, even with trusted values, the `sha` recorded (PR head) doesn't match the `ref` it's paired with (`refs/pull/N/merge`, which points at the merge commit `checkout` actually produced) - worth an end-to-end test against a real fork PR.
- The upload gate (`workflow_run.conclusion == 'success'`) is run-level, so one matrix language failing blocks the upload of the language that succeeded - the same "stuck forever" failure this PR sets out to fix.
- `push` and `schedule` share one concurrency group and can cancel each other's runs.

**Also noted (not blocking, worth a look):**
- `codeql-upload.yml` won't exist on `scarthgap` until this PR merges, so this PR's own analyze run has nothing to consume its artifacts - if the `code_scanning` ruleset is already active, this PR may need an admin bypass to land, and the upload path can't be exercised end-to-end until after merge.
- `upload: ${{ github.event_name != 'pull_request' }}` passes the legacy `false`, which codeql-action maps to `failure-only` rather than `never` - harmless on success, but a failed fork-PR run still attempts a debug upload with the read-only token and adds a confusing secondary failure.
- `$GITHUB_OUTPUT` is populated directly from `cat sha`/`cat number` with no delimiter - fragile in principle, though today it's contained to the two consumed keys and only risks denying the upload, not escalating. Moot if the artifact-trust fix above lands.
- Two comment-accuracy nits in `codeql-analyze.yml`: the `schedule` comment claims it catches "any PR analysis that silently failed or never ran," but `schedule` only ever runs against the default branch and can't do that; and the `security-events: write` comment says it "only takes effect for push/schedule," but it's also genuinely granted on same-repo (non-fork) PRs - harmless today since upload is suppressed for all `pull_request` events regardless, but worth a one-word fix.

jasontwong:
Added — both `pull_request` and `push` triggers now include `wrynose` alongside `scarthgap`.

jetm:
All four fixes check out - verified against the actual workflow YAML, not just the commit message. `head_sha` now comes from the `workflow_run` event instead of the artifact, checkout and the upload's ref/sha pairing agree, one language failing no longer blocks the other's upload, and push/schedule no longer share a concurrency group. Nice fix.

Once this comes out of draft, I'll approve it.

jetm (approved):

PR #368, opened 2026-09-04
test: ENG-2578 fork-PR CodeQL scanning verification

jasontwong:
Temporary no-op test PR from an actual fork, to verify the fork-PR CodeQL path added in #366 (ENG-2578) works end-to-end: CodeQL Analyze produces SARIF without uploading, CodeQL Upload resolves this PR's number from the head SHA via workflow_run and uploads results keyed to the real head commit.

Not meant to be merged — will be closed once verified.

Linear: https://linear.app/peridio/issue/ENG-2578/codeql-default-setup-never-scans-fork-prs-permanently-blocking-merges

PR #369, opened 2026-09-04
ENG-2578: fix PR-number resolution for fork PRs in CodeQL Upload

jasontwong:
Follow-up to #366. Testing the fork-PR path (opened #368 from an actual fork to verify) found that `GET /repos/.../commits/{sha}/pulls` — the server-side PR-number resolution added in #366 in response to review — returns empty for a commit that only exists in a fork. Confirmed via both REST and the GraphQL `associatedPullRequests` equivalent against #368's real head commit.

Effect: the upload job's `if` guards on an empty PR number, so every downstream step (download SARIF, upload SARIF) became a skipped no-op that still reported job success — the fork PR got zero CodeQL results, silently reproducing the exact "stuck forever waiting on results" failure #366 exists to fix.

## Fix

`workflow_run.head_repository.owner.login` and `.head_branch` are the same kind of GitHub-verified, non-attacker-controlled run metadata as `head_sha` (confirmed populated even for forks), and `GET /repos/.../pulls?head=<owner>:<branch>` reliably resolves the PR from them — verified against #368.

Linear: https://linear.app/peridio/issue/ENG-2578/codeql-default-setup-never-scans-fork-prs-permanently-blocking-merges

## Test plan

- [x] Confirmed `commits/{sha}/pulls` and GraphQL `associatedPullRequests` return empty for #368's head commit
- [x] Confirmed `pulls?head=jasontwong:codeql-fork-test` correctly resolves #368
- [ ] Re-run the fork-PR test (push to #368's branch or open a fresh fork PR) against this fix and confirm CodeQL Upload actually uploads results this time

### DIFF
--- a/.github/workflows/codeql-upload.yml
+++ b/.github/workflows/codeql-upload.yml
@@ -64,8 +64,19 @@ jobs:
         id: pr
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # A fork PR author fully controls their fork's owner name and
+          # branch name -- interpolating them directly into this run block
+          # via ${{ }} would splice attacker-chosen text into the script
+          # itself (classic Actions script injection), running with this
+          # job's real security-events: write token. Passed as env vars
+          # instead, so they're only ever consumed as inert data by `gh`,
+          # never re-parsed as shell syntax.
+          HEAD_OWNER: ${{ github.event.workflow_run.head_repository.owner.login }}
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
         run: |
-          number=$(gh api "repos/${{ github.repository }}/pulls?head=${{ github.event.workflow_run.head_repository.owner.login }}:${{ github.event.workflow_run.head_branch }}&state=open" \
+          number=$(gh api "repos/${{ github.repository }}/pulls" \
+            -f "head=${HEAD_OWNER}:${HEAD_BRANCH}" \
+            -f state=open \
             --jq '.[0].number // empty')
           echo "number=$number" >> "$GITHUB_OUTPUT"
 
--- a/.github/workflows/codeql-upload.yml
+++ b/.github/workflows/codeql-upload.yml
@@ -15,8 +15,16 @@ name: CodeQL Upload
 # attribute a fabricated result to any commit it chose (github/codeql-action
 # applies no server-side ref/sha consistency check on this trigger path), so
 # `head_sha` comes from the workflow_run event payload (GitHub-verified, not
-# author-controlled) and the PR number is resolved server-side from it via
-# the API rather than trusted from the artifact.
+# author-controlled).
+#
+# The PR number is resolved the same way, from workflow_run's
+# head_repository/head_branch -- NOT from GET .../commits/{sha}/pulls (an
+# earlier version of this file used that, on review feedback, and it looked
+# right but doesn't actually work: that endpoint, and its GraphQL equivalent,
+# come back empty for a commit that only exists in a fork, confirmed against
+# a real fork PR. head_repository/head_branch are run-level metadata like
+# head_sha -- not attacker-controlled -- and `pulls?head=<owner>:<branch>`
+# reliably finds the PR from them.
 #
 # ref/sha are paired as refs/pull/<n>/head + the head sha, matching that
 # codeql-analyze.yml now checks out and analyzes the head commit explicitly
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-upload.yml
```diff
@@ -52,12 +60,23 @@ jobs:
...
+          # via ${{ }} would splice attacker-chosen text into the script
+          # itself (classic Actions script injection), running with this
+          # job's real security-events: write token. Passed as env vars
+          # instead, so they're only ever consumed as inert data by `gh`,
+          # never re-parsed as shell syntax.
+          HEAD_OWNER: ${{ github.event.workflow_run.head_repository.owner.login }}
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
         run: |
-          number=$(gh api "repos/${{ github.repository }}/commits/${{ github.event.workflow_run.head_sha }}/pulls" \
+          number=$(gh api "repos/${{ github.repository }}/pulls" \
```
jetm: **`gh api` switches to POST as soon as a `-f` is present, so this now POSTs to the create-pull-request endpoint instead of querying it.**

From `gh api --help`: "The default HTTP request method is `GET` normally and `POST` if any parameters were added... To send the parameters as a `GET` query string instead, use `--method GET`." Both `-f` flags here are parameters, so this fires `POST /repos/avocado-linux/meta-avocado/pulls` - the create-PR endpoint - rather than the list query the `--jq` downstream expects.

The job's token is `pull-requests: read`, so that request is rejected. `gh` then exits non-zero, and since the step runs under `bash -e` and this is a bare `number=$(...)` assignment, the failure propagates and the step fails. Net effect: CodeQL Upload errors on every fork PR instead of uploading. Loud this time rather than #368's silent skip, but equally non-functional.

Adding `--method GET` is the whole fix:

```yaml
number=$(gh api --method GET "repos/${{ github.repository }}/pulls" \
  -f "head=${HEAD_OWNER}:${HEAD_BRANCH}" \
  -f state=open \
  --jq '.[0].number // empty')
```

Verified against #368's real head: with `--method GET` the query returns `368`. The `-f` form still handles URL-encoding of the branch name, so nothing else needs to change.

.github/workflows/codeql-upload.yml
```diff
@@ -52,12 +60,23 @@ jobs:
...
+            -f "head=${HEAD_OWNER}:${HEAD_BRANCH}" \
+            -f state=open \
             --jq '.[0].number // empty')
```
jetm: **`.[0]` is not pinned to the commit that was actually analyzed, so this can attribute results to the wrong PR or to a stale head.**

Two reachable cases, both surviving this fix:

- One fork branch can back two open PRs at once - one against `scarthgap`, one against `wrynose`, since the analyze workflow now runs on both. `head=<owner>:<branch>` with `state=open` returns both and `.[0]` picks arbitrarily.
- If the fork pushes again after analysis starts, the PR's head has moved by the time this runs, so `ref: refs/pull/N/head` resolves to the new commit while `sha:` is still the analyzed one. That is the same ref/sha mismatch class #366 set out to fix.

Pinning the selection to the trusted `workflow_run.head_sha` closes both, and keeps that value out of the run block too:

```yaml
env:
  HEAD_SHA: ${{ github.event.workflow_run.head_sha }}
run: |
  number=$(gh api --method GET "repos/${{ github.repository }}/pulls" \
    -f "head=${HEAD_OWNER}:${HEAD_BRANCH}" -f state=open \
    --jq '[.[] | select(.head.sha == env.HEAD_SHA)] | .[0].number // empty')
```

`jq`'s `env` reads the variable directly, so there is no interpolation to worry about. Verified both directions against #368: with the matching sha it returns `368`, and with a deliberately wrong sha it returns empty, so a moved head declines to attribute rather than misattributing.

jetm (commented):
Two blocking findings inline, both on the resolve step.

The injection fix in 85884df2 looks right - moving the fork-controlled values into `env:` and consuming them via `-f` keeps the branch name inert, and `${{ github.repository }}` staying inline is fine since it is not attacker-controlled. The root-cause analysis in the description matches what I found independently on #368: both `commits/{sha}/pulls` and the GraphQL `associatedPullRequests` equivalent come back empty for a fork-only commit, and `head_repository`/`head_branch` are the right run-level metadata to key on instead.

One caution on the CodeQL bot's suggested changeset above: do not apply it as written. Its `python3 -c '...' <<<"" HEAD_PARAM="$head_param"` puts the assignment after the command name, so it is passed to `python3` as a positional argument rather than exported. I ran it and it raises `KeyError: 'HEAD_PARAM'`. The `-f` form already in 85884df2 handles the URL-encoding that suggestion was reaching for, without the python3 dependency.

Worth landing the third test-plan box before merge rather than after: with the POST issue above, a fresh fork-PR run is the only thing that distinguishes "resolves correctly" from "fails the same way #368 did, in a different place".

jetm (approved):
Both findings addressed in e1aae16f. Ran the resolve command as written against #368's real head: `--method GET` returns `368`, and the `select(.head.sha == env.HEAD_SHA)` filter returns empty when handed a mismatched sha, so a moved head declines to attribute rather than guessing.

One sequencing note rather than an objection: the remaining test-plan box cannot be ticked before this merges. `workflow_run` always runs the default branch's copy of the upload workflow, so #369's own green checks exercised the pre-fix code on a same-repo PR and never touched the fork path. That run did confirm the diagnosis from the other side though - `commits/{sha}/pulls` resolved fine for this same-repo commit, where it came back empty for #368's fork commit. Worth re-running #368 straight after merge to close it out for real.

PR #373, opened 2026-09-08
ENG-2592: give the Jetson boot chain names and versions the advisory data uses

hiagofranco:
The Jetson boot chain reported one patched CVE. Not because it is clean, but
because four recipes hand `cve-check` names NVD does not index, so the queries
match nothing and the recipes scan clean.

`edk2-firmware-tegra`, `-rcmboot` and `standalone-mm-optee-tegra` set no
`CVE_PRODUCT` at all, so they fall back to `${BPN}` and the UEFI firmware that
owns early boot reports nothing. `optee-os` and `arm-trusted-firmware` do set
one, and most of what they name does not exist: `op-tee:op-tee_os` matches no
NVD record, and vendor `arm:` indexes none of the three TF-A products meta-tegra
lists under it.

bbappends rather than recipe edits, as with `meta-avocado-nxp`'s `imx-atf`.

## Measured on hiagodk, `build-jetson-orin-nano-devkit`

`bitbake <recipes> -c cve_check` — no image, no kernel, which matters because the
Jetson kernel does not currently build there. meta-tegra reset to the
`kas/vendor/nvidia.yml` pin `d2613418` first; the first round of these numbers was
taken on `053a4e97`, a tree kas does not produce.

| recipe | before | after |
| -- | -- | -- |
| `edk2-firmware-tegra`, `-rcmboot` | 0 issues | **19 Unpatched**, 25 Patched |
| `standalone-mm-optee-tegra` | 0 issues | **19 Unpatched**, 25 Patched |
| `optee-os` | 2, both Patched | **11 Unpatched**, 20 Patched |
| `arm-trusted-firmware` | 1, Patched | **2 Unpatched**, 6 Patched |

## Only edk2 needed `CVE_VERSION` to match

`CVE_PRODUCT` alone took edk2 to 44 issues, which is every record NVD holds for
`tianocore:edk2` and `edk_ii` — the version filter excluded nothing. PV is the
L4T release; NVD versions edk2 as `YYYYMM` stable releases, so it sorts below all
of them and every bounded range matched.

`202408` is the newest `edk2-stable*` tag that is an ancestor of the pinned
`SRCREV_edk2 30419afc`, resolved from the fetched mirror rather than read off the
branch name. The `python ()` guard warns if that SRCREV moves, because a stale low
pin silently *hides* CVEs whose ranges start after it, and NVD carries exact-match
rows for edk2.

edk2 is the only one that has to be pinned this way — its PV shares no scheme with
NVD's. `arm-trusted-firmware` and `optee-os` derive theirs from PV
(`${@d.getVar('PV').split('-l4t')[0]}`), so an L4T bump cannot leave them asserting
a stale version behind a `%` append that matches every future PV.

## `CVE_VERSION` on TF-A and OP-TEE is for the CPEs, not the ranges

`oe.cve_check.get_cpe_ids()` strips `+git` and nothing else, and it feeds both
`spdx30_tasks.py` and `vex.bbclass`. Without this the published SPDX would carry
`cpe:2.3:*:trustedfirmware:trusted_firmware-a:2.8-l4t-r36.5.0:*:...`, which matches
no NVD record — the vendor:product finally right and the CPE still dead. The
`imx-atf` sibling escapes this only because its PV is `2.12+git...`, the one suffix
`get_cpe_ids` trims.

OP-TEE normalizes to `4.2.0` because NVD writes it as X.Y.Z throughout (`3.3.0`,
`4.3.0`, `4.5.0`, `4.10.0`). That is a CPE-consumer concern only: the sole op-tee
row in `nvdcve_2-2.db` using operator `=` is `3.22.0_rc1`, so cve-check's
string-compare branch never sees these values, and `_cmpkey` drops the trailing
zero for the range compares.

Confirmed after `-c create_spdx`:

```
cpe:2.3:*:tianocore:edk2:202408:*:*:*:*:*:*:*
cpe:2.3:*:tianocore:edk_ii:202408:*:*:*:*:*:*:*
cpe:2.3:*:trustedfirmware:trusted_firmware-a:2.8:*:*:*:*:*:*:*
cpe:2.3:*:arm_trusted_firmware_project:arm_trusted_firmware:2.8:*:*:*:*:*:*:*
cpe:2.3:*:trustedfirmware:op-tee:4.2.0:*:*:*:*:*:*:*
cpe:2.3:*:linaro:op-tee:4.2.0:*:*:*:*:*:*:*
```

## Deliberately not covered

`tos-optee` reports zero and that is correct — it is a `nopackages` recipe
assembling an image out of `optee-os` and `arm-trusted-firmware` output, building
no source of its own. `optee-os-tadevkit` compiles the same OP-TEE source but
installs only headers and mk fragments under `${includedir}/optee`; no runtime
code ships. Naming either would double-count.

`optee-client` is a real gap and is named in the bbappend comment rather than
fixed: it ships libteec and tee-supplicant on every Jetson, but NVD files no
product for the client alone, so the only available name would attribute
optee_os bugs to it. That needs a decision, not a mechanical fix.

The 19 edk2 findings that remain are NVD rows carrying no version bound at all —
the 2018/2019 Intel UEFI advisories. Nothing in `cve-check` can filter those;
clearing them is `CVE_STATUS` triage.

cve-check also warns three times that it cannot compare `202408` against
`svn_16280` for CVE-2014-8271, an NVD row versioned by SVN revision from edk2's
pre-git era. The except path marks it not-vulnerable, which is the right answer
against a 202408 tree.

Fixes ENG-2592.

### DIFF
--- a/meta-avocado-nvidia/recipes-bsp/arm-trusted-firmware/arm-trusted-firmware_%.bbappend
+++ b/meta-avocado-nvidia/recipes-bsp/arm-trusted-firmware/arm-trusted-firmware_%.bbappend
@@ -1,0 +1,5 @@
+# meta-tegra names vendor arm:, which indexes none of these products, so three of
+# its four pairs match nothing and TF-A reports one CVE instead of its real set.
+# Same two pairs as meta-avocado-nxp's imx-atf append, and vendor-qualified for the
+# same reason: amd: and renesas: ship unrelated forks under these product names.
+CVE_PRODUCT = "trustedfirmware:trusted_firmware-a arm_trusted_firmware_project:arm_trusted_firmware"
--- a/meta-avocado-nvidia/recipes-bsp/arm-trusted-firmware/arm-trusted-firmware_%.bbappend
+++ b/meta-avocado-nvidia/recipes-bsp/arm-trusted-firmware/arm-trusted-firmware_%.bbappend
@@ -4,7 +4,14 @@
 # same reason: amd: and renesas: ship unrelated forks under these product names.
 CVE_PRODUCT = "trustedfirmware:trusted_firmware-a arm_trusted_firmware_project:arm_trusted_firmware"
 
-# cve-check parses 2.8 out of the -l4t-r36.5.2 suffix, but get_cpe_ids() only
+# cve-check parses 2.8 out of the -l4t- suffix on its own, but get_cpe_ids() only
 # strips +git, so the published SPDX and VEX would carry a CPE version no NVD
-# record uses. Same string drives cve-check's exact-match branch.
-CVE_VERSION = "2.8"
+# record uses. Derived rather than pinned because this append matches every future
+# PV: a literal would keep asserting the old version after an L4T bump, and both
+# the CPE and cve-check's exact-match branch read this string verbatim.
+#
+# Not padded the way optee-os is: NVD spells trusted_firmware-a two-component (1.2,
+# 2.0, 2.8), so PV's leading components are already the right shape. The r38/r39
+# meta-tegra branches carry a 2.8.16 PV, which would emit a version no NVD row uses
+# - revisit here when the pin reaches one, rather than guessing the shape now.
+CVE_VERSION = "${@d.getVar('PV').split('-l4t')[0]}"
--- a/meta-avocado-nvidia/recipes-bsp/arm-trusted-firmware/arm-trusted-firmware_%.bbappend
+++ b/meta-avocado-nvidia/recipes-bsp/arm-trusted-firmware/arm-trusted-firmware_%.bbappend
@@ -3,3 +3,8 @@
 # Same two pairs as meta-avocado-nxp's imx-atf append, and vendor-qualified for the
 # same reason: amd: and renesas: ship unrelated forks under these product names.
 CVE_PRODUCT = "trustedfirmware:trusted_firmware-a arm_trusted_firmware_project:arm_trusted_firmware"
+
+# cve-check parses 2.8 out of the -l4t-r36.5.2 suffix, but get_cpe_ids() only
+# strips +git, so the published SPDX and VEX would carry a CPE version no NVD
+# record uses. Same string drives cve-check's exact-match branch.
+CVE_VERSION = "2.8"
--- a/meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
+++ b/meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
@@ -5,3 +5,12 @@
 # The % matches edk2-firmware-tegra and edk2-firmware-tegra-rcmboot, which are the
 # same sources built for two boot paths.
 CVE_PRODUCT = "tianocore:edk2 tianocore:edk_ii"
+
+# PV is the L4T release, which shares no scheme with the YYYYMM edk2-stable
+# versions NVD records, so every bounded range compares true and the recipe
+# reports every edk2 CVE ever filed. edk2-stable202408 is the newest stable tag
+# that is an ancestor of SRCREV_edk2 c80eba3c on NVIDIA's r36.5.1-updates branch.
+#
+# Only edk2 needs this. optee-os 4.2 and arm-trusted-firmware 2.8 carry the same
+# -l4t-r36.5.2 suffix and compare correctly against NVD's semver without help.
+CVE_VERSION = "202408"
--- a/meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
+++ b/meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
@@ -1,27 +1,3 @@
-# NVIDIA's UEFI builds edk2, edk2-platforms and edk2-nvidia from source, but sets
-# no CVE_PRODUCT, so cve-check falls back to ${BPN} and the firmware that owns the
-# whole early boot scans clean on every Jetson. NVD files edk2 under two spellings.
-#
-# The % matches edk2-firmware-tegra and edk2-firmware-tegra-rcmboot, which are the
-# same sources built for two boot paths.
-CVE_PRODUCT = "tianocore:edk2 tianocore:edk_ii"
-
-# PV is the L4T release, which shares no scheme with the YYYYMM edk2-stable
-# versions NVD records, so every bounded range compares true and the recipe
-# reports every edk2 CVE ever filed. edk2-stable202408 is the newest stable tag
-# that is an ancestor of SRCREV_edk2 c80eba3c on NVIDIA's r36.5.1-updates branch.
-#
-# Only edk2 needs this. optee-os 4.2 and arm-trusted-firmware 2.8 carry the same
-# -l4t-r36.5.2 suffix and compare correctly against NVD's semver without help.
-CVE_VERSION = "202408"
-
-# CVE_VERSION is derived from a SRCREV this append cannot see change: bump the
-# meta-tegra pin past edk2-stable202408 and a stale 202408 silently hides every
-# CVE whose range starts after it. NVD carries exact-match rows for edk2, so the
-# stale direction is not the safe one.
-python () {
-    reviewed = "c80eba3cb8aafaac243f45434f6d3a42a4145ed3"
-    if d.getVar("SRCREV_edk2") != reviewed:
-        bb.warn("edk2 SRCREV moved; re-derive CVE_VERSION (now %s) from the newest "
-                "edk2-stable tag that is an ancestor of it" % d.getVar("CVE_VERSION"))
-}
+# The % matches edk2-firmware-tegra and edk2-firmware-tegra-rcmboot, the same
+# sources built for two boot paths.
+require edk2-cve.inc
--- a/meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
+++ b/meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
@@ -14,3 +14,14 @@ CVE_PRODUCT = "tianocore:edk2 tianocore:edk_ii"
 # Only edk2 needs this. optee-os 4.2 and arm-trusted-firmware 2.8 carry the same
 # -l4t-r36.5.2 suffix and compare correctly against NVD's semver without help.
 CVE_VERSION = "202408"
+
+# CVE_VERSION is derived from a SRCREV this append cannot see change: bump the
+# meta-tegra pin past edk2-stable202408 and a stale 202408 silently hides every
+# CVE whose range starts after it. NVD carries exact-match rows for edk2, so the
+# stale direction is not the safe one.
+python () {
+    reviewed = "c80eba3cb8aafaac243f45434f6d3a42a4145ed3"
+    if d.getVar("SRCREV_edk2") != reviewed:
+        bb.warn("edk2 SRCREV moved; re-derive CVE_VERSION (now %s) from the newest "
+                "edk2-stable tag that is an ancestor of it" % d.getVar("CVE_VERSION"))
+}
--- a/meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
+++ b/meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
@@ -1,0 +1,7 @@
+# NVIDIA's UEFI builds edk2, edk2-platforms and edk2-nvidia from source, but sets
+# no CVE_PRODUCT, so cve-check falls back to ${BPN} and the firmware that owns the
+# whole early boot scans clean on every Jetson. NVD files edk2 under two spellings.
+#
+# The % matches edk2-firmware-tegra and edk2-firmware-tegra-rcmboot, which are the
+# same sources built for two boot paths.
+CVE_PRODUCT = "tianocore:edk2 tianocore:edk_ii"
--- a/meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
+++ b/meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
@@ -4,4 +4,12 @@
 #
 # Not set on tos-optee or standalone-mm-optee-tegra: those assemble images out of
 # this recipe's output and build no OP-TEE source, so naming them would double-count.
+# optee-client is left blind on purpose, not by oversight - it ships libteec and
+# tee-supplicant on every Jetson, but NVD files no product for the client alone, so
+# the only name available would attribute optee_os bugs to it. Needs a decision.
 CVE_PRODUCT = "trustedfirmware:op-tee linaro:op-tee"
+
+# As with arm-trusted-firmware: the -l4t-r36.5.2 suffix reaches the SPDX and VEX
+# CPEs verbatim. NVD versions OP-TEE as X.Y.Z, hence 4.2.0 rather than the recipe's
+# 4.2 - identical to cve-check either way, correct only one way to a CPE consumer.
+CVE_VERSION = "4.2.0"
--- a/meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
+++ b/meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
@@ -2,14 +2,29 @@
 # records; the bulk of OP-TEE lives under trustedfirmware:, and the second pair
 # exists in no NVD record at all. Replaced rather than appended, to drop it.
 #
-# Not set on tos-optee or standalone-mm-optee-tegra: those assemble images out of
-# this recipe's output and build no OP-TEE source, so naming them would double-count.
+# Deliberately not set on the three siblings, for three different reasons:
+#   tos-optee            nopackages, assembles an image out of this recipe's and
+#                        arm-trusted-firmware's output, builds no source at all.
+#   optee-os-tadevkit    compiles this same source but installs only headers and
+#                        mk fragments under ${includedir}/optee; no runtime code
+#                        ships, and naming it would double-count every optee_os
+#                        CVE against the same tree.
+#   standalone-mm-optee-tegra
+#                        builds no OP-TEE source either - but it does compile
+#                        edk2, so it is covered under that lens instead, in
+#                        recipes-bsp/uefi/edk2-cve.inc.
+#
 # optee-client is left blind on purpose, not by oversight - it ships libteec and
 # tee-supplicant on every Jetson, but NVD files no product for the client alone, so
 # the only name available would attribute optee_os bugs to it. Needs a decision.
 CVE_PRODUCT = "trustedfirmware:op-tee linaro:op-tee"
 
-# As with arm-trusted-firmware: the -l4t-r36.5.2 suffix reaches the SPDX and VEX
-# CPEs verbatim. NVD versions OP-TEE as X.Y.Z, hence 4.2.0 rather than the recipe's
-# 4.2 - identical to cve-check either way, correct only one way to a CPE consumer.
-CVE_VERSION = "4.2.0"
+# As with arm-trusted-firmware: the -l4t- suffix reaches the SPDX and VEX CPEs
+# verbatim. NVD versions OP-TEE as X.Y.Z throughout (3.3.0, 4.3.0, 4.5.0, 4.10.0),
+# so the value is padded to exactly three components; oe.cve_check._cmpkey drops a
+# trailing zero again for the range compares, and no NVD op-tee row uses operator
+# '=' at 4.2 or 4.2.0, so cve-check's string-compare branch is unaffected either
+# way. Padded rather than suffixed because other meta-tegra branches already spell
+# PV three-component (scarthgap-l4t-r35.x ships optee-os_3.21.0-l4t-r35.6.4), and a
+# bare .0 would publish 3.21.0.0 there - a string no NVD record uses.
+CVE_VERSION = "${@'.'.join((d.getVar('PV').split('-l4t')[0].split('.') + ['0', '0'])[:3])}"
--- a/meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
+++ b/meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
@@ -1,0 +1,7 @@
+# meta-tegra sets "linaro:op-tee op-tee:op-tee_os". The first is real but holds 2
+# records; the bulk of OP-TEE lives under trustedfirmware:, and the second pair
+# exists in no NVD record at all. Replaced rather than appended, to drop it.
+#
+# Not set on tos-optee or standalone-mm-optee-tegra: those assemble images out of
+# this recipe's output and build no OP-TEE source, so naming them would double-count.
+CVE_PRODUCT = "trustedfirmware:op-tee linaro:op-tee"
--- a/meta-avocado-nvidia/recipes-bsp/uefi/edk2-cve.inc
+++ b/meta-avocado-nvidia/recipes-bsp/uefi/edk2-cve.inc
@@ -1,0 +1,28 @@
+# Shared by every recipe that compiles NVIDIA's edk2 tree: edk2-firmware-tegra,
+# edk2-firmware-tegra-rcmboot and standalone-mm-optee-tegra all require the same
+# edk2-firmware-tegra-<PV>.inc, so all three build the same SRCREV_edk2. Kept in
+# one file because CVE_VERSION and the SRCREV it was derived from must not drift
+# apart, and a per-append copy is exactly how they would.
+#
+# meta-tegra sets no CVE_PRODUCT, so cve-check falls back to ${BPN} and the
+# firmware that owns the whole early boot scans clean on every Jetson. NVD files
+# edk2 under two spellings.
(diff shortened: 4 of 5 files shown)


meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
```diff
@@ -0,0 +1,27 @@
...
+# Only edk2 needs this. optee-os 4.2 and arm-trusted-firmware 2.8 carry the same
+# -l4t-r36.5.2 suffix and compare correctly against NVD's semver without help.
+CVE_VERSION = "202408"
+
+# CVE_VERSION is derived from a SRCREV this append cannot see change: bump the
+# meta-tegra pin past edk2-stable202408 and a stale 202408 silently hides every
+# CVE whose range starts after it. NVD carries exact-match rows for edk2, so the
+# stale direction is not the safe one.
+python () {
+    reviewed = "c80eba3cb8aafaac243f45434f6d3a42a4145ed3"
```
jetm: **edk2 drift guard SHA is not the pinned SRCREV_edk2**

The guard's `reviewed` SHA never equals the `SRCREV_edk2` the pinned meta-tegra actually builds, so the drift warning fires on every parse from the first build.

`kas/vendor/nvidia.yml:9` pins meta-tegra to `d2613418be82dd95733824a0839c185ec0e02f2d`. At that commit `recipes-bsp/uefi/edk2-firmware-core-tegra-36.5.0.inc:7` sets `SRCREV_edk2 = "30419afc4a6ea666434f9275057f579b46a07a5f"`, and nothing in meta-avocado overrides it. This append compares against `c80eba3cb8aafaac243f45434f6d3a42a4145ed3`, which appears in no ref of `avocado-linux/vendor-meta-tegra`.

So a Jetson build today emits `WARNING: edk2 SRCREV moved; ...` twice per parse (the `%` matches both `edk2-firmware-tegra` and `edk2-firmware-tegra-rcmboot`) on a correct configuration. A warning that is always on cannot signal the pin bump it exists to signal, which is the fail-open case lines 18-21 call out as not the safe direction.

meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
```diff
@@ -0,0 +1,27 @@
...
+CVE_PRODUCT = "tianocore:edk2 tianocore:edk_ii"
+
+# PV is the L4T release, which shares no scheme with the YYYYMM edk2-stable
+# versions NVD records, so every bounded range compares true and the recipe
+# reports every edk2 CVE ever filed. edk2-stable202408 is the newest stable tag
+# that is an ancestor of SRCREV_edk2 c80eba3c on NVIDIA's r36.5.1-updates branch.
+#
+# Only edk2 needs this. optee-os 4.2 and arm-trusted-firmware 2.8 carry the same
+# -l4t-r36.5.2 suffix and compare correctly against NVD's semver without help.
+CVE_VERSION = "202408"
```
jetm: **CVE_VERSION 202408 derived from an unpinned SRCREV**

`CVE_VERSION = "202408"` is documented as the newest `edk2-stable` ancestor of `c80eba3c` on branch `r36.5.1-updates`, but the build fetches `30419afc` on `r36.5-updates`, so the value was never derived against the tree that compiles.

A build of `edk2-firmware-tegra` fetches `NVIDIA/edk2` at `30419afc` (branch `r36.5-updates`, per the pinned `edk2-firmware-core-tegra-36.5.0.inc:7`). If `30419afc` descends from a later stable tag than `edk2-stable202408`, cve-check compares against `202408` and every edk2 CVE whose affected range starts after 202408 is reported not-vulnerable, while the shipped SPDX/VEX asserts `cpe:2.3:*:tianocore:edk2:202408:*` for a newer binary.

Worth keeping separate from the drift-guard point above: bumping `reviewed` to `30419afc` silences the warning without re-deriving this date.

meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
```diff
@@ -0,0 +1,27 @@
+# NVIDIA's UEFI builds edk2, edk2-platforms and edk2-nvidia from source, but sets
+# no CVE_PRODUCT, so cve-check falls back to ${BPN} and the firmware that owns the
+# whole early boot scans clean on every Jetson. NVD files edk2 under two spellings.
+#
+# The % matches edk2-firmware-tegra and edk2-firmware-tegra-rcmboot, which are the
```
jetm: **standalone-mm-optee-tegra left on default CVE_PRODUCT**

`standalone-mm-optee-tegra` compiles the same edk2 sources as the two recipes this glob covers, but the `edk2-firmware-tegra%` pattern cannot match a filename starting `standalone-mm-`, so it keeps the unmatchable `${BPN}` default.

At the pin, `recipes-bsp/uefi/standalone-mm-optee-tegra_36.5.0.bb:1` is `require edk2-firmware-tegra-36.5.0.inc`, so it builds the same `SRCREV_edk2` / `edk2-nvidia` tree. It is named in `meta-avocado-nvidia/conf/machine/include/avocado-jetson.inc:40`, so the report labels it `boot-chain`, and its `CVE_PRODUCT` still resolves to `standalone-mm-optee-tegra`, which NVD does not index. The S-EL0 secure partition that owns UEFI variable storage then scans clean on every Jetson while the rcmboot variant of the same sources is covered.

`optee-os_%.bbappend:5-6` dismisses this recipe correctly under the OP-TEE lens (it builds no OP-TEE source); it reads as though it was not reconsidered under the edk2 lens.

meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
```diff
@@ -0,0 +1,15 @@
...
+# this recipe's output and build no OP-TEE source, so naming them would double-count.
+# optee-client is left blind on purpose, not by oversight - it ships libteec and
+# tee-supplicant on every Jetson, but NVD files no product for the client alone, so
+# the only name available would attribute optee_os bugs to it. Needs a decision.
+CVE_PRODUCT = "trustedfirmware:op-tee linaro:op-tee"
+
+# As with arm-trusted-firmware: the -l4t-r36.5.2 suffix reaches the SPDX and VEX
+# CPEs verbatim. NVD versions OP-TEE as X.Y.Z, hence 4.2.0 rather than the recipe's
+# 4.2 - identical to cve-check either way, correct only one way to a CPE consumer.
+CVE_VERSION = "4.2.0"
```
jetm: **CVE_VERSION 4.2.0 defeats cve-check exact-match**

The comment's claim that `4.2.0` and `4.2` are identical to cve-check either way does not hold for the exact-match branch, which is a raw string compare, and the ATF append in this same PR picks the opposite convention.

oe-core scarthgap `cve-check.bbclass:309` sets `pv` from `CVE_VERSION` stripping only `+git`, and line 366 is `if (operator_start == '=' and pv == version_start)`. For an NVD OP-TEE row with `operator_start='='` and `version_start='4.2'`, `"4.2.0" == "4.2"` is False; the else branch tests only `>=` / `>` (both False for `=`) and `operator_end` is empty, so the CVE is reported not-vulnerable. `arm-trusted-firmware_%.bbappend:10` chose bare `"2.8"` and does match.

Querying the build's `nvdcve_2-2.db` for the version strings on the `linaro:op-tee` / `trustedfirmware:op-tee` rows would settle which convention is right. The range comparisons really are equal either way, since `oe.cve_check._cmpkey` drops trailing zeros, so only the `=` rows decide it.

meta-avocado-nvidia/recipes-bsp/arm-trusted-firmware/arm-trusted-firmware_%.bbappend
```diff
@@ -0,0 +1,10 @@
+# meta-tegra names vendor arm:, which indexes none of these products, so three of
+# its four pairs match nothing and TF-A reports one CVE instead of its real set.
+# Same two pairs as meta-avocado-nxp's imx-atf append, and vendor-qualified for the
+# same reason: amd: and renesas: ship unrelated forks under these product names.
+CVE_PRODUCT = "trustedfirmware:trusted_firmware-a arm_trusted_firmware_project:arm_trusted_firmware"
+
+# cve-check parses 2.8 out of the -l4t-r36.5.2 suffix, but get_cpe_ids() only
+# strips +git, so the published SPDX and VEX would carry a CPE version no NVD
+# record uses. Same string drives cve-check's exact-match branch.
+CVE_VERSION = "2.8"
```
jetm: **CVE_VERSION literal in a version-wildcard bbappend**

ATF and optee-os pin a literal `CVE_VERSION` behind a `%` append that matches every future PV, with neither a derivation nor the drift guard the third file in this PR got for the identical hazard.

When meta-tegra is bumped to an L4T release carrying `arm-trusted-firmware_2.10-l4t-*.bb`, the `%` append still matches, `CVE_VERSION` stays `"2.8"`, and every TF-A CVE whose range starts at 2.9 falls outside it and is reported not-affected; the published SPDX/VEX then asserts `cpe:2.3:*:trustedfirmware:trusted_firmware-a:2.8:*` for a 2.10 binary. Nothing in the log or the report changes, because the recipe still returns a non-empty CVE list so `no_cve_record_recipes` never names it.

Both values are the leading PV component (`2.8-l4t-r36.5.0`, `4.2-l4t-r36.5.0`), so `${@d.getVar('PV').split('-l4t')[0]}` derives ATF exactly; optee-os needs a `.0` normalization on top, or the same guard treatment edk2 got. Same point applies at `optee-os_%.bbappend:15`.

jetm (commented):
Cold review of the three CVE metadata bbappends. Five blocking findings are inline, anchored to the lines they concern: two on the edk2 append's drift guard and its `CVE_VERSION` derivation, one on `standalone-mm-optee-tegra` falling outside the glob, and one each on the `4.2.0` exact-match convention and the literal `CVE_VERSION` behind a `%` append.

Seven further observations came out of the pass that did not clear the bar for a review comment and are withheld here.

The change's core premise checks out: `get_cpe_ids()` strips only `+git`, so the `-l4t-*` suffix would have reached the SPDX and VEX CPEs verbatim. No BitBake build was run; nothing in this diff is reachable without a Jetson build, and no CI workflow in the repo builds one.

meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
hiagofranco: Confirmed, and the root cause was worse than the SHA. The meta-tegra checkout I measured on was at `053a4e97` (36.5.2), ahead of the `d2613418` pin — so the guard was written against a tree kas does not produce, and so were all the numbers in the description.

Guard now compares `30419afc`, the revision the pinned `edk2-firmware-core-tegra-36.5.0.inc:7` actually fetches. Reset meta-tegra to the pin and re-ran: `bitbake-getvar -r edk2-firmware-tegra CVE_VERSION` and the three siblings resolve clean with no `edk2 SRCREV moved` warning anywhere in the parse.

Fixed in d9382df6.

meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
hiagofranco: You're right that it was derived against the wrong tree — same root cause as the guard SHA. The value survives the correction, though.

Checked the ancestry directly in the fetched `NVIDIA/edk2` mirror rather than off the branch name:

```
$ git tag --list 'edk2-stable*' --merged 30419afc | sort -V | tail -1
edk2-stable202408
$ git tag --list 'edk2-stable*' --merged c80eba3c | sort -V | tail -1
edk2-stable202408
$ git rev-list --count 30419afc..c80eba3c
1
```

The two revisions are one commit apart and share a newest stable ancestor, so `202408` is correct for the pinned tree. Only the comment's provenance was wrong; it now names `30419afc` on `r36.5-updates` in `edk2-cve.inc`.

Re-measured on the pinned tree: 19 Unpatched / 25 Patched, unchanged from the description's figure.

meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
hiagofranco: Confirmed, and the description's justification for skipping it was simply wrong: `standalone-mm-optee-tegra` is not a `nopackages` recipe. Only `tos-optee` is. It `require`s the same `edk2-firmware-tegra-<PV>.inc` and compiles a DSC of its own.

You read the cause correctly — it was dismissed under the OP-TEE lens, where "builds no OP-TEE source" is true, and never reconsidered under the edk2 one.

Added `standalone-mm-optee-tegra_%.bbappend`. Rather than duplicate the product, version and guard across two appends — which is exactly how the SHA in the thread above would drift again — all three now `require` a shared `recipes-bsp/uefi/edk2-cve.inc`.

On the pinned tree it goes from 0 issues to **19 Unpatched, 25 Patched**, identical to the two recipes that were already covered, and its SPDX now carries `cpe:2.3:*:tianocore:edk2:202408:*`.

meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
hiagofranco: Your reading of `cve-check.bbclass:366` is right — it is a raw string compare, and my comment claiming the two are "identical to cve-check either way" was imprecise. I ran the query you suggested against the build's `nvdcve_2-2.db`:

```
SELECT VENDOR,PRODUCT,VERSION_START,OPERATOR_START,VERSION_END,OPERATOR_END
FROM PRODUCTS WHERE PRODUCT LIKE 'op%tee%'
```

The only `=` row in the whole table is `trustedfirmware:op-tee 3.22.0_rc1`. Neither `4.2` nor `4.2.0` matches any exact-match row, so that branch never fires for this recipe either way. Every other op-tee row is a bounded range, where `_cmpkey` makes them equal as you note.

What the query does settle is the convention: NVD writes op-tee as X.Y.Z throughout — `3.3.0`, `3.5.0`, `4.3.0`, `4.5.0`, `4.10.0` — so `4.2.0` is the form a CPE consumer needs. Keeping it, with the comment rewritten to say this instead of the claim you caught.

Same check on TF-A, since it is the counter-example you cited: `trustedfirmware:trusted_firmware-a` has no `=` rows at all, and `arm_trusted_firmware_project:arm_trusted_firmware` has only `1.2` and `1.3`. So bare `2.8` is safe there for the same reason, not a conflicting convention.

meta-avocado-nvidia/recipes-bsp/arm-trusted-firmware/arm-trusted-firmware_%.bbappend
hiagofranco: Agreed — a literal behind a `%` append that matches every future PV is the same hazard, and your derivation is the fix. Both are now derived, so neither needs a guard:

```
# arm-trusted-firmware_%.bbappend
CVE_VERSION = "${@d.getVar('PV').split('-l4t')[0]}"        -> 2.8

# optee-os_%.bbappend
CVE_VERSION = "${@d.getVar('PV').split('-l4t')[0]}.0"      -> 4.2.0
```

Verified with `bitbake-getvar -r` on the pinned tree; the `.0` on optee is the X.Y.Z normalization discussed in the other thread.

edk2 keeps the pin-and-guard because it cannot derive: PV there is the L4T release, which shares no scheme with NVD's `YYYYMM`, so the value has to come from the SRCREV's stable-tag ancestry. The guard SHA is corrected to the pinned `30419afc` in the first thread.

meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
```diff
@@ -0,0 +1,27 @@
...
+# tee-supplicant on every Jetson, but NVD files no product for the client alone, so
+# the only name available would attribute optee_os bugs to it. Needs a decision.
+CVE_PRODUCT = "trustedfirmware:op-tee linaro:op-tee"
+
+# As with arm-trusted-firmware: the -l4t- suffix reaches the SPDX and VEX CPEs
+# verbatim. NVD versions OP-TEE as X.Y.Z throughout (3.3.0, 4.3.0, 4.5.0, 4.10.0),
+# so the trailing .0 is what a CPE consumer needs; oe.cve_check._cmpkey drops it
+# again for the range compares, and no NVD op-tee row uses operator '=' at 4.2 or
+# 4.2.0, so cve-check's string-compare branch is unaffected either way.
+CVE_VERSION = "${@d.getVar('PV').split('-l4t')[0]}.0"
```
jetm: **optee CVE_VERSION appends .0 to an already X.Y.Z PV**

The derivation splices a literal `.0` onto whatever PV yields, which is only right while NVIDIA spells optee-os two-component. meta-tegra's own `scarthgap-l4t-r35.x` branch ships `optee-os_3.21.0-l4t-r35.6.4.bb`, and kirkstone, mickledore, nanbield and langdale carry three-component PVs too. Against any of those this append produces `3.21.0.0`.

The failure is quiet. `_cmpkey` drops the trailing zero, so the range compares still land and cve-check output looks unchanged; nothing warns. But the SPDX and VEX CPE published to customers carries `3.21.0.0`, a version string no NVD record uses - the exact failure the comment above it says the derivation exists to prevent.

Padding to exactly three components keeps `4.2` at `4.2.0` and leaves `3.21.0` alone:

```
${@'.'.join((d.getVar('PV').split('-l4t')[0].split('.') + ['0','0'])[:3])}
```

Worth a line in the comment about the opposite direction on ATF, too. NVD spells `trusted_firmware-a` two-component - the live rows are `1.2`, `2.0`, `2.8` - so if NVIDIA moves to the `2.8.16` PV its r38/r39 branches already carry, that derivation emits `2.8.16` and the `<=2.8` row stops matching. Still better than a stale literal, so not raising it separately, but the three-component question cuts both ways.

jetm (commented):
This covers it. The guard SHA, the derivation and the standalone-mm gap all check out against the pinned tree. One issue with the `.0`, inline.

Re-reviewed the delta `b437d263..c3c26b37` against the pinned vendor tree rather than against the commit message, and all five earlier findings close:

- The drift guard now compares against `30419afc`, which is what `edk2-firmware-core-tegra-36.5.0.inc:7` sets at the pinned meta-tegra `d2613418`.
- `202408` is right, and it was re-derived rather than carried over - that was the trap worth checking, since bumping the guard SHA alone would have silenced the warning without touching the date. Comparing tags on `NVIDIA/edk2`, `edk2-stable202408` is an ancestor of `30419afc` and `edk2-stable202408.01` is not, so it is the newest stable ancestor. The branch name reads `r36.5-updates` now too.
- `standalone-mm-optee-tegra_%.bbappend` does match the recipe under bitbake's own prefix test, the layer's `BBFILES` glob picks it up, and `require edk2-cve.inc` resolves because the including file's directory is prepended to `BBPATH`. Factoring into the `.inc` did not narrow reach for the two original recipes and adds the third.
- On the `4.2.0` question, the argument holds. NVD's op-tee rows are 21 match entries, all ranges on a wildcard version except one exact row at `3.22.0:rc1`, so nothing reaches the `=` branch at either spelling, and NVD does spell op-tee `X.Y.Z` throughout. The asymmetry with ATF is real and now documented.
- Both literals are gone and both derived values are correct at the current pin.

One advisory withheld from a comment: `edk2-basetools-tegra-native_36.5.0.bb` also builds edk2 sources and is matched by neither append. It is pre-existing and native recipes ship nothing, so no action is implied.

No build was run; this was judged by reading the pinned trees.

meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
jetm: Fixed. The guard now compares against `30419afc4a6ea666434f9275057f579b46a07a5f`, which is what `edk2-firmware-core-tegra-36.5.0.inc:7` sets at the pinned meta-tegra `d2613418`. The warning no longer fires on a correct config.

meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
jetm: Fixed, and the trap was avoided - the date was re-derived rather than carried over when the guard SHA moved.

Checked it independently by comparing tags on `NVIDIA/edk2`: `edk2-stable202408` is an ancestor of `30419afc` and `edk2-stable202408.01` is not, so `202408` is the newest stable ancestor of what actually builds. The comment names `r36.5-updates` now too.

meta-avocado-nvidia/recipes-bsp/uefi/edk2-firmware-tegra%.bbappend
jetm: Fixed. `standalone-mm-optee-tegra_%.bbappend` matches the recipe under bitbake's own prefix test, the layer's `BBFILES` glob `recipes-*/*/*.bbappend` picks it up, and the recipe requires `edk2-firmware-tegra-36.5.0.inc`, so `SRCREV_edk2` is set and the guard runs there as well.

Also confirmed the factoring did not narrow reach: `require edk2-cve.inc` resolves because the including file's directory is prepended to `BBPATH`, and the `edk2-firmware-tegra%` append still covers both `edk2-firmware-tegra` and `-rcmboot`.

meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
jetm: Fixed - the `.0` stays and the argument for it holds, which I checked rather than took on trust.

NVD's op-tee rows are 21 match entries, all ranges on a wildcard version except one exact row at `3.22.0:rc1`, so nothing reaches the `=` branch at either `4.2` or `4.2.0`. `_cmpkey` strips the trailing zero for the range compares, and NVD does spell op-tee `X.Y.Z` throughout, so `4.2.0` is the better string for the published CPE. The asymmetry with ATF is real and now documented.

meta-avocado-nvidia/recipes-bsp/arm-trusted-firmware/arm-trusted-firmware_%.bbappend
jetm: Fixed. Both literals are gone - ATF derives `2.8` from PV `2.8-l4t-r36.5.0` and optee-os derives `4.2`, and both are correct at the current pin.

The `.0` splice on the optee side is a separate problem from the literal this thread was about; raised it as its own comment.

meta-avocado-nvidia/recipes-security/optee/optee-os_%.bbappend
hiagofranco: Fixed with your expression. `4.2` -> `4.2.0`, `3.21.0` stays `3.21.0`.

Left ATF unpadded — NVD spells `trusted_firmware-a` two-component, so padding would emit `2.8.0`, which matches nothing. Noted the r38/r39 `2.8.16` direction in the comment there as a revisit-at-the-pin rather than guessing the shape now.

Checked the expansion offline only (bitbake's `__expand_python_regexp__` consumes the whole value, and the eval yields `4.2.0`) — no bitbake run on my side this round.

hiagofranco: Measured now, on the pinned tree (`build-jetson-orin-nano-devkit`, meta-tegra `d2613418`):

```
$ bitbake-getvar -r optee-os CVE_VERSION
CVE_VERSION="4.2.0"
$ bitbake-getvar -r arm-trusted-firmware CVE_VERSION
CVE_VERSION="2.8"
```

Same strings as before the change, so the `-c cve_check` figures in the description stand. The `3.21.0` case stays argued rather than measured — no tree here carries an r35 PV.

jetm: The padding lands. `'.'.join((PV.split('-l4t')[0].split('.') + ['0','0'])[:3])` gives exactly three components for every optee-os PV meta-tegra ships on the branches this layer could pin: 4.2, 4.4 and 4.6 pad to 4.2.0, 4.4.0 and 4.6.0, and 3.21.0 and 3.19.0 pass through unchanged instead of growing a fourth component. I ran both expressions through bitbake's own data_smart rather than a hand-simulated split, so those values are what the parser actually emits.

The ATF point was advisory and the comment answers it: it names the two-component NVD spelling, names 2.8.16 as the shape that would break it, and says to revisit when the pin reaches one. Documenting it beats guessing the shape now - though `master-l4t-r38.x` and `wip-l4t-r39.2.0` also carry `2.12.0` and `2.13.0`, so the comment's `2.8.16` example undersells how much three-component territory is already out there. Worth widening that sentence whenever the revisit happens; not blocking.

Nothing left here.

jetm (commented):
Re-reviewed the delta `c3c26b37..7217354c` against the pinned vendor tree. The optee `.0` splice finding is fixed - verified by expanding the new expression through bitbake's own data_smart against every optee-os PV the fork ships on any branch, not a hand-simulated split, and it produces exactly three components in every case. The ATF advisory point was also picked up with a documenting comment, though the shape it names undersells how much three-component territory already exists on other branches - noted on the thread, not blocking.

Nothing blocking in this delta. Replied on the open thread with the detail; leaving it for the author to resolve.

jetm (approved):

PR #374, opened 2026-09-08
ENG-2198: give i.MX OP-TEE a product name that matches, and a matchable CPE

hiagofranco:
The 2026-09-02 measurement on ENG-2198 concluded that OP-TEE "was never broken"
on i.MX, because meta-arm's `optee-os.inc` already sets `linaro:op-tee
op-tee:op-tee_os` and the recipe surfaces findings today. Verified on
`build-imx95-frdm`, that is wrong in both halves.

- `op-tee:op-tee_os` matches **no NVD record at all** — no such vendor/product pair.
- `linaro:op-tee` holds **2** of the 31 OP-TEE records. The other **29** are
  `trustedfirmware:op-tee`.

So the recipe reports 2 findings, both Patched, 0 Unpatched. That is a worse
failure than the `imx-atf` zero this issue already fixed: a zero invites
suspicion, a 2 reads like the scanner is working. It passed review on exactly
that basis.

Found while fixing the identical defect on Jetson in ENG-2592 (#373), where
meta-tegra's `optee-os` carries the same meta-arm-derived value.

## Measured on hiagodk, `build-imx95-frdm`

`bitbake imx-atf u-boot-imx optee-os -c cve_check` — no image, no rebuild.

```
optee-os   version 4.2.0.imx   products [('op-tee','Yes'), ('op-tee_os','No')]
           2 Patched, 0 Unpatched
```

## `CVE_VERSION` is for the CPE, not the ranges

`oe.cve_check.get_cpe_ids()` strips `+git` and nothing else, and feeds both
`spdx30_tasks.py` and `vex.bbclass`. PV `4.2.0.imx` therefore reaches the
published SPDX and VEX verbatim, naming a version no NVD record uses — the
product name correct and the CPE still dead. `cve-check`'s range comparison
parses the suffix fine, which is what hides it.

`4.2.0` because that is what the suffix means and what NVD uses.

## Also verified in the same run: #358 works

Not part of this diff, but it closes this issue's outstanding build-verification
box, which was the reason it went back to In Progress:

```
imx-atf    version 2.12+git   products [('trusted_firmware-a','Yes'),
                                        ('arm_trusted_firmware','Yes')]
           8 Patched, 0 Unpatched
```

8 records in, zero unpatched — exactly the offline prediction from 2026-09-02,
now observed on a real build with the bbappend applied, and the `imx-atf_%`
wildcard binds correctly.

`u-boot-imx` still reports `u-boot-imx` / `cvesInRecord: No` in that tree, but
the tree's oe-core is `06b1c4f636`, which predates `6ac428e32a` — so it
reproduces the old state and says nothing about whether the current pin delivers
`denx:u-boot`. Still open, and it is a 50-record question.

### DIFF
--- a/meta-avocado-nxp/recipes-security/optee-imx/optee-os_%.bbappend
+++ b/meta-avocado-nxp/recipes-security/optee-imx/optee-os_%.bbappend
@@ -18,7 +18,16 @@ SRC_URI:remove:avocado-imx95-frdm = "file://0007-allow-setting-sysroot-for-clang
 # and reads like a scan that works. Same defect fixed for Jetson in ENG-2592.
 CVE_PRODUCT = "trustedfirmware:op-tee linaro:op-tee"
 
-# PV is 4.2.0.imx, and get_cpe_ids() strips only +git, so the published SPDX and
-# VEX would carry a CPE version no NVD record uses. cve-check's range comparison
-# parses the suffix fine, which is what hides this.
-CVE_VERSION = "4.2.0"
+# get_cpe_ids() strips only +git, so the .imx suffix would reach the published
+# SPDX and VEX as a CPE version no NVD record uses. cve-check's range comparison
+# parses the suffix fine, which is what hides this. Derived rather than a literal
+# because the seven i.MX machines pin different meta-imx releases (4.2.0.imx on
+# FRDM, 4.4.0.imx on EVK/CompuLab/Variscite) and the next vendor bump moves it
+# again.
+CVE_VERSION = "${@d.getVar('PV').split('.imx')[0]}"
+
+# imx95-frdm is repointed at lf-6.18.2_1.0.0 above, which NXP ships as
+# optee-os_4.8.0.imx.bb (vendor-meta-imx 8f1cbb7a21); the recipe selected here is
+# still 4.2.0.imx, so PV describes neither the source built nor the CVE ranges to
+# match against.
+CVE_VERSION:avocado-imx95-frdm = "4.8.0"
--- a/meta-avocado-nxp/recipes-security/optee-imx/optee-os_%.bbappend
+++ b/meta-avocado-nxp/recipes-security/optee-imx/optee-os_%.bbappend
@@ -11,3 +11,14 @@ SRCREV:avocado-imx95-frdm = "e7ed997213779e3d1b7417461c5b4847d3230db9"
 # to lf-6.18.2 (mk/clang.mk was restructured).  The build uses GCC so the
 # patch is not needed.
 SRC_URI:remove:avocado-imx95-frdm = "file://0007-allow-setting-sysroot-for-clang.patch"
+
+# meta-arm's optee-os.inc names "linaro:op-tee op-tee:op-tee_os". The second pair
+# is in no NVD record at all, and linaro:op-tee holds 2 of the 31 that exist -
+# the other 29 are trustedfirmware:op-tee. So this recipe finds 2, both Patched,
+# and reads like a scan that works. Same defect fixed for Jetson in ENG-2592.
+CVE_PRODUCT = "trustedfirmware:op-tee linaro:op-tee"
+
+# PV is 4.2.0.imx, and get_cpe_ids() strips only +git, so the published SPDX and
+# VEX would carry a CPE version no NVD record uses. cve-check's range comparison
+# parses the suffix fine, which is what hides this.
+CVE_VERSION = "4.2.0"


jetm (commented):
Cold review of the i.MX OP-TEE CVE metadata change. The `CVE_PRODUCT` half checks out: the hard `=` correctly overrides `optee-os.inc:11`, the bbappend really does reach the recipe at both PVs in play, and dropping `op-tee:op-tee_os` costs nothing since no NVD record is claimed to carry it.

Two blocking findings are inline, both on `CVE_VERSION`. The short version is that one literal has to cover seven machines resolving to at least two different OP-TEE versions, and the machine whose branch this same file repoints is a third.

Three further observations came out of the pass that did not clear the bar for a review comment and are withheld here.

No build was run; both findings turn on metadata resolution rather than compilation, and each was checked against the pinned vendor trees.

meta-avocado-nxp/recipes-security/optee-imx/optee-os_%.bbappend
```diff
@@ -11,3 +11,14 @@ SRCREV:avocado-imx95-frdm = "e7ed997213779e3d1b7417461c5b4847d3230db9"
...
+# meta-arm's optee-os.inc names "linaro:op-tee op-tee:op-tee_os". The second pair
+# is in no NVD record at all, and linaro:op-tee holds 2 of the 31 that exist -
+# the other 29 are trustedfirmware:op-tee. So this recipe finds 2, both Patched,
+# and reads like a scan that works. Same defect fixed for Jetson in ENG-2592.
+CVE_PRODUCT = "trustedfirmware:op-tee linaro:op-tee"
+
+# PV is 4.2.0.imx, and get_cpe_ids() strips only +git, so the published SPDX and
+# VEX would carry a CPE version no NVD record uses. cve-check's range comparison
+# parses the suffix fine, which is what hides this.
+CVE_VERSION = "4.2.0"
```
jetm: **CVE_VERSION 4.2.0 is wrong on 4 of 7 i.MX machines**

`CVE_VERSION` is set unconditionally in a layer seven i.MX machines share, but only the three FRDM boards resolve `optee-os` to PV 4.2.0.imx. The EVK, CompuLab and Variscite boards pin a later meta-imx and resolve to 4.4.0.imx.

`kas/vendor/nxp-frdm.yml` pins meta-imx at `92ad51ef` (scarthgap-6.6.36-2.1.0), whose `meta-imx-bsp/conf/machine/include/imx-base-extend.inc:103-106` sets `PREFERRED_VERSION_optee-os:mx{6,7,8,9}-nxp-bsp ??= "4.2.0.imx"`. `kas/vendor/nxp-evk.yml` and `kas/vendor/compulab.yml` both pin `a0cf309d` and `kas/vendor/variscite.yml` pins `e1ec1a20`; in those trees the same file sets `4.4.0.imx`, and the recipe present is `optee-os_4.4.0.imx.bb`.

So building `kas/machine/imx8mp-evk.yml`, `imx93-evk.yml`, `ucm-imx8m-plus.yml` or `imx8mp-var-dart.yml` with `kas/feature/cve-check.yml` evaluates NVD ranges against 4.2.0 while 4.4.0 source is on the device: a CVE fixed in 4.3.0 or 4.4.0 comes back Unpatched, and a record whose affected range starts at 4.3.0 is missed entirely. `oe/cve_check.py:236-255` (`get_cpe_ids`) strips only `+git`, and `vex.bbclass:251` plus `spdx30_tasks.py:665,847` feed it `CVE_VERSION` verbatim, so the published VEX and SPDX carry `4.2.0` for a device shipping 4.4.0.

A literal behind a `%` bbappend also keeps matching after the next vendor bump, so this desyncs again silently. `PACKAGE_ARCH = "${MACHINE_ARCH}"` is already set at line 3, so either deriving from PV (`CVE_VERSION = "${@d.getVar('PV').split('.imx')[0]}"`) or per-machine overrides are safe here.

meta-avocado-nxp/recipes-security/optee-imx/optee-os_%.bbappend
```diff
@@ -11,3 +11,14 @@ SRCREV:avocado-imx95-frdm = "e7ed997213779e3d1b7417461c5b4847d3230db9"
...
 # patch is not needed.
 SRC_URI:remove:avocado-imx95-frdm = "file://0007-allow-setting-sysroot-for-clang.patch"
+
+# meta-arm's optee-os.inc names "linaro:op-tee op-tee:op-tee_os". The second pair
+# is in no NVD record at all, and linaro:op-tee holds 2 of the 31 that exist -
+# the other 29 are trustedfirmware:op-tee. So this recipe finds 2, both Patched,
+# and reads like a scan that works. Same defect fixed for Jetson in ENG-2592.
+CVE_PRODUCT = "trustedfirmware:op-tee linaro:op-tee"
+
+# PV is 4.2.0.imx, and get_cpe_ids() strips only +git, so the published SPDX and
```
jetm: **imx95-frdm builds 4.8.0 source, declared 4.2.0**

(Anchored here because the line this concerns, the `SRCBRANCH:avocado-imx95-frdm` at line 7, is outside this PR's hunks. This comment's premise, "PV is 4.2.0.imx", is where the two meet.)

Seventeen lines above the new `CVE_VERSION`, this same file repoints `avocado-imx95-frdm` at the `lf-6.18.2_1.0.0` branch, which NXP pairs with its 4.8.0.imx recipe. The version declared here describes neither the recipe filename's intent nor the code that gets built on that board.

In `avocado-linux/vendor-meta-imx`, commit `8f1cbb7a21` ("optee-os: Switch branch to lf-6.18.2_1.0.0") edits `meta-imx-bsp/dynamic-layers/meta-arm/recipes-security/optee/optee-os_4.8.0.imx.bb`, so `lf-6.18.2_1.0.0` is NXP's 4.8.0 line. Line 7 sets `SRCBRANCH:avocado-imx95-frdm = "lf-6.18.2_1.0.0"` and line 8 an SRCREV on it, while PREFERRED_VERSION on the FRDM meta-imx pin still selects `optee-os_4.2.0.imx.bb`.

Building `kas/machine/imx95-frdm.yml` with `kas/feature/cve-check.yml` therefore scans 4.8.0-era source as 4.2.0: every CVE fixed between 4.2.0 and 4.8.0 is reported Unpatched, every record whose affected range opens above 4.2.0 is missed, and the SPDX/VEX CPE published for the i.MX95 board states a version the image does not contain.

An explicit `CVE_VERSION:avocado-imx95-frdm = "4.8.0"` alongside the machine-general fix above would cover this.

jetm (commented):
This covers both. Nothing blocking.

Re-reviewed the delta `7ee08d6f..7becdd33` against the pinned meta-imx trees rather than against the commit message.

`CVE_VERSION = "${@d.getVar('PV').split('.imx')[0]}"` tracks whichever recipe the machine's pin selects: 4.2.0 on the three FRDM boards, where pin `92ad51ef` ships only `optee-os_4.2.0.imx.bb`, and 4.4.0 on `imx8mp-evk`, `imx93-evk`, `imx8mp-var-dart` and `ucm-imx8m-plus`, where pins `a0cf309d` and `e1ec1a20` ship only `optee-os_4.4.0.imx.bb`. No `PV` override exists in `optee-os-imx.inc`, `optee-os-common-imx.inc`, or meta-arm's `optee.inc` and `optee-os.inc`, so PV is the filename version and the split is exact.

`CVE_VERSION:avocado-imx95-frdm = "4.8.0"` covers the one machine whose PV does not describe what it builds. vendor-meta-imx `8f1cbb7a21` does set `SRCBRANCH = "lf-6.18.2_1.0.0"` in `optee-os_4.8.0.imx.bb`, which is the branch lines 7-8 repoint that machine at, and none of the three meta-imx pins sets a competing `CVE_VERSION`.

Worth recording because it bit the Jetson change in ENG-2592: splitting on the vendor suffix is the right shape here, where appending a component would not have been. `4.2.0.imx`, `4.4.0.imx` and `4.8.0.imx` each yield a three-component version an NVD record uses, and a PV with no `.imx` passes through unchanged, so the expression is a no-op if meta-arm's plain recipe is ever selected.

Two advisories did not clear the bar for a comment and are withheld.

No build was run; this was judged by reading the pinned trees.

meta-avocado-nxp/recipes-security/optee-imx/optee-os_%.bbappend
jetm: Fixed. `CVE_VERSION = "${@d.getVar('PV').split('.imx')[0]}"` now tracks whichever recipe the machine's meta-imx pin selects - 4.2.0 on the three FRDM boards, 4.4.0 on `imx8mp-evk`, `imx93-evk`, `imx8mp-var-dart` and `ucm-imx8m-plus`.

Checked that PV really is the filename version here: no `PV` override exists in `optee-os-imx.inc`, `optee-os-common-imx.inc`, or meta-arm's `optee.inc` and `optee-os.inc`, so the split is exact rather than incidental.

meta-avocado-nxp/recipes-security/optee-imx/optee-os_%.bbappend
jetm: Fixed. `CVE_VERSION:avocado-imx95-frdm = "4.8.0"` covers the one machine whose PV does not describe the source it builds.

Confirmed the pairing rather than assuming it: vendor-meta-imx `8f1cbb7a21` sets `SRCBRANCH = "lf-6.18.2_1.0.0"` in `optee-os_4.8.0.imx.bb`, which is the branch lines 7-8 repoint this machine at, and none of the three meta-imx pins sets a competing `CVE_VERSION`.

jetm (approved):

PR #379, opened 2026-09-11
qcom: RB3 Gen 2 support — apply device tree overlays, flash the var image, share avocado-shm

mobileoverlord:
Adds the Qualcomm Robotics RB3 Gen 2 as a second QCS6490 board alongside the
RUBIK Pi 3, and fixes what stood between it and a working device.

Verified on hardware throughout — every claim below was checked on a board, not
inferred from a build.

## The board

RB3 Gen 2 is Qualcomm's own reference board, so the whole boot chain comes from
the public QCM6490_bootbinaries package rather than a vendor prebuilt repo. It
reuses the QCS6490 support the RubikPi established: `avocado-qcs6490.inc`, the
generic `avocado-build-qcom` / `avocado-provision-qcom` hooks, EL2 + KVM, and
the PREEMPT_RT multiconfig.

## Fixes

**The var partition was never written.** `partition_ufs_rb3gen2.xml` declared it
with `filename=""`, so qdl created the GPT entry and wrote nothing. The board
booted, mounted nothing at `/var` and stopped in emergency mode after a 90 s
wait. `blkid /dev/sda4` returned `PARTLABEL`/`PARTUUID` and no `TYPE=` — zeros.
base-files' generated rule only claims `/dev/disk/by-avocado/var` when
`ID_FS_TYPE` is set, so the name never appeared and fstab waited forever.
rubikpi3 and exmp-q911 both name their image; this one did not.

**Device tree overlays were built and never applied.** The qcom flow flashes one
dtb to `dtb_a` and has no overlay-application step, so the board booted the base
tree and the hardware its overlays describe was absent. On this board that was
its own Ethernet:

```
tc956x_pci-eth 0001:05:00.1: error -ENOENT: Failed to get phy-reset-gpios
tc956x_pci-eth 0001:05:00.1: probe with driver tc956x_pci-eth failed with error -1
```

`phy-reset-gpios` lives in `qcs6490-rb3gen2-staging`, not the base dts.

`linux-avocado-qcom-uki` now merges `AVOCADO_UKI_DTB_OVERLAYS` onto the base dtb
with `fdtoverlay` and passes the result to `ukify --devicetree`. sd-stub installs
it over the firmware's via `EFI_DT_FIXUP_PROTOCOL` — confirmed present in the
stub that ships here, systemd-stub 259.5 with `install_embedded_devicetree` and
`devicetree_fixup`. `avocado-build-qcom`, which rebuilds the UKI around the
pinned kernel, lifts the `.dtb` across with objcopy the way it already lifts
`.cmdline`/`.osrel`.

Gated on `AVOCADO_UKI_DTB_OVERLAYS`, **not** `QCOM_DTB_DEFAULT`: every qcom
machine sets the latter, so keying off it would move rubikpi3 and exmp-q911 off
the firmware's dtb at the same time. Naming overlays is the opt-in; machines
that name none boot exactly as before.

A side effect worth having: the device tree now lives inside the UKI, which
`os_artifacts` already A/Bs as `file:efi:EFI/Linux/avocado-{a,b}+3.efi`, so DT
changes become OTA-updatable with no new manifest machinery.

**avocado-shm was unreachable.** The inter-VM shared-memory driver lived in
meta-avocado-innodisk, so rb3gen2 could not see it and `avocado install` failed
with `No match for argument: kernel-module-avocado-shm`. Ported here unchanged —
nothing in it is board-specific — with `COMPATIBLE_MACHINE` widened to its
consumers.

**Out-of-tree modules were built for one kernel only.** An OOT module recipe is
not in the RT multiconfig's graph just because the RT kernel is; only in-tree
modules come along with the kernel recipe. So the feed carried stock-kernel
builds and the RT runtime refused them:

```
package qps615-dlkm ... requires kernel-module-tc956x-pcie-eth-6.18.37-...,
but none of the providers can be installed ... filtered out by exclude filtering
```

Same shape as `jetson-l4t:nvidia-kernel-oot`. Also records the corollary:
consume these as `kernel-module-<name>`, never the recipe's base package, which
is built once per multiconfig under one name differing only by PR.

**`avocado-var-grow` missed a vardep** on `AVOCADO_VAR_PART_DEV`, which
`do_install` bakes in with sed. sstate handed back a stale package pointing at
the old partition while base-files picked up the new one, and the board stalled
90 s then dropped to initrd emergency. Generic to any machine setting that
variable; kept as its own commit.

## What the RubikPi gets

`avocado-shm` wiring (it had the identical gap — `vmm` names the module
unconditionally), the `avocado-var-grow` fix, and the boot-path work in the
existing commits. It does **not** opt into DTB embedding; that stays per-machine
and untested there.

## Results

RT kernel boots, `/var` on btrfs, `eth0` is the QPS615 rather than USB
networking, `wlan-ce@4cd000` in the live tree confirms `kodiak-el2` applied,
extensions merge in dependency order. The inter-VM shared-memory demo runs
stably:

| | |
|---|---|
| rtt p50 | 625 ns (identical across four 20000-iteration cycles) |
| rtt p99 | 677 ns |
| throughput | 7045–7312 MiB/s |
| mismatches | 0 |

p50/p99 repeating exactly is the timer floor, not suspicious precision: the arch
counter is 19.2 MHz, so a tick is 52 ns and 625 ns is 12 ticks.

## Known gaps

- **Cameras do not work.** The Vision Mezzanine's IMX577 needs
  `qcs6490-rb3gen2-vision-mezzanine`, deliberately not in the overlay list. A
  mezzanine is removable and varies per product, so its tree wants sd-stub's
  `.dtbauto` (several trees, picked by compatible) rather than one UKI per
  mezzanine. Mezzanine support lives in separate extensions, not in this PR.
- `cpus/rb3gen2` in the reference project is still the RubikPi's pinning,
  inherited on the grounds that both are QCS6490. The numbers above are with
  that layout; alternative pinnings are unexplored, as are `irqaffinity` and
  `cpuidle.off`.
- `efivars` is empty because PREEMPT_RT sets `CONFIG_EFI_DISABLE_RUNTIME=y`, so
  `bootctl` reports "Not booted with EFI". Harmless to the stub, which runs
  before ExitBootServices, but userspace cannot introspect loader variables
  without `efi=runtime` — which reintroduces firmware calls on an RT system, so
  it is left off.


mobileoverlord:
## Update: overlay delivery works end to end, plus a firmware size limit found on the way

Three commits added. The mezzanine overlay now reaches the board and the camera comes up.

### The delivery path (`62e2b22`)

An extension declares `device_tree_overlays`; the CLI compiles and stages them and calls two hooks by name — `avocado-dtc-overlay` (SDK: cpp + dtc) and `device-tree-overlay-deliver` (target: validate + record apply order). `avocado-build-qcom` merges them onto the tree it lifts from the Yocto UKI. Since the UKI is an `os_artifact`, installing a mezzanine extension and redeploying changes the device tree over the normal A/B OTA path.

Both hook names are a contract with the CLI, so each needs `DEPENDS` as well as `RDEPENDS` — `RDEPENDS` alone leaves a package unbuilt and absent from the feed, and the failure surfaces at `avocado install` as `No match for argument` long after Yocto reports success.

Also corrects a comment in the machine conf that guessed `.dtbauto` was the mechanism. It isn't — `.dtbauto` selects by matching the **root** compatible, and a mezzanine overlay doesn't change it.

### The firmware limit (`002595e`)

Worth reading even if you skip the rest. **Qualcomm's UEFI faults walking a device tree above roughly 207.5 KB**, and the board drops into dload (`05c6:900e`), which `qdl` cannot flash — recovery is a physical EDL power cycle. The fault is in DXE, *before* systemd-boot counts a boot attempt, so A/B rollback never gets a vote.

Bisected on hardware with content held constant:

| tree | result |
|---|---|
| 207461 bytes | boots |
| 207799 bytes | panics |
| same content, 168501 bytes | boots |

The base tree alone is 207433, so this board sat 400 bytes under the line and merging any mezzanine overlay went over.

Fixed by stripping `__symbols__` from the finished tree — 776 entries, ~40 KB. It exists so `fdtoverlay` can resolve `&labels`; every merge is done by UKI assembly time and nothing on the device reads it, since this platform applies no overlays at runtime. Not stripped in `linux-avocado-qcom-uki.bb`, because that tree is what the hook lifts and merges onto.

Plus a build-time size guard (`AVOCADO_UKI_DTB_MAX_BYTES`, default 207000) so this fails the build rather than the board.

### Verified on hardware

Full chain, RB3 Gen 2 Vision Kit:

```
slot           : b
camss          : okay
17-001a driver : imx412
17-001a name   : imx577
v4l-subdev27   : imx577 17-001a
media graph    : entity 419: imx577 17-001a (1 pad, 1 link)
```

The sensor probes on CCI and links into the CAMSS pipeline. A device-tree change was also delivered over OTA, which had not worked on this platform before.

### Needs review attention

- `2219322` repins `avocadoctl` to `jschneck/os-bundle-content-check`, an unmerged branch — same shape as the `jschneck/bls-entry-updates` pin it replaces. Without it a device-tree-only change is built, bundled, shipped, then skipped, because `os_bundle_satisfied` compared only the rootfs and initramfs build ids. **Both pins must drop before this merges.**
- The CLI half is avocado-linux/avocado-cli#255 (overlay staging was never cleared when the declared set went empty, so a core-kit build inherited the previous build's mezzanine overlay).

meta-avocado-qcom/recipes-kernel/avocado-shm/files/avocado-shm.c
```diff
@@ -0,0 +1,128 @@
...
+	of_node_put(np);
+	if (!rmem)
+		return dev_err_probe(&pdev->dev, -EINVAL, "memory-region is not a reserved-memory node\n");
+
+	if (!rmem->size || (rmem->base | rmem->size) & ~PAGE_MASK)
+		return dev_err_probe(&pdev->dev, -EINVAL,
+				     "region %pa+%pa is empty or not page aligned\n",
+				     &rmem->base, &rmem->size);
+
+	shm_base = rmem->base;
```
jetm: **The globals are committed before the check that would reject a second instance.**

Both assignments land above `misc_register`, so a second device tree node with `compatible = "avocado,shm"` overwrites `shm_base`/`shm_size` and only then fails its own probe. The error return leaves the damage in place: `/dev/avocado-shm`, registered by the first device and possibly already open, now maps the second node's region. Both probes log what reads as correct behaviour, and `shm_mmap` bounds-checks against the wrong size, so the mapping it hands out is inside a window nobody asked for.

Moving the two assignments below the `misc_register` failure check closes it. Nothing between them needs the values.

Single-instance is clearly the intent and no in-tree device tree declares two nodes today, so this is latent rather than live. Worth fixing anyway: the driver's contract is handing out exactly one known window, and this is the only path where it can hand out a different one instead.

jetm (commented):
Reviewed against `wrynose` at 60c46db.

Most of the risk here sits in the boot path, so that is where I spent the time: the UKI rebuild around the lockfile-pinned kernel rather than a glob, the overlay merge taking its order from the delivery record instead of a directory listing, the `__symbols__` strip, and the 207000-byte tree ceiling. Those hold up. The size check matters most of the four, because above the limit the failure lands in firmware before systemd-boot counts an attempt, so A/B rollback never gets a chance at it.

One issue in the new kernel module is worth fixing before this merges. Commented inline.

The rest came back clean. Splitting the SoC settings into `avocado-qcs6490.inc` tracks what the two boards genuinely share rather than what happened to sit in one file, consuming `kernel-module-*` instead of the base package is the right answer to the per-multiconfig PR collision described in `qcom-rt.conf`, and the `qps615-dlkm` and `avocado-shm` alt-mc entries match the jetson precedent the comments point at.

jetm (commented):
Superseded - this review's findings have been consolidated into a newer review on this pull request. See that review for the current state.

meta-avocado-qcom/recipes-kernel/avocado-shm/files/avocado-shm.c
mobileoverlord: Fixed in 3b1224d7c3: moved the shm_base/shm_size assignments below the misc_register success check. A second node with compatible = "avocado,shm" now fails at misc_register before touching the globals, so the first device's window is preserved; and an mmap racing the window between register and assignment hits the existing !shm_size guard in shm_mmap (-ENODEV) rather than a wrong region. dev_info still prints the values (it runs after the assignment). Nothing between needed them, as you noted.

nicksinas (approved):

PR #391, opened 2026-09-16
Pin the rootfs PARTUUID at provision time

jetm:
## Problem

Provisioning to NVMe on a board that also carries a bootable SD card writes the NVMe correctly, reports success, and then boots the SD card.

The initrd locates its rootfs with `blkid -t PARTLABEL=APP`. Both disks carry that label, so it takes whichever the kernel enumerated first. On a Jetson Orin Nano devkit the SD controller is up at 0.6s while PCIe enumeration finishes at 2.5s, so the SD wins by 641 ms:

```
11:55:30.606   mmcblk0: p1 ... p16            SD partitions enumerated
11:55:32.017   Starting Detect and mount APP_<slot> partition on /sysroot
11:55:32.570   tegra194-pcie 14160000.pcie: PCIe Gen.3 x4 link up
11:55:32.658   nvme0n1: p1 ... p16            NVMe enumerated, 641 ms too late
11:55:32.702   erofs: (device mmcblk0p1): mounted
```

Nothing in the boot path says which device UEFI actually chose. L4TLauncher sets no `LoaderDevicePartUUID`, and `systemd-gpt-auto-generator` logs `EFI loader partition unknown, not processing /sysroot`.

The root cause sits further back than the timing. `external-flash.xml` declares the rootfs GUID as an `APPUUID` placeholder, and `tegraflash-bsp.bb` was blanking it at deploy time, so every flash left the GUID for the flashing tool to invent and nothing downstream could predict it. The label was the only stable handle anyone had.

## Solution

Name the rootfs rather than rediscovering it. Provisioning generates a PARTUUID per slot, writes it into the layout it is about to flash, and puts the same value on the repacked boot.img command line. The initrd prefers that over the label.

Waiting longer in the initrd would not have worked: once both disks are up the label is ambiguous rather than missing, so there is nothing to wait for.

## Key changes

- `tegraflash-bsp.bb` keeps the `APPUUID`/`APPUUID_b` placeholders instead of blanking them. The QSPI layout still gets them stripped, since it never sees a provision-time pass and would hand the literal placeholder to the flashing tool.
- `stone-provision-tegraflash.sh` generates a PARTUUID per slot, substitutes both into the layout, and appends `avocado.root_partuuid=` and `avocado.root_partuuid_b=` to the kernel command line. Both slots are carried because one boot.img is written to `A_kernel` and `B_kernel` alike.
- `avocado-tegra-init` resolves the rootfs by that PARTUUID, selecting the slot it already reads from `BootChainOsCurrent`.
- Label discovery survives for the one case where it cannot be ambiguous: exactly one disk carrying `APP` means a board whose GPT was rewritten out from under its command line, so it boots that and says so. Two or more refuses rather than guessing.
- The repacked boot.img command line carries the two PARTUUID arguments and nothing else. An earlier revision of this PR read the prebuilt image's existing cmdline and appended to it; that is what hung the board on hardware, and it is dropped - see the reviewer notes.
- `make-sdcard.sh` passes the parsed `unique_guid` to `sgdisk --partition-guid`. It read the GUID out of the layout and then built the sgdisk command without it, so the pinned value never reached the GPT and sgdisk invented a random one. Without this the rest of the change is inert.

The UUIDs are generated per flash rather than fixed per medium. A constant living in the repo would be identical on every board, so removable media carrying it could offer itself as the rootfs, and revising one later would orphan every board already flashed with the old value.

## Reviewer notes

**Verified on a Jetson Orin Nano devkit, with the SD card left installed.** Full output in a comment below. The board boots the NVMe rootfs while both disks still carry a partition named `APP`, and the PARTUUID on the command line is the one the GPT holds:

```
# grep ' / ' /proc/mounts
/dev/nvme0n1p1 / erofs ro,relatime,...

# cat /proc/cmdline
avocado.root_partuuid=49138F00-0C01-4B18-88A8-40E5D8E8DD54 avocado.root_partuuid_b=623AFD90-...

# ls -l /dev/disk/by-partuuid/49138f00-0c01-4b18-88a8-40e5d8e8dd54
... -> ../../nvme0n1p1

# grep PARTNAME /sys/block/mmcblk0/mmcblk0p1/uevent /sys/block/nvme0n1/nvme0n1p1/uevent
/sys/block/mmcblk0/mmcblk0p1/uevent:PARTNAME=APP
/sys/block/nvme0n1/nvme0n1p1/uevent:PARTNAME=APP
```

That run found two defects in the original revision of this PR, both fixed here.

The first is the cmdline carry-forward, which hung the board in userspace with no panic and no further console output. It was written to duplicate #347, a wrynose fix verified on AGX Thor (tegra264) against a tegra264-specific console mismatch; tegra234 does not have that problem and does not survive the fix. The PR body already said the forward-port should drop the hunk - it should have been dropped here too.

The second is `make-sdcard.sh` never passing `--partition-guid`. This one is worth a reviewer's attention because nothing about it is visible from the diff of the other files: the GUID is parsed by `nvflashxmlparse.py`, emitted into the shell line `make-sdcard.sh` evals, assigned to a declared local, and then silently not used. Provisioning succeeded and the initrd searched for a PARTUUID that was never written. I first suspected a timing race and disproved it - a 30s diagnostic showed NVMe fully enumerated before the search started, and 56 seconds of searching still failed.

The stub coverage from before still holds and still runs: the substitution fills both slots with no placeholder surviving and the XML still parses, the guard rejects a layout missing only the slot-A placeholder, five command-line shapes select the right slot, an SD card carrying `PARTLABEL=APP` no longer wins when the pinned NVMe is absent, and `blkid` exiting 2 does not trip `set -e` into an early abort.

**The feedback I most want is on the placeholder contract.** Making `APPUUID` survive `process_flash_xml` means the deployed layout is no longer flashable as-is by any path that skips `stone-provision-tegraflash.sh`. I traced `external-flash.xml.in` to that one producer and that one consumer, but a second opinion on whether some other flow reaches these layouts is worth more than review of the shell itself.

Second, the repack now hands mkbootimg a command line holding only the two PARTUUID arguments, so a machine whose `KERNEL_ARGS` carry something the kernel genuinely needs would lose it at provision time. On this machine that is demonstrably not the case - an empty mkbootimg cmdline reaches a login prompt every time, on both a from-scratch flash and the published reference image - but I have only tested tegra234. If another Tegra machine in this layer depends on its `KERNEL_ARGS` reaching the repacked image, this hunk is wrong for it and I would rather hear that than discover it.

The offending token in the carried-forward `KERNEL_ARGS` was never isolated, only the fact that carrying them hangs. Narrowing that is worth its own investigation and is not blocking here, since the fix is to pass only what this change needs.

**This is stacked on #392** (shellcheck sweep), which is its base. `make-sdcard.sh` is a vendored NVIDIA helper carrying about forty pre-existing shellcheck findings including three `SC2168` errors, so touching one line of it fails the gate; stacking was cheaper than a drive-by cleanup buried in this diff. GitHub will retarget this PR to `scarthgap` when #392 merges.

Related: the OTA writer resolved its target slot by label too, which this change made more visible rather than caused. That was fixed separately in #390, now merged, and the conflict the two had in `avocado-tegra-init` is already resolved here.


jetm:
Related: #393 adds `avocado-set-boot-device`, which writes `BootOrder` so the firmware loads the kernel from the disk that was provisioned.

The two are halves of the same thing. This PR makes the rootfs follow whichever kernel booted; #393 decides which kernel that is. With only this one, a board with both disks provisioned still boots whichever the firmware prefers - it just now boots that disk's kernel and rootfs together instead of mixing them. With only #393, the firmware loads the right kernel and the initrd still resolves `APP` by label across both disks.

jetm:
## Confirmed on hardware: the kernel and the rootfs came from different disks

Read off the bench Orin Nano, with an SD card and an NVMe both carrying a full provisioned layout.

**What the firmware chose** — `BootCurrent` and `BootOrder` from efivarfs, decoded (4-byte attributes, then little-endian `u16` entry numbers):

```
BootCurrent: 06 00 00 00 | 03 00                     -> Boot0003
BootOrder:   07 00 00 00 | 03 00 01 00 02 00 04 00   -> 0003,0001,0002,0004,...
```

**What those entries are** — descriptions pulled out of the UTF-16 payload:

```
Boot0001: UEFI SD Device
Boot0003: UEFI Samsung SSD 960 EVO 250GB S3ESNX0JA13241W 1
```

So the firmware booted **Boot0003, the NVMe**.

**What the running system reports:**

```
# findmnt -no SOURCE /
/dev/mmcblk0p1

# cat /etc/hostname
avocado-jetson-orin-nano
```

`/dev/mmcblk0p1` is the **SD card**, and that hostname is the stock image on the SD - the NVMe carries a marker image with a different one.

The kernel came from the NVMe. The rootfs came from the SD card. Nothing logged a warning; the only outward sign was the wrong hostname on the login banner.

## Why, on the same board

Both disks carry identical partition labels, which is what makes the label useless as an identifier:

```
mmcblk0    nvme0n1
p1  APP    p1  APP
p2  APP_b  p2  APP_b
...
```

And the kernel command line names no root at all:

```
# cat /proc/cmdline
 bl_prof_dataptr=2031616@0x271E10000 bl_prof_ro_ptr=65536@0x271E00000
```

No `root=`, no `console=`. That is the boot.img repack dropping the command line, which is the half of this PR that restores it - and without a `root=` there is nothing for the initrd to honour even if it wanted to, so it falls through to `blkid -t PARTLABEL=APP` and takes whichever disk the kernel enumerated first.

## What this does and does not prove

It proves the defect, precisely, on the hardware this change targets: the mismatch is real, it is silent, and it survives a boot order that points at the right disk. Reordering `BootOrder` alone does not fix it - this board's boot order already pointed at the NVMe.

It does not yet prove the fix. That needs an image built from this branch, flashed with the SD still in, and `findmnt -no SOURCE /` reporting `/dev/nvme0n1p1` on the next boot. That run has not happened.

jetm:
Rebased onto scarthgap to clear the conflict with #390, which touched the same file (avocado-tegra-init) to fix shellcheck findings on the command-line parsing loop. Kept #390's `read -r cmdline` / `for bootarg in $cmdline` form and folded in this PR's `partuuid_a`/`partuuid_b` declarations and case arms, and kept this PR's restructuring that closes the slot-detection `if` block early so the PARTUUID lookup runs unconditionally before the PARTLABEL fallback. The other two touched files (tegraflash-bsp.bb and stone-provision-tegraflash.sh) merged with no manual intervention since #390's edits sat in unrelated parts of those files.

No behavior changed beyond the two PRs' own intent: the PARTUUID pin still tries first, and the PARTLABEL wait still falls back at the end exactly as before, just wrapped in the shellcheck-clean loop form scarthgap now expects.

jetm:
## Verified end to end on a Jetson Orin Nano devkit

Provisioned NVMe with the SD card left installed, which is the only configuration in which this test means anything. Both disks carry a full layout and both are bootable.

**The ambiguity the label-based discovery could not resolve is still present on the board:**

```
# grep PARTNAME /sys/block/mmcblk0/mmcblk0p1/uevent /sys/block/nvme0n1/nvme0n1p1/uevent
/sys/block/mmcblk0/mmcblk0p1/uevent:PARTNAME=APP
/sys/block/nvme0n1/nvme0n1p1/uevent:PARTNAME=APP
```

**The command line names the rootfs:**

```
# cat /proc/cmdline
avocado.root_partuuid=49138F00-0C01-4B18-88A8-40E5D8E8DD54 avocado.root_partuuid_b=623AFD90-F9EA-4C59-A65E-9AF6306AC837 bl_prof_dataptr=2031616@0x271E10000 bl_prof_ro_ptr=65536@0x271E00000
```

**That PARTUUID is on the disk, and it is the NVMe:**

```
# ls -l /dev/disk/by-partuuid/49138f00-0c01-4b18-88a8-40e5d8e8dd54
lrwxrwxrwx 1 root root 15 ... /dev/disk/by-partuuid/49138f00-0c01-4b18-88a8-40e5d8e8dd54 -> ../../nvme0n1p1
```

**And it is what got mounted:**

```
# grep ' / ' /proc/mounts
/dev/nvme0n1p1 / erofs ro,relatime,user_xattr,acl,cache_strategy=readaround 0 0

# cat /etc/hostname
avocado-orin-sdtest
```

`avocado-orin-sdtest` is the marker image written to the NVMe. The SD card carries the stock image, whose hostname is `avocado-jetson-orin-nano`. Before this change the same board booted the NVMe kernel and the SD rootfs and printed that stock hostname at the login banner, with nothing else to tell them apart.

The chain is closed: provisioning generates the UUID, writes it into the layout, `make-sdcard.sh` hands it to `sgdisk`, the GPT holds it, the repacked boot.img names it, and the initrd resolves it.

## Two defects this run found

Neither was visible without hardware, and the second was not visible from this PR's diff at all.

**The `KERNEL_ARGS` carry-forward hung the board.** Silent, partway into userspace, no panic, no further console output, reproduced on two independent flashes. The hunk existed to duplicate #347, a wrynose fix verified on AGX Thor against a tegra264-specific console mismatch. tegra234 does not have that problem and does not survive the remedy. Dropped in 4607d6f0; an empty mkbootimg cmdline reaches a login prompt on this machine every time.

**`make-sdcard.sh` never passed `--partition-guid`.** `nvflashxmlparse.py` parses `unique_guid` out of the layout and emits `partguid="..."` into the shell line `make-sdcard.sh` evals. `make_partitions()` assigned it to a declared local and then built the sgdisk command without it, so sgdisk chose a random GUID and the pinned value reached nothing. Provisioning reported success while the initrd searched for a PARTUUID that had never been written. Fixed in 1cf95fd9.

I lost time on the second one to a wrong theory worth recording: I assumed a timing race, because the symptom looks exactly like one. It is not. A diagnostic extending the search from 10s to 30s showed NVMe fully enumerated *before* the search began, and 56 seconds of searching still failed. The diagnostic was reverted in 63edeba8 once it had answered the question.

PR #392, opened 2026-09-16
Make every shell script pass shellcheck, and gate it

jetm:
> Stacked on #390. Review that one first; this targets its branch, not `scarthgap`.

## Problem

120 of the repository's 186 shell scripts failed the shellcheck hook this repo already ships.

That combination is worse than having no gate. The hook only sees files a commit touches, so anyone editing one of those 120 met findings they did not introduce, and the way past that is `--no-verify`, which switches off every hook rather than the one in the way. These are the scripts that decide which disk a board boots from and which device an image is written to, so that is the wrong reflex to be training.

The formatter had the same shape of problem, more quietly. `.pre-commit-config.yaml` ran shfmt with `args: [-i, "2", -ci, -bn]` and no `-w`, so shfmt printed the formatted result to stdout and exited 0. pre-commit reported a passing hook on every commit while never changing a file, which is why the tree was still 4-space indented long after that config landed.

## Solution

Fix the formatter, apply it once, fix the findings, then gate it so a finding in CI is a finding the pull request introduced.

The commits are ordered so the mechanical output is separable from the code:

1. `pre-commit: give shfmt the -w that makes it write` - one line
2. `treewide: apply the shfmt style the repo already declares` - 147 files, pure formatting, **skip it**
3. `scripts:` host build and packaging shell - 22 files
4. `bsp:` flashing and provisioning shell - 35 files
5. `sdk:` per-machine build and provision hooks - 51 files
6. `ci: gate every shell script on shellcheck`

Commits 3-5 are 997 insertions of pure code - identical with and without `git diff -w`. Before the split, the same content read as 3570 insertions.

## Key changes

- **Four live bugs, not lint.** `make-sdcard.sh` declared `local` at top level, so under `set -e` it aborted right after "Disconnecting" and the USB flush and final disconnect never ran. The rubikpi3 UFS script's `cd ... && tar ... || tar ...` fell through on a failed `cd` and would repack the caller's working directory into the image, immediately after the original was moved aside to `.backup`. `stone-provision-img.sh`, in four byte-identical copies, read its manifest through an unquoted `cat $AVOCADO_STONE_MANIFEST`, so a build path containing a space left `fwup -d` pointed at `/.img`, failing open. `avocado-devicetree-apply` died on any fragment filename containing whitespace.
- **The qcom init scripts used `[[ ]]` under busybox ash**, where it works by accident: `start` and `stop` dispatch, but any other argument writes "unknown operand" and an argument containing a glob character is expanded against the cwd.
- **The SDK wrappers quoted wrongly inside `eval`.** `\"$x\"` lets the outer shell split and glob before eval sees the value, which collapsed runs of spaces in a path. Now `"\"$x\""`.
- **False positives are silenced with the reason beside them**, never quoted into something that breaks. The recurring one is a variable holding a built-up argument list - quoting `$flash_args` or `$include_flags` passes one argument containing spaces and breaks the flash.
- **The CI job discovers by shebang as well as extension.** More than half these scripts carry no `.sh` suffix, which is how they drifted. It also refuses a discovery count below 150 rather than trusting its own exit status, since a glob matching nothing lints nothing and exits 0.

## Reviewer notes

**Skip commit 2 and read 3-5.** That is what the split is for. Commit 2 is 147 files of indentation, `-bn` moving pipes to the start of a continuation line, and `> file` losing its space.

**The feedback I most want is on the silenced findings.** Roughly 140 findings are disabled rather than fixed, each with a stated reason. Those reasons are the judgement calls in this change - particularly every argument-list `SC2086`, where I chose to keep word-splitting. If any of those is wrong, it breaks a flash, and it will not show up in CI because CI now passes by construction.

Behaviour-risky substitutions were checked by execution rather than inspection: argv probes against a stub `stone` for the eval quoting, old-vs-new under both bash and busybox for the `[[ ]]` rewrite, and byte-comparison for `awk` replacing a nested `echo $(...)` and `printf` replacing `echo -n`.

Not hardware-validated. Nothing here was run on a board; the provisioning and flashing paths want a real flash before anyone leans on them.

Commit 2 will conflict with any in-flight shell work in this repo. Worth landing or rejecting quickly rather than letting it sit.


jetm:
Rebased this onto current scarthgap. Turned out the two tegra commits here (resolve the OTA target slot, and the shellcheck pass on those flash scripts) landed separately as #390 with byte-identical content, so git dropped both as already-applied and there was nothing left to resolve by hand. Everything else in this branch applied cleanly on top. Tree is clean, no conflict markers left anywhere.

PR #393, opened 2026-09-16
Add a runtime tool for choosing the boot device

jetm:
## Problem

Provisioning decides where an image is written. Firmware decides which disk boots. On a board carrying two bootable disks those are different decisions, and only the first one is ours today.

`avocado provision --profile tegraflash-nvme` writes a complete, bootable NVMe and reports success. If the board also has a provisioned SD card, it keeps booting the SD card. A Jetson makes this vivid: its UEFI creates a `UEFI SD Device` entry unprompted and puts it at the front of `BootOrder`. Nothing in this tree writes `BootOrder` on any target except x86-64, which does it for A/B slot activation.

There is no way to change that from a running board either, because the Jetson image ships neither `efibootmgr` nor `efivar`.

## Solution

A small tool, shipped as a package so a runtime extension or a rootfs package list can pull it in. It moves no data - it tells the firmware which of the disks already present to prefer.

Real output from the bench Orin Nano, entries elided for length:

```console
# avocado-set-boot-device --list
Current boot order, first entry wins:

  Boot0002* UEFI USB Device	VenHw(1e5a432c-...)/MemoryMapped(11,0x3610000,...)
  Boot0003* UEFI Samsung SSD 960 EVO 250GB S3ESNX0JA13241W 1	VenHw(1e5a432c-...)
  Boot0001* UEFI SD Device	VenHw(1e5a432c-...)/MemoryMapped(11,0x3400000,...)
  ...

Recognised device classes on this board:
  nvme  0003
  sd    0001
  emmc  -
  usb   0002

# avocado-set-boot-device --dry-run sd
Boot device: sd (entry 0001)
  before: 0002,0003,0001,0004,0007,0006,0005,0000,0008,0009
  after:  0001,0002,0003,0004,0007,0006,0005,0000,0008,0009
(dry run, BootOrder not written)
```

## Key changes

- `avocado-set-boot-device` takes a device class - `nvme`, `sd`, `emmc`, `usb` - and moves the matching boot entry to the front, keeping the rest of the order so a board whose chosen disk later fails still has a fallback.
- Matching is on the EFI device path node (`NVMe(`, `SD(`) rather than the entry description, which is vendor text differing per drive. The description is a fallback for firmware that omits the node.
- `--once` writes `BootNext` instead, so a test selection reverts by itself. `--dry-run` prints the order and changes nothing.
- The recipe is `allarch` and lives in the common layer, since `BootOrder` is a UEFI concept rather than a Tegra one. Only the tegra feed is wired up here, because that is the hardware it was designed against.
- New doc at `docs/boot-device-selection.md` covering the layer side: where the recipe lives and why it is in the common layer, the `efibootmgr`/`efivar` dependency, and how to wire another BSP's feed. How to install and use the tool belongs on the docs site, and that guide is a separate PR.

Three refusals are deliberate. No matching entry fails and prints the entries that do exist, because the firmware only creates an entry for a disk it has seen and booted - so the answer is usually "reboot once", which the error says. More than one match warns and names them. And both write paths read the variable back and fail if it does not hold what was written, because efivarfs will accept a write the firmware then discards, and a selection that silently did not take is the exact failure this tool exists to prevent.

## Reviewer notes

**Run on a Jetson Orin Nano devkit, with both an SD card and an NVMe provisioned and bootable.** Full output in a comment below. `--list` classifies every entry on the board correctly, including `emmc` as absent, which is right for an Orin Nano module:

```
Recognised device classes on this board:
  nvme  0003
  sd    0001
  emmc  -
  usb   0002
```

`--dry-run` prints the reordering and writes nothing. `--once sd` wrote `BootNext`, read it back, the firmware honoured it and the board came up on the SD card, and the following boot returned to the NVMe on its own with `BootNext` gone and `BootOrder` untouched.

The stub coverage from before still holds: list, select, already-first, next-boot-only, dry-run, an absent device class, and a firmware that accepts the write and discards it.

**This is half of boot-device selection, and the half it is not is #391.** Choosing the boot device settles which kernel the firmware loads. Which rootfs that kernel mounts is decided by the PARTUUID provisioning writes into the kernel command line, which is what #391 adds. Before that change the initrd finds its rootfs by PARTLABEL, and on a board with two provisioned disks both carry a partition labelled `APP` - so reordering `BootOrder` alone gives you the kernel you asked for and the rootfs of whichever disk won the probe race. Neither change is much use without the other.

**The feedback I most want is on the device-class matching.** Four classes are recognised by device-path node, and I have only seen a Jetson's entry list - an Orin's SD entry, an NVMe entry, a network entry and the UEFI shell. If another board's firmware spells its eMMC or USB nodes differently, the class simply will not match and the tool refuses; that is safe, but it is also useless on that board, and I would rather hear about it than guess.


jetm:
## Matchers checked against real firmware, and one of them was wrong

Read the boot entries off the bench Orin Nano rather than reasoning about them. Two results, one good and one not.

**The primary matcher is right.** The device-path nodes are there, in the bytes:

```
Boot0003 (NVMe)  @0xb0:  03 17 10 00  01 00 00 00 ...
                         ^^ ^^         Type 3 (Messaging), SubType 0x17 (NVMe Namespace)

Boot0001 (SD)    @0x54:  03 1a 05 00  00
                         ^^ ^^         Type 3 (Messaging), SubType 0x1a (SD)
```

`efibootmgr` prints those as `NVMe(...)` and `SD(...)`, which is exactly what `class_pattern()` matches on. That was an assumption before; it is now read off the board.

**The description fallback was wrong.** This is the NVMe entry's description:

```
UEFI Samsung SSD 960 EVO 250GB S3ESNX0JA13241W 1
```

Vendor, model, capacity, serial - and the word NVMe nowhere in it. The fallback matched `NVMe\|NVME\|nvme`, so on firmware that omits the device-path node (the only case the fallback exists for) it would have found nothing and refused on a board with an NVMe plainly in it.

Widened in 4a2ffab to take `SSD` and `M.2` as well, and the comment now says outright that a pattern over vendor descriptions is a second guess rather than a second source - a SATA SSD will answer to `nvme` here, and the existing more-than-one-match warning is what surfaces that.

## Also confirmed

`efibootmgr` is **absent** from the running Jetson image, which is the premise for shipping it in this package:

```
# command -v efibootmgr || echo ABSENT
ABSENT
```

The board has ten boot entries (`Boot0000`-`Boot0009`) covering setup, SD, USB, the NVMe, PXEv4/v6 and HTTPv4/v6 - so the `usb` class has a real entry to match here too, not just `nvme` and `sd`. There is no eMMC on an Orin Nano, so `emmc` is correctly empty.

## Still not verified

The tool has not executed on the board, because `efibootmgr` is not there to run it against - that arrives with this package. What is verified is everything it depends on: the node types it matches, the descriptions it falls back to, and separately that a `BootOrder` write through efivarfs persists across a reboot with `BootCurrent` following it.

The end-to-end run wants an image built from this branch, then `avocado-set-boot-device --once nvme`, a reboot, and `BootCurrent` naming the NVMe entry.

jetm:
## Hardware build caught a real defect: fixed in 4d101bd3

Building this on pc2 against a real jetson-orin-nano target (`avocado-complete`) failed at 12423/17898 tasks:

```
ERROR: avocado-boot-device-1.0-r0 do_unpack: Directory name ${@d.getVar('S')
contains unexpanded bitbake variable. This may cause build failures and WORKDIR
polution.
```

The recipe set `S = "${UNPACKDIR}"` with a comment claiming `avocado-tegra-init.bb` does the same "because wrynose no longer auto-creates ${S}". Neither half of that held up: `avocado-tegra-init.bb` does not set `S` at all, on this branch or any other, and I had not actually checked. I wrote a justification for a pattern I invented.

Every other single-script, no-compile `allarch` recipe in this tree - `avocado-dtc-overlay-deliver`, `avocado-tegra-init`, `avocado-devicetree` - leaves `S` at its default and installs from `${WORKDIR}` directly, which is where a plain `file://` fetch with no archive actually lands. Matched that pattern instead of chasing why my override broke.

Rebuilding now to confirm the fix from the failure point (sstate should cover everything up to the point that failed).

jetm:
## Run on the bench Orin Nano

The tool is in the image and executes. Both an SD card and an NVMe are provisioned and bootable, which is what makes a selection test mean anything.

```
# command -v avocado-set-boot-device
/usr/sbin/avocado-set-boot-device

# avocado-set-boot-device --list
Current boot order, first entry wins:

  Boot0002* UEFI USB Device	VenHw(1e5a432c-0466-4d31-b009-d4d9239271d3)/MemoryMapped(11,0x3610000,...)
  Boot0003* UEFI Samsung SSD 960 EVO 250GB S3ESNX0JA13241W 1	VenHw(1e5a432c-...)
  Boot0001* UEFI SD Device	VenHw(1e5a432c-0466-4d31-b009-d4d9239271d3)/MemoryMapped(11,0x3400000,...)
  Boot0004* UEFI PXEv4 (MAC:4CBB47C8C25B)	...
  Boot0007* UEFI HTTPv6 (MAC:4CBB47C8C25B)	...
  Boot0006* UEFI HTTPv4 (MAC:4CBB47C8C25B)	...
  Boot0005* UEFI PXEv6 (MAC:4CBB47C8C25B)	...
  Boot0000* Enter Setup	FvVol(49a79a15-...)/FvFile(462caa21-...)
  Boot0008* BootManagerMenuApp	FvVol(49a79a15-...)/FvFile(eec25bdc-...)
  Boot0009* UEFI Shell	FvVol(49a79a15-...)/FvFile(7c04a583-...)

Recognised device classes on this board:
  nvme  0003
  sd    0001
  emmc  -
  usb   0002
```

All ten entries classify correctly. `emmc` reads `-` because an Orin Nano module has no eMMC, which is the right answer rather than a miss. The `usb` class matched a real entry too, so three of the four classes have hardware behind them here.

`--dry-run` is inert:

```
# avocado-set-boot-device --dry-run sd
Boot device: sd (entry 0001)
  before: 0002,0003,0001,0004,0007,0006,0005,0000,0008,0009
  after:  0001,0002,0003,0004,0007,0006,0005,0000,0008,0009
(dry run, BootOrder not written)
```

## The `--once` round trip

`avocado-set-boot-device --once sd` wrote `BootNext`, read it back and matched. The board rebooted onto the SD card: `/etc/hostname` reported `avocado-jetson-orin-nano` and the root device was `/dev/mmcblk0p1`, against `avocado-orin-sdtest` on `/dev/nvme0n1p1` before and after. The next boot returned to the NVMe with no further action.

`BootNext` is gone and `BootOrder` is untouched, which is the property `--once` is for:

```
# ls /sys/firmware/efi/efivars/ | grep -E '^Boot(Next|Order|Current)-8be4'
BootCurrent-8be4df61-93ca-11d2-aa0d-00e098032b8c
BootOrder-8be4df61-93ca-11d2-aa0d-00e098032b8c
```

## One thing to know before this is useful to anyone

Getting the package into an image does not work the way the docs-site guide originally described. `extensions.<name>.packages` in `avocado.yaml` is accepted, dnf reports the package installed during `avocado install`, and the extension sysroot is never persisted, so the built extension image does not contain the binary. I filed that as avocado-linux/avocado-cli#283 with a reproduction.

The path that works today, and the one the guide now documents, is `rootfs.packages`:

```yaml
rootfs:
  permissions: dev
  packages:
    avocado-pkg-rootfs: "*"
    avocado-boot-device: "*"
```

That is what produced the image the output above came from. Nothing in this PR needs changing for it; the recipe and the packagegroup wiring are correct either way.

PR #394, opened 2026-09-17
fluentbit: set S to the git unpack directory

mobileoverlord:
Both distro legs of build 215 that reached it failed at do_patch:

  ERROR: fluentbit-5.0.8-r0 do_patch: Applying patch
  '0001-lib-Do-not-use-private-makefile-targets-in-CMakelist.patch' on target
  directory '.../fluentbit/5.0.8/fluentbit-5.0.8'
  can't find file to patch at input line 35
  No file to patch.  Skipping patch.

SRC_URI is a git:// fetch, which unpacks to ${WORKDIR}/git. The recipe never set
S, so it took the scarthgap default ${WORKDIR}/${BP} = fluentbit-5.0.8, a
directory nothing creates. quilt ran against an empty path, so every patch
failed. Every other git-fetched recipe in this layer sets S = "${WORKDIR}/git"
(avocadoctl, stone, sscg, mkfat, uv, systemd-zram-generator).

This has never worked. fluentbit is absent from the 2024 feed and no commit has
ever carried an S line. The recipe arrived in #270 and #280 fixed a parse failure
in it, so do_patch had not once been reached: builds died earlier every time
(211 on SPDX, 211/212 on yubico-piv-tool, 213 on ipp-usb, 214 on dav1d). 215 is
the first build to get this far, which is also why the fix is the whole recipe
rather than one task -- every task past do_patch was equally unproven.

Verified by building it, not by parsing it: a qemuarm64 build dir in an
avocadolinux/yocto-build container against the shared sstate cache, on the stock
2.8-avocado bitbake pin.

  before:  S=".../fluentbit/5.0.8/fluentbit-5.0.8"
           No file to patch.  Skipping patch.
           Task (fluentbit_5.0.8.bb:do_patch) failed with exit code '1'

  after:   task do_fetch: Succeeded
           task do_unpack: Succeeded
           task do_patch: Succeeded
           full build: 0 tasks failed, warnings only

           fluentbit-5.0.8-r0.0.cortexa57.rpm       2621371
           fluentbit-dbg-5.0.8-r0.0.cortexa57.rpm   9210177
           fluentbit-dev-5.0.8-r0.0.cortexa57.rpm    170316
           fluentbit-src-5.0.8-r0.0.cortexa57.rpm   3930535


nicksinas (approved):

jetm (commented):
Reviewed the `S` fixup for `fluentbit_5.0.8.bb`. Confirmed the mismatch is real (git fetcher unpacks to `${WORKDIR}/git`, BitBake's default `S` is `${WORKDIR}/${BP}`), matched against the quoted `do_patch` error, and checked it against 20+ sibling git-fetched recipes in this layer that all set `S = "${WORKDIR}/git"` - this follows established convention rather than introducing a new one. `do_configure`/`do_compile` already reference `${S}` generically, so the fix makes that existing assumption correct. No issues found.

PR #395, opened 2026-09-17
intel-x86-64: wrynose parity for v2/v3, and make v4 buildable

mobileoverlord:
Brings the Intel x86-64 targets on wrynose up to scarthgap parity, and makes
x86-64-v4 build for the first time.

## What

Forward-ports five scarthgap commits that never reached wrynose:

| Commit | Brings |
|---|---|
| #243 | NVIDIA dGPU 580.119.02 -> 595.84 (firmware, userspace, open modules), `CONFIG_BINFMT_SCRIPT`, br_netfilter + IPv6 netfilter modules, driver on all x86-64 targets, disk-image provisioning fixes |
| #251 | `IMAGE_PACKAGES_ONLY_FORCE_PROVIDERS` so `kernel-module-nvidia` reaches the feed |
| #255 (part) | `elfutils-native` so objtool finds libelf |
| #254 | reject a non-512 block size when provisioning |
| #258 | CAN, TSN shapers, automotive PHY and board I/O modules |

Plus three wrynose-only fixes found by building:

- **`KERNEL_CC` for the NVIDIA modules.** `KERNEL_CC` carries `-ffile-prefix-map`
  for the kernel dirs; `module.bbclass` passes it for compiles it drives, but this
  recipe has its own `do_compile` and overrode `CC` with a plain `${CC}`. The
  modules then recorded real build paths via `__FILE__` in inlined
  `WARN_ON`/`BUG_ON`, failing the `buildpaths` QA. Only visible on 6.18, where
  `file_ref.h` and `ucopysize.h` inline those macros.
- **`MACHINE_FEATURES_BACKFILL_CONSIDERED` -> `MACHINE_FEATURES_OPTED_OUT`.**
  oe-core renamed it and registers the old spelling in `BB_RENAMED_VARIABLES`, a
  hard parse error, so x86-64-v4 did not parse at all. Under the old spelling the
  opt-out it guards would also have silently done nothing.
- **Native-exec `qemuwrapper` for x86-64-v4.** See below.

## How x86-64-v4 is unblocked

Four facts that cannot all hold under emulation:

- v4 opts out of `qemu-usermode` because QEMU's TCG implements no AVX-512
  (`target/i386/cpu.c`: no AVX-512 bit in any `TCG_*_FEATURES` mask, AVX2 is the
  ceiling), and the v4 userland uses it -- the built `udevadm` carries 197
  AVX-512 instructions.
- `do_rootfs` runs postinst intercepts (`update_udev_hwdb`, glib's gio module
  cache) by invoking the *target* binary through `qemuwrapper`.
- The stock wrapper hard-exits when `qemu-usermode` is absent.
- The rootfs is read-only erofs, so postinsts cannot defer to first boot.

The build host is x86-64 too, so the target binary needs no emulation. A
`qemuwrapper-cross` bbappend scoped to `intel-x86-64-v4` installs a wrapper that
execs through the target's own dynamic loader, re-rooting the binary's absolute
RUNPATH inside the rootfs the way `qemu -L` does. `qemu-usermode` stays out of
`MACHINE_FEATURES`, so meson still gets no `exe_wrapper`, which is what the
opt-out existed for.

v2 and v3 are untouched: SSE4.2 and AVX2 are inside TCG's range, so their proven
qemu path is unchanged (`AVOCADO_NATIVE_EXEC_REQUIRES_FLAGS` and `SRC_URI` are
both empty there).

**This makes x86-64-v4 require an AVX-512 build host.** An anonymous python check
compares the required flags against `/proc/cpuinfo` and fails at parse naming the
missing ones, rather than dying ~95% into the build.

## Results

| Target | Build | NVIDIA RPMs in feed | `nvidia.ko` TMPDIR refs |
|---|---|---|---|
| intel-x86-64-v2 | pass | 5 modules + firmware + userspace | 11 -> 0 (11 now `/usr/src/kernel/...`) |
| intel-x86-64-v4 | pass | 5 modules + firmware + userspace | 0 |

Both produce kernel, erofs rootfs, initramfs, var image and stone bundle.
`bitbake -p` and `bitbake -g avocado-complete` are clean on v2, v3 and v4.

intel-x86-64-v3 is not yet rebuilt against the final tree; it shares every change
with v2 and is expected to match, but it is not claimed as verified here.


meta-avocado-x86-64/stone/intel-x86-64-v2/stone-provision-img.sh
```diff
@@ -53,6 +64,54 @@ size_to_mib() {
...
+    local src="$1" dst="$2" seek_mib="$3" label="$4"
+    local total_bytes total_mib dd_pid pos written seek_bytes
+    total_bytes=$(stat -c%s "$src")
+    total_mib=$(( (total_bytes + 1048575) / 1048576 ))
+    seek_bytes=$(( seek_mib * 1048576 ))
+    echo "  Writing ${label}: ${total_mib} MiB"
+
+    # bs=1M with a whole-MiB seek avoids the newer oflag=seek_bytes dd operand.
+    dd if="$src" of="$dst" bs=1M seek="$seek_mib" \
+       conv=notrunc,sparse status=none </dev/null &
```
jetm, line 94: **`conv=sparse` can leave stale bytes on rebuild**

`write_image()` switched to `conv=notrunc,sparse`, but nothing zeroes `IMAGE_FILE` first: `truncate -s` doesn't clear existing bytes and `sgdisk --zap-all` only touches GPT metadata. A partition rebuilt with different content than a prior run can keep stale data wherever the new source has a zero block.

Rebuild the same board's image in the same `BUILD_DIR` - the normal incremental dev loop - with a var or rootfs image that differs from the last run. `dd` skips every zero block in the new source, so the destination keeps whatever bytes the previous build left there. The shipped image carries leftover data from an earlier build, and nothing prints an error. Applies identically to the v3 and v4 copies of this script.

meta-avocado-x86-64/recipes-devtools/qemu/files/qemuwrapper-native
```diff
@@ -0,0 +1,111 @@
...
+            fi
+        done
+    done
+done
+
+# Strip any leading ':' so the loader does not read it as the current directory.
+LIBPATH=$(echo "$LIBPATH" | sed 's/^://')
+
+if [ -n "$LIBPATH" ]; then
+    exec "$LOADER" --library-path "$LIBPATH" "$@"
```
jetm, line 108: **`--library-path` can fall back to host libs**

The comment above this block says the rootfs's library directories go on `--library-path` "the way `qemu -L` does." They don't: `--library-path` only prepends to ld.so's search order. `qemu -L` fully re-roots lookups; glibc's loader still falls through to `/etc/ld.so.cache` and the build host's own `/lib` and `/usr/lib` for anything it doesn't find first.

A `NEEDED` library that's missing, unpacked into a package not yet staged, or nested more than one level below the scanned roots (the comment's own stated ceiling) resolves to the host's copy instead of failing. `update_udev_hwdb` or `gio-querymodules` then runs "successfully" against a host library and bakes a cache built against the wrong version into the image. Nothing in the log says so.

meta-avocado-x86-64/recipes-devtools/qemu/qemuwrapper-cross_%.bbappend
```diff
@@ -0,0 +1,83 @@
...
+            "emulate AVX-512, so they are executed directly on the build host instead.\n"
+            "That requires a host CPU implementing the full v4 feature set.\n"
+            "\n"
+            "Build %s on an AVX-512 host (Intel Skylake-SP and later Xeon, AMD Zen 4\n"
+            "and later). x86-64-v2 and x86-64-v3 have no such requirement and build\n"
+            "anywhere."
+            % (d.getVar('MACHINE'), ' '.join(missing), d.getVar('MACHINE')))
+}
+
+do_install:append:intel-x86-64-v4() {
```
jetm, line 80: **v4 native-exec swap also hits nativesdk**

oe-core's stock `qemuwrapper-cross` recipe exempts the nativesdk build from its qemu-usermode guard explicitly (`"${PN}" != "nativesdk-qemuwrapper-cross"`). This bbappend's `do_install:append:intel-x86-64-v4` and its parse-time AVX-512 check (lines 39/44) both fire for every recipe carrying the `intel-x86-64-v4` override, with no equivalent carve-out - that includes `nativesdk-qemuwrapper-cross`.

Fetching or parsing the x86-64-v4 SDK on a non-AVX-512 build host now fails at parse time with the new `bb.fatal`. An SDK that does get built on an AVX-512 host ships a wrapper that execs AVX-512 target binaries straight on whatever laptop it lands on - SIGILL on any host that isn't also AVX-512, where the old qemu wrapper would have emulated.

meta-avocado/classes/image-packages-only.bbclass
```diff
@@ -25,3 +25,34 @@ IMAGE_FEATURES = ""
...
+# the machine/vendor conf next to the PKG_EXTRA_INSTALL that adds the packagegroup.
+# The class turns each into a hard build+deploy dependency of do_build. do_build
+# is deliberately chosen over the packagegroup's own do_package_write_rpm: it is
+# never sstate/setscene-covered, so a shared-sstate build that restores the
+# packagegroup from setscene cannot prune the dependency, and the provider is
+# always built or setscene-restored (either path deploys its RPMs into
+# DEPLOY_DIR_RPM, which is what the feed index scans).
+IMAGE_PACKAGES_ONLY_FORCE_PROVIDERS ??= ""
+
+python () {
```
jetm, line 53: **SDK builds now hard-depend on nvidia modules**

`IMAGE_PACKAGES_ONLY_FORCE_PROVIDERS` is set once, machine-globally, in `avocado-x86-64.inc`. The new anonymous `python()` here applies it to every recipe inheriting the class - `avocado-pkg-sdk.bb`, `avocado-pkg-sdk-extra.bb` and `avocado-pkg-sdk-all.bb` all inherit it too, not just the feed-only `avocado-pkg-extra.bb` this was written for.

Building any of the three x86-64 SDK recipes now carries a hard `do_build` dependency on `nvidia-gpu-modules:do_package_write_rpm`. An SDK build that has no use for a target kernel-module RPM pulls in a large out-of-tree git fetch and a full kernel-module compile it never needed before this change.

meta-avocado-x86-64/recipes-avocado/packagegroups/packagegroup-avocado-x86-64-extra.bb
```diff
@@ -65,7 +65,10 @@ RDEPENDS:${PN} = " \
   ${CPU_MICROCODE} \
   ${WIRELESS_TOOLS} \
   ${HARDWARE_TOOLS} \
-  packagegroup-avocado-nvidia-gpu \
   kernel-modules \
   swupdate \
 "
+
+# NVIDIA GPU driver is installed unconditionally on every x86-64 target via
+# PKG_EXTRA_INSTALL in conf/machine/include/avocado-x86-64.inc, so it is not
```
jetm, line 73: **Extra packagegroup no longer pulls nvidia**

`packagegroup-avocado-nvidia-gpu` comes out of this packagegroup's own `RDEPENDS` on the reasoning that `PKG_EXTRA_INSTALL` covers it now. `PKG_EXTRA_INSTALL` only feeds the feed-only `avocado-pkg-extra.bb` image at build time - it has no effect on what installing this packagegroup pulls at runtime.

A device running `avocado install packagegroup-avocado-x86-64-extra` after deployment used to pull the NVIDIA driver stack transitively. It no longer does, even though the RPMs still sit in the feed under their own package names.

meta-avocado-x86-64/recipes-devtools/qemu/files/qemuwrapper-native
```diff
@@ -0,0 +1,111 @@
...
+            # emulated binary, -cpu picks the emulated model. Running natively,
+            # the real kernel and real CPU answer instead.
+            shift 2
+            ;;
+        --)
+            shift
+            break
+            ;;
+        -*)
+            echo "qemuwrapper(native): ignoring unsupported qemu option '$1'" >&2
```
jetm, line 47: **Unknown option desyncs arg parsing**

Only `-L`, `-E`, `-r`, `-cpu` and `-g` consume a following argument. Any other value-taking option OE ever passes (a future `QEMU_EXTRAOPTIONS`:tune-* addition, say) falls into the catch-all branch, which shifts by one and logs "ignoring" - leaving that option's value behind as the next positional argument.

That leftover value gets read as the target binary path once parsing ends, so the wrapper execs the loader against a path that doesn't exist instead of the real binary. It fails with a plain "No such file or directory" right after the log line said the option was safely ignored.

jetm (commented):
6 findings inline, all in the Intel x86-64 provisioning/qemu-wrapper/SDK-dependency changes: a rebuild that can leave stale bytes in a partition, a library-path comment that overstates what `--library-path` actually does (real host-library fallback risk), the v4 native-exec swap missing the upstream nativesdk carve-out, an SDK dependency scope widened to all three SDK recipes instead of just the feed-only one, an "extra" packagegroup that silently stopped pulling the NVIDIA driver stack it used to, and an arg-parsing gap that can desync and exec the wrong path. A longer advisory list came up too (a sibling qemux86-64 script left with the same bugs this PR fixes elsewhere, an inconsistent jq-escaping pattern, some pre-existing partition-overrun and symlink-following gaps, a factual slip in an earlier commit message) - noted for the author but not posted here.

PR #396, opened 2026-09-18
RB3 Gen 2: board support and Adreno GPU enablement

mobileoverlord:
## What

Avocado OS support for the Qualcomm RB3 Gen 2 (QCS6490) plus Adreno 643 GPU enablement.

- **Board:** machine + SoC config (avocado-qcs6490.inc, qcom-rt multiconfig), UKI-embedded device tree with overlay delivery, avocado-shm, qcom SDK build/provision hooks, bootbins/partconf/partition XML, stone provision scripts, avocado-var-grow.
- **GPU:** Mesa/Freedreno userspace in packagegroup-avocado-qcom-extra, and a UKI-embedded overlay (`qcs6490-rb3gen2-dp-disable`) that disables the broken USB-C DisplayPort output so the msm DPU aggregate completes and the GPU binds.

## How

The USB-C DP controller (`mdss_dp` @ae90000) has an incomplete retimer/altmode chain and fails probe (-EINVAL). Because the msm DPU binds every enabled display output plus the GPU as one component aggregate, that one deferred output stalls the whole thing and there is no `/dev/dri`.

It is disabled via a UKI-embedded device-tree overlay merged through `AVOCADO_UKI_DTB_OVERLAYS`, **not** a base-dts change: UEFI's BoardInfoDxe DT-fixup panics at boot on a base dtb with the node disabled and bricks the board into XBL dload, whereas sd-stub's `EFI_DT_FIXUP` handles the overlaid embedded tree (the same path `kodiak-el2` and `qcs6490-rb3gen2-staging` ride). The aggregate then completes over the working outputs — mini-DP (`mdss_edp`) and HDMI (DSI -> LT9611).

The eDP PHY and LT9611UXC bridge kernel modules that those outputs need are carried by the BSP extension (separate PR in bsp-rb3gen2).

## Results

Verified on RB3 Gen 2 hardware:
- **Cold provisioned boot:** `/dev/dri` (card0 + renderD128), `card0-DP-1` (mini-DP) + `card0-HDMI-A-1` connectors, a660 GMU firmware loaded, GPU bound.
- **A/B OS update** (`avocado deploy`): a rootfs delta applied over the air, booted clean, GPU intact.

## Scope

GPU-only. The Hexagon CDSP/NPU is not wired up and its feed additions are deliberately dropped from this branch.

_History note: the branch carries the earlier GPU+NPU commits with a later commit removing the NPU feed; squash-merge for a clean result._


meta-avocado-qcom/conf/machine/avocado-rb3gen2.conf
```diff
@@ -25,6 +25,7 @@ KERNEL_DEVICETREE = " \
     qcom/qcs6490-rb3gen2-industrial-mezzanine.dtbo \
     qcom/kodiak-el2.dtbo \
     qcom/qcs6490-rb3gen2-staging.dtbo \
+    qcom/qcs6490-rb3gen2-dp-disable.dtbo \
```
nicksinas, line 28: A brick-class overlay enters `KERNEL_DEVICETREE` with no marker here.

**Reasoning.** `qcs6490-rb3gen2-dp-disable.dtbo` now lands in `DEPLOY_DIR_IMAGE` beside the other dtbos, and on this line it reads as an ordinary overlay. The rule that keeps the board alive, embedded tree only and never the base dtb, is 44 lines below at line 72. Lines 114 to 120 of this same file propose the next step for mezzanine support: "Applying one at build time with fdtoverlay, or teaching the deploy recipe to emit a FIT."

**How it happens.** Someone writes that mezzanine step by iterating the dtbos in `KERNEL_DEVICETREE`, or adds this name to a base-dtb merge list, and does not read as far as line 72. The build stays green and produces a base dtb with `mdss_dp` disabled.

**Impact.** Per your own testing, UEFI's BoardInfoDxe then panics and the board drops into XBL dload. That failure lands before systemd-boot counts a boot attempt, so A/B rollback cannot recover it. Recovery is a physical EDL power cycle, which is not available on a fielded unit.

**Suggested fixes.**

1. Put the constraint at the hazard. Add an inline marker on this line:

```
    qcom/qcs6490-rb3gen2-dp-disable.dtbo \   # EMBEDDED TREE ONLY - see AVOCADO_UKI_DTB_OVERLAYS below; merging this into the base dtb bricks the board
```

2. Stronger: make it a parse-time error rather than a comment. Add an anonymous python function, or a check in `linux-avocado-qcom-uki.bb`, that fails the build if this overlay name reaches any base-dtb merge list. The comment cannot survive a refactor. A check can.

meta-avocado-qcom/recipes-kernel/linux/files/qcs6490-rb3gen2-dp-disable.dtso
```diff
@@ -0,0 +1,33 @@
...
+
+/*
+ * Disable the broken USB-C DisplayPort altmode output (mdss_dp @ae90000) on the
+ * RB3 Gen 2. Its retimer/altmode chain is incomplete on this board (pmic_glink
+ * altmode fails to acquire the retimer-switch), so the DP controller fails probe
+ * with -EINVAL ("invalid resource"). Because the msm DPU binds every display
+ * output plus the Adreno GPU as one component aggregate, that permanently-
+ * deferred output stalls the whole aggregate: the GPU/GMU never bind and there
+ * is no /dev/dri. Disabling it lets the aggregate complete over the working
+ * outputs (HDMI via DSI -> LT9611).
```
nicksinas: The stated failure model contradicts itself, and it contradicts the PR body.

**Reasoning.** Lines 9 to 12 state a rule: the msm DPU binds every display output plus the GPU as one aggregate, so a permanently deferred output stalls the whole thing. Three statements then disagree about what follows from it.

- This line says the aggregate completes over "the working outputs (HDMI via DSI -> LT9611)", and names no DP output.
- Lines 22 to 27 say the eDP PHY at `aec2a00` "currently never finishes probe".
- The PR body says the aggregate completes over mini-DP (`mdss_edp`) and HDMI, and Results reports `card0-DP-1` present with the GPU bound.

Under the rule as written, an eDP PHY that never finishes probe stalls the aggregate too, and there is no `/dev/dri`. The board demonstrably works, so the rule as written is incomplete or one of the statements is wrong.

**How it bites.** The next engineer to debug display on this board reads the rule, concludes that every deferred output must be disabled, and disables `mdss_edp`. Lines 22 to 27 record that exact change as a brick into XBL dload. The file both suggests the change and forbids it, and the reasoning for the distinction is missing.

**Impact.** A bricked board recoverable only by a physical EDL power cycle, reached by following the file's own logic.

**Suggested fixes.**

1. Resolve which statement holds. Either the eDP PHY does finish probe, and the NOTE at line 27 is stale, or a deferred `mdss_edp` does not stall the aggregate while a failed `mdss_dp` does. If it is the second, say why the two differ, for example a hard probe failure versus `-EPROBE_DEFER`, because that distinction is the whole rule.
2. Name `card0-DP-1` as `mdss_edp` explicitly in this comment. A reader meets `DP-1` in the Results and maps it to the node you just disabled.
3. Make this line list both working outputs if mini-DP is one of them, so the file and the PR body agree.

meta-avocado-qcom/recipes-avocado/packagegroups/packagegroup-avocado-qcom-extra.bb
```diff
@@ -54,6 +58,11 @@ RDEPENDS:${PN}:append:qcm6490 = " \
     packagegroup-qcom-boot-essential \
     packagegroup-qcom-boot-additional \
     mesa \
+    mesa-vulkan-drivers \
```
nicksinas, line 64: The Turnip ICD ships with no Vulkan loader, so nothing on the device can use it.

**Reasoning.** `mesa-vulkan-drivers` installs `libvulkan_freedreno.so` plus an ICD manifest under `/usr/share/vulkan/icd.d/`. An application never loads an ICD directly. It links `libvulkan.so.1`, which comes from the separate `vulkan-loader` recipe, and the loader is what reads the manifest and dispatches to Turnip. `mesa-vulkan-drivers` does not RDEPEND on it. `vulkan-loader` appears nowhere in this tree:

```
$ grep -rn "vulkan-loader" .
kas/machine/imx95-frdm.yml:22:    # vulkan-loader which is only built when 'vulkan' is in DISTRO_FEATURES.
```

**How it happens.** A BSP extension installs `mesa-vulkan-drivers` from the feed, which is the flow this packagegroup exists for, and the user runs any Vulkan application.

**Impact.** The application fails at load with a missing `libvulkan.so.1`, or `vulkaninfo` reports zero ICDs. `/dev/dri` is present and the GPU is bound, so the failure reads as a driver problem rather than a missing package. GLES and EGL work at the same time, which makes it look stranger still.

**Suggested fix.** `vulkan` is already in `DISTRO_FEATURES_EXTRA` in `kas/vendor/qcom.yml`, so the loader builds. Add it here:

```
    mesa-vulkan-drivers \
    vulkan-loader \
```

Consider `vulkan-tools` as well. It provides `vulkaninfo`, which is the check for the "Turnip ICD loaded" claim during bring-up.

mobileoverlord:
On-device verification (RB3 Gen 2, bsp 0.3.2 with vulkan-loader): `vulkaninfo --summary` now reports the Turnip driver —
```
deviceName = Turnip Adreno (TM) 643
driverID   = DRIVER_ID_MESA_TURNIP
driverName = turnip Mesa driver (Mesa 26.0.5)
apiVersion = 1.3.335
```
Confirms the vulkan-loader addition; the Turnip ICD is loaded and usable, not just present.

Issue #399, opened 2026-09-18
OTA A/B pre-install is inoperative on Tegra: rootfs-pre.sh needs util-linux blkid, the rootfs has busybox's


jetm:
## Summary

#390 made `rootfs-pre.sh` resolve the OTA target slot on the disk the system actually booted from, which is the right fix and matters on any board provisioned to more than one medium. Both of the constructs it added depend on `util-linux` `blkid`, and the Tegra rootfs ships busybox's. Neither works, and one of them fails in a way that produces a misleading diagnosis.

Found while trying to verify #390 on hardware. Measured 2026-09-18 on a Jetson Orin Nano devkit with SD and NVMe both provisioned, which is the exact configuration #390 exists for.

## The rootfs has busybox blkid

```
# ls -l $(command -v blkid)
lrwxrwxrwx 1 root root 23 ... /usr/sbin/blkid -> /usr/bin/busybox.nosuid
```

Busybox `blkid` reports `TYPE`, `UUID` and `LABEL`. It does not report `PARTLABEL` or `PARTUUID`, and it does not implement `-t`, `-o` or `-s`.

## Consequence 1: the target is never resolved

`resolve_partlabel_on_disk()` iterates `blkid -t "PARTLABEL=$label" -o device`. On the board:

```
# for c in $(blkid -t PARTLABEL=APP_b -o device); do echo "cand=$c"; done
(no output)
```

Zero candidates, so `count` stays 0 and the function returns 1. `rootfs-pre.sh` then logs `Could not resolve PARTLABEL='APP_b' on /dev/nvme0n1` and exits 1. The update cannot proceed on any board, single-disk or dual.

This is fail-closed, so nothing gets written to the wrong place. It does mean the A/B path is inoperative rather than merely unverified.

## Consequence 2: the safety cross-check reports a mismatch that does not exist

This is the one worth fixing carefully. The check reads:

```sh
RUNNING_LABEL="$(blkid -o value -s PARTLABEL "$ACTIVE_DEV" 2>/dev/null || true)"
if [ -n "$RUNNING_LABEL" ] && [ "$RUNNING_LABEL" != "$ACTIVE_PARTLABEL" ]; then
    log "ERROR: nvbootctrl reports slot $CURRENT_SLOT (active label"
    log "       '$ACTIVE_PARTLABEL') but / is '$RUNNING_LABEL' on $ACTIVE_DEV"
    exit 1
fi
```

[...]

Issue #353, opened 2026-09-01
Jetson A/B updates cannot run: swupdate handlers require nvbootctrl, which the default image does not install


WarrDoge:
`meta-avocado-nvidia`'s own SWUpdate handler scripts require `nvbootctrl` and exit if it is missing:

`recipes-core/swupdate/swupdate/rootfs-pre.sh`
```sh
59:# Ensure nvbootctrl exists
60:if ! command -v nvbootctrl >/dev/null 2>&1; then
61:    log "ERROR: nvbootctrl not found"
...
71:CURRENT_SLOT="$(nvbootctrl -t rootfs get-current-slot ...)"
```

`recipes-core/swupdate/swupdate/rootfs-post.sh` does the same and then calls `nvbootctrl -t rootfs set-active-boot-slot`. Every `stone/stone-jetson-*.json` also invokes it for `slot_detection` and `activate`.

`nvbootctrl` is not in the built `avocado-image-rootfs`:

```
# command -v nvbootctrl
(nothing)
```

The binary itself is available — meta-tegra's `tegra-redundant-boot-base` installs `${sbindir}/nvbootctrl`, and `packagegroup-avocado-tegra-extra` lists that package. But that packagegroup is not part of the default Jetson image, and nothing in the swupdate bbappend pulls it in, so on a default build the A/B update path fails at its first check.

### Suggested fix

Either add the dependency where the scripts are shipped, in `recipes-core/swupdate/swupdate_%.bbappend`:

```
RDEPENDS:${PN}:append:tegra = " tegra-redundant-boot-base"
```

or install `packagegroup-avocado-tegra-extra` in the Jetson image. The RDEPENDS is the tighter of the two, since it ties the requirement to the scripts that actually have it — though it is worth noting `tegra-redundant-boot-base` itself RDEPENDS on `setup-nv-boot-control-service` and `tegra-configs-bootloader`, so it is not a single-binary addition.

Either way, a build-time link seems better than a runtime `log "ERROR: nvbootctrl not found"`, since today the image builds clean and the problem only appears at the first update attempt.

### Note

This does not block us — we drive our own update handler — so treat the priority accordingly. Filing it because the reference configuration looks unintentionally broken rather than deliberately minimal.

[...]

PR #400, opened 2026-09-18
Resolve OTA partition labels from sysfs, not busybox blkid

jetm:
Fixes #399.

## Problem

The Jetson A/B pre-install handler decides which partition SWUpdate writes the rootfs payload to. It made that decision with `blkid`, in two places, and both need util-linux. The Tegra rootfs ships busybox's blkid, which reports TYPE, UUID and LABEL only and implements none of `-t`, `-o` or `-s`.

Nothing declared the dependency, and it is easy to miss because the sibling initrd code gets away with it: `packagegroup-avocado-initramfs` pulls `util-linux-blkid` explicitly, so `avocado-tegra-init` works. Nothing in the rootfs path does, and `tegra-redundant-boot-base` pulls `util-linux-lsblk` rather than blkid, so installing the whole of `packagegroup-avocado-tegra-extra` does not help either.

Measured on the bench board:

```
# ls -l $(command -v blkid)
/usr/sbin/blkid -> /usr/bin/busybox.nosuid

# blkid -t PARTLABEL=APP_b -o device
(nothing)

# v=$(blkid -o value -s PARTLABEL /dev/nvme0n1p1); echo "[$v]"
[/dev/nvme0n1p1: TYPE="erofs"]
```

The target lookup matched nothing, so the update aborted on every board. The cross-check failed worse: the value is not empty, so the guard's "label unavailable, skip this check" branch never ran, and it instead reported that nvbootctrl disagreed with the running partition. Anyone hitting that goes looking for a slot or GPT fault that does not exist.

## Solution

Read `PARTNAME` from each partition's sysfs `uevent`. The kernel publishes the same GPT name there and it needs no package. `avocado-tegra-init` already reads partitions this way in `find_datapart_on_disk()`, so this is the tree's existing answer rather than a new one.

Enumerating one disk's own partition directories is also what scopes the search to the booted disk, which is the property #390 added. The prefix anchoring that kept `/dev/nvme0n11p2` from passing for `/dev/nvme0n1` is gone because the shape no longer admits the mistake: `nvme0n11` is a sibling of `nvme0n1` under `/sys/block`, never a child of it.

## Key changes

- `partname_of()` reads `PARTNAME` from a partition's `uevent`. `resolve_partlabel_on_disk()` walks `/sys/block/<disk>/<disk>*` instead of searching every disk and filtering by name.
- The running-slot cross-check resolves its sysfs path directly from the two values already in hand, rather than searching for the partition.
- An unreadable label now refuses instead of being skipped. Reading the kernel's own PARTNAME changes what a blank means: every GPT partition has one, and nothing else in this script works without GPT names, so a blank says the partition is not the shape the rest of it assumes.
- A missing sysfs directory for the derived disk name says so, rather than reporting the same thing as "no partition carries this label".
- The failure dump lists the candidate set the resolver actually considered, using the resolver's own glob, then says whether the label exists on a disk that was deliberately not searched. That second half is the interesting failure once the search is disk-scoped, and by construction the first half cannot show it.
- `rootfs-post.sh` carried the same construct in its filesystem check and warned on every successful update, which is how a warning stops being read. It parses the value out of either form now.

## The first commit is a separate defect

`1022cd0e` is not about blkid and is worth reading on its own. `resolve_partlabel_on_disk()` returns its answer by echoing it, the caller captures that with `$(...)`, and `log()` wrote to stdout. On the one path where the function logs and refuses - two partitions on the booted disk carrying the same PARTLABEL - the log line became the captured value. It is not empty, so the caller read the refusal as a successful resolution, symlinked the message text as a device path, and reported success while nvbootctrl flipped a slot nothing had been written to.

That is the exact outcome the ambiguity check exists to prevent, and the check was the thing producing it. It predates this PR.

## Verification

Run on a Jetson Orin Nano with SD and NVMe both provisioned and both carrying `PARTNAME=APP`. The functions were extracted from the file rather than retyped, and the md5 of the script that ran on the board matches the local copy, so this exercised the code in the diff:

```
A: running partition label
APP
B: APP_b on booted disk (the OTA target)
/dev/nvme0n1p2
C: APP on the disk we did NOT boot
/dev/mmcblk0p1
D: unknown label must refuse
(refused)
E: nonexistent disk must name the real reason
[ts] ERROR: no sysfs directory for /dev/nvme9n9 (/sys/block/nvme9n9)
(refused)
F: capture is clean when refusing (log must not poison it)
captured=[]
```

C is the property this whole line of work is about: the same label resolving to a different device depending on which disk is asked. F is the first commit.

A full A/B update still cannot be run end to end, because the image ships neither `nvbootctrl` nor `swupdate` - that is #353, and it is independent of this.

## Reviewer notes

**Stacked on #392**, which is the base. `rootfs-pre.sh` is already shfmt-clean there and is not on `scarthgap`, so branching from `scarthgap` would have mixed a whole-file reindent into the diff. GitHub will retarget this when #392 merges.

**The feedback I most want is on the refuse-on-blank change.** It is the one behaviour change here that could bite a configuration I have not seen. My reasoning is that `disk_of()` already rejects anything that is not `/dev/nvme*`, `/dev/mmcblk*` or `/dev/sd*`, so a dm or by-partlabel rootfs exits earlier with a clearer message and never reaches this guard. If there is a Tegra layout where `/` is a GPT partition whose PARTNAME is legitimately empty, this is wrong for it.

One thing I looked at and deliberately did not do: verifying that `/dev/<name>` really is the partition its sysfs directory describes, by comparing the sysfs `dev` file against the node's major:minor. On devtmpfs the names and minors agree by construction, and the check needs a decimal/hex conversion that costs more in complexity than it buys against a static-`/dev` scenario this image does not have.

Separately, `avocado-tegra-init`'s own sysfs walk uses `"${diskname}"p*`, which cannot match `sda1`, so its `/dev/sd*` branch is unreachable. That is a different file on a different branch and wants its own change.


jetm:
## Verified end to end on hardware, with a real nvbootctrl

Jetson Orin Nano devkit, SD and NVMe both provisioned, booted from NVMe. This image carries `nvbootctrl` and `swupdate`, so the handler runs for real rather than exiting at its first check:

```
# nvbootctrl -t rootfs get-current-slot
0
```

Both disks carry the target label, which is the condition the whole change exists for:

```
# grep -m1 PARTNAME /sys/block/mmcblk0/mmcblk0p2/uevent
PARTNAME=APP_b
# grep -m1 PARTNAME /sys/block/nvme0n1/nvme0n1p2/uevent
PARTNAME=APP_b
```

### The merged code, on this board

Both constructs it relies on return nothing useful:

```
# for c in $(blkid -t PARTLABEL=APP_b -o device); do echo "cand=$c"; done
(no candidates)

# v=$(blkid -o value -s PARTLABEL /dev/nvme0n1p1); echo "RUNNING_LABEL=[$v]"
RUNNING_LABEL=[/dev/nvme0n1p1: TYPE="erofs"]
```

So the target never resolves, and the cross-check compares a device description against a partition label, fires, and reports an nvbootctrl disagreement that does not exist.

### This branch, same board, same boot

```
[15:08:07] === Tegra A/B Pre-install Script ===
[15:08:07] Current slot: 0
[15:08:07] Inactive slot: 1
[15:08:07] Target partition label: APP_b
[15:08:07] Booted rootfs: /dev/nvme0n1p1 on /dev/nvme0n1
[15:08:07] Target device resolved: /dev/nvme0n1p2
[15:08:07] Created symlink: /tmp/target_rootfs -> /dev/nvme0n1p2
[15:08:07] Variables written to /tmp/tegra-ab-vars
[15:08:07] Pre-install completed successfully
EXIT=0
```

```
# cat /tmp/tegra-ab-vars
TEGRA_INACTIVE_SLOT=1
TEGRA_TARGET_DEV=/dev/nvme0n1p2
TEGRA_TARGET_PARTLABEL=APP_b
TEGRA_ACTIVE_PARTLABEL=APP
TEGRA_TARGET_SYMLINK=/tmp/target_rootfs

# ls -l /tmp/target_rootfs
/tmp/target_rootfs -> /dev/nvme0n1p2
```

`/dev/nvme0n1p2` is the booted disk's inactive slot. `/dev/mmcblk0p2` carries the same label and was correctly not chosen - which is the property #390 set out to provide and could not deliver on this rootfs.

The script that ran was extracted from this branch rather than retyped, and its md5 on the board matches the local file (`43c3e8204052d706536a01b073cd155e`), so this exercised the committed code.

### Scope of the claim

This verifies the pre-install handler: slot detection through real nvbootctrl, the running-slot cross-check, disk-scoped target resolution, and the artifacts SWUpdate consumes. It does not verify a full `.swu` apply-and-reboot cycle, which additionally needs a bundle and the post-install path.

Getting `nvbootctrl` and `swupdate` onto the image needed a separate step, and that turned out not to be a meta-avocado packaging problem - see #402, which I closed. The published BSP extension already declares both.

Issue #401, opened 2026-09-18
ESP mount unit resolves /dev/disk/by-partlabel/esp, which is ambiguous on a dual-provisioned board


jetm:
## Summary

`setup-nv-boot-control` ships an ESP mount unit that identifies its device by PARTLABEL:

```ini
[Mount]
What=/dev/disk/by-partlabel/esp
Where=/boot/efi
Type=vfat
```

A PARTLABEL names a role, not a disk. A board provisioned to more than one medium carries an `esp` partition on each, so the symlink resolves to whichever partition udev happened to process, and that is not necessarily the disk the system booted from.

This is the same defect class as #390 (OTA target slot) and #391 (initrd rootfs discovery), in a third place.

## Measured

Jetson Orin Nano devkit, SD and NVMe both provisioned, **booted from NVMe** (`/` is `/dev/nvme0n1p1`):

```
# ls -l /dev/disk/by-partlabel/esp
/dev/disk/by-partlabel/esp -> ../../mmcblk0p11
```

Both disks carry one:

```
nvme0n1p11 esp        nvme0n1p14 esp_alt
mmcblk0p11 esp        mmcblk0p14 esp_alt
```

The symlink points at the SD card while the system runs from the NVMe. So `/boot/efi` would mount the ESP of the disk that was not booted.

## Why it is latent right now

The unit is not active on a stock image, because `setup-nv-boot-control` is not installed - it arrives only with `packagegroup-avocado-tegra-extra`, which populates the feed rather than the image. Nothing in meta-avocado references `/boot/efi`, and `setup-nv-boot-control.sh` itself writes EFI *variables* through efivarfs rather than files on the ESP, so nothing reads the wrong mount today.

That changes the moment the package lands in the rootfs, which is what #402 does to make the A/B update path usable. After that the unit is enabled (the recipe's `pkg_postinst` symlinks it into `local-fs.target.wants`) and `/boot/efi` is mounted from an arbitrary disk on any dual-provisioned board.

## Impact

Today: a wrong mount that nothing reads, so the cost is a misleading system state during diagnosis.

When bootloader updates land: anything writing the ESP writes it on the wrong disk, and the board boots firmware from the other one. That is the failure #390 and #391 were about, with the payload being the bootloader rather than the rootfs.

## Options

[...]

PR #402, opened 2026-09-18
Install the A/B update tools in the Jetson rootfs, not just the feed

### LINKED ISSUE #353: Jetson A/B updates cannot run: swupdate handlers require nvbootctrl, which the default image does not install

jetm:
Fixes #353.

## Problem

A stock Jetson image ships SWUpdate handlers it cannot run. `rootfs-pre.sh` exits at its own first check:

```
# command -v nvbootctrl || echo ABSENT
ABSENT
# command -v swupdate || echo ABSENT
ABSENT
```

The packages were never missing. `packagegroup-avocado-tegra-extra` already lists `swupdate` and `tegra-redundant-boot-base`, but that group is wired into `PKG_EXTRA_INSTALL`, which populates the feed rather than the image:

```
kas/vendor/nvidia.yml:32  ROOTFS_IMAGE_EXTRA_INSTALL:append = " packagegroup-avocado-tegra-rootfs "
kas/vendor/nvidia.yml:34  PKG_EXTRA_INSTALL:append       = " packagegroup-avocado-tegra-extra"
```

So both were installable and not installed, and the only way to get them was to also take the extra group - the one carrying TensorRT, DeepStream and the CUDA stack. That is the wrong thing to have to buy in order to be able to update.

## Solution

Add them to `packagegroup-avocado-tegra-rootfs`, which is the group every Jetson rootfs already takes.

`tegra-redundant-boot-base` rather than a bare `nvbootctrl`, because the config chain has to come with it. It pulls `setup-nv-boot-control-service`, whose unit renders `/etc/nv_boot_control.conf` at boot; that pulls `tegra-nv-boot-control-config` for the template it renders from, and `tegra-eeprom-tool-boardspec` for the boardspec it reads. `nvbootctrl` on its own is a tool that fails on a missing config file.

`swupdate` carries its own `/etc/swupdate.cfg` and `/etc/hwrevision`. The pre- and post-install handlers are not part of it - the tegra bbappend deploys those into the `.swu`, so the update bundle carries them rather than the image.

## Verification

Every package in the closure resolves against a built `jetson-orin-nano` feed, checked against the repo metadata rather than assumed:

```
present: tegra-redundant-boot-base    present: setup-nv-boot-control
present: swupdate                     present: setup-nv-boot-control-service
present: tegra-nv-boot-control-config present: tegra-eeprom-tool-boardspec
present: efivar                       present: util-linux-lsblk
```

This has not been built or flashed. It is a packagegroup edit whose names I verified against the feed; the image-level check is a build.

## Reviewer notes

**This activates a latent defect, filed as #401.** `setup-nv-boot-control` ships an ESP mount unit with `What=/dev/disk/by-partlabel/esp`, and its recipe enables that unit from `pkg_postinst`. A PARTLABEL names a role rather than a disk, so on a board provisioned to two media it resolves arbitrarily. Measured on the bench board, booted from NVMe:

```
# ls -l /dev/disk/by-partlabel/esp
/dev/disk/by-partlabel/esp -> ../../mmcblk0p11
```

That is the SD card. Both disks carry `esp` at p11 and `esp_alt` at p14.

It is inert today because the package is not in the image, which is exactly what this PR changes. It stays harmless in the near term - nothing in meta-avocado references `/boot/efi`, and `setup-nv-boot-control.sh` writes EFI variables through efivarfs rather than files on the ESP - but it stops being harmless when bootloader updates land, because then something writes an ESP and it would be the wrong one. #401 lays out the options; I did not fix it here because a `.mount` override needs a build to verify and it is a different piece of work from this one.

Worth deciding before merge whether you want the mask in the same change.

**I left the two packages listed in `packagegroup-avocado-tegra-extra` alone.** They are now redundant there, but a duplicate RDEPENDS costs nothing and removing them is churn in a group this PR otherwise does not touch.

**This does not make an A/B update work end to end**, it only removes the first of the reasons it could not. #399 was the second: the handler resolved partitions with `blkid` flags busybox does not implement, fixed in #400.


jetm:
Marking this draft. The premise I opened it on is incomplete, and the correction matters enough that someone should decide the direction before this merges.

**The BSP extension already declares both packages.** `avocado-bsp-jetson-orin-nano-devkit`, published in the `-ext` feed for `2024/edge` and `2026/edge`, is a manifest-only RPM whose `avocado.yaml` carries:

```yaml
    packages:
      ...
      swupdate: '*'
      tegra-redundant-boot-base: '*'

    on_merge:
      - nvbootctrl verify
```

So the intended delivery vehicle for these two is the BSP extension, not the rootfs packagegroup, and it has been all along. `on_merge` running `nvbootctrl verify` says as much.

**That same manifest documents when to use the rootfs packagegroup instead**, which is the part that makes this a real design question rather than an oversight:

```
# tegra-firmware-tegra234 (the ga10b kernel firmware blobs) is delivered via
# packagegroup-avocado-tegra-rootfs in the rootfs erofs base, *not* here,
# because the kernel firmware loader returns -40 (ELOOP) when reading firmware
# files through the sysext /usr overlay.
```

The rule that implies is "extension by default, rootfs only where a sysext overlay cannot serve the file". Nothing suggests `nvbootctrl` or `swupdate` hit that constraint, so by that rule this PR puts them in the wrong place.

**Why the packages are still absent in practice**, which is what #353 is actually reporting:

- The extension route is broken by avocado-linux/avocado-cli#283. `packages:` resolves and installs through dnf and leaves an empty extension sysroot, so a BSP extension currently delivers nothing. I posted evidence there that this is a regression: an older install of the same manifest-only RPM produced 268 files including `usr/sbin/nvbootctrl`, `usr/bin/swupdate` and `etc/swupdate.cfg`.
- The locally built feed has no jetson `-ext` repo at all, so in a bakar/kas workflow the BSP extension is not fetchable even when the CLI works.

So there are three options, and the choice is not mine:

1. Land this as a workaround while #283 is open, and revert it once the extension path works. Cheap, but it duplicates a declaration that already exists and the duplicate will outlive the reason for it.
2. Close this and treat #353 as blocked on #283 plus publishing a jetson `-ext` repo for locally built feeds.
3. Land it permanently on the argument that the A/B update tooling belongs in the base image rather than in an extension, since an update path that depends on an extension having merged is a weaker guarantee than one baked into the rootfs. This is the only option that survives #283 being fixed, and it is a real position rather than a workaround.

I lean 3 for the update tooling specifically, because the thing that repairs a broken system should not itself depend on the extension layer having come up. But that is a product call.

The #401 ESP finding in the reviewer notes is unaffected by any of this and stands on its own.

jetm:
Further correction, and it weakens the case for this PR again.

I said the extension route was blocked by avocado-cli#283. It is not. I tested a fetched extension on the current CLI and it populates correctly:

```
[SUCCESS] Created sysroot for extension 'avocado-bsp-jetson-orin-nano-devkit'.
```

266 files, 103 kernel modules, and the two packages this PR adds:

```
/etc/nv_boot_control.template
/etc/swupdate.cfg
/usr/bin/swupdate
/usr/sbin/nvbootctrl
```

avocado-cli#283 turns out to affect only an extension declared inline in a project's own `avocado.yaml` with a `packages:` map, which is the shape I happened to test first. `source: {type: package}` is fine, and that is what the BSP extension uses.

So on a published-feed workflow, a Jetson consumer already gets `nvbootctrl` and `swupdate` today by taking the BSP extension, and this PR is unnecessary for them.

What remains is narrower and is not a meta-avocado recipe problem:

- No `avocado-bsp-jetson-orin-nano` is published. The in-tree source is `avocado-os/bsp/jetson-orin-nano-devkit/`, still on the pre-rename machine name, and 2026/edge publishes that extension's `-ext` repo without a matching main target repo.
- A locally built bakar feed has no jetson `-ext` repo at all, so the extension is not fetchable in the maintainer workflow. `scripts/dev-package-extensions.sh -t <target>` is what produces one, and nobody has run it for this target here.

My recommendation is now to close this rather than land it, unless option 3 from my previous comment is the position you want to take deliberately - that the A/B update tooling belongs in the base image because an update path which depends on an extension having merged is a weaker guarantee than one baked into the rootfs. That argument is untouched by any of the above, but it is a product decision and not something to land as a side effect of a bug report.

Leaving it draft. #401 stands on its own either way.

jetm:
Closing. The premise did not survive checking.

A Jetson consumer on the published feed already gets `nvbootctrl` and `swupdate` by taking the BSP extension, which declares both and which I verified installs correctly on the current CLI - 266 files including `/usr/sbin/nvbootctrl`, `/usr/bin/swupdate` and `/etc/swupdate.cfg`. The extension is the intended delivery vehicle, its own manifest documents the rootfs packagegroup as the exception for files a sysext overlay cannot serve, and neither of these is that.

What #353 is really reporting is narrower and lives elsewhere:

- No `avocado-bsp-jetson-orin-nano` is published. The in-tree source is still `avocado-os/bsp/jetson-orin-nano-devkit/`, on the machine name meta-avocado renamed in cc328deb.
- A locally built feed carries no jetson `-ext` repo, so the extension is not fetchable in a bakar/kas workflow. `scripts/dev-package-extensions.sh -t <target>` is what produces one.

Neither is fixed by editing a packagegroup, so this PR is the wrong shape for them.

The one argument I would still make on its merits is that update tooling arguably belongs in the base image, because an update path that depends on an extension having merged is a weaker guarantee than one baked into the rootfs. If that is a direction anyone wants, it deserves its own issue and its own reasoning rather than arriving as a side effect of a bug report. The branch stays if it is wanted.

#401 came out of this and stands on its own.

Issue #398, opened 2026-09-18
tegra234: KERNEL_ARGS carry-forward is safe when the serial console is ordered last (console=tty0 steals /dev/console)


jetm:
## What happens

On `avocado-jetson-orin-nano` (tegra234), repacking `boot.img` at provision time with the machine's `KERNEL_ARGS` on the command line boots to a silent hang. The board reaches real systemd, prints the early "mounting virtual filesystems" work, and then produces no further console output. No panic, no error, no prompt. Reproduced on two independent flashes, and waiting five minutes does not change it. Ctrl-C and sysrq produce nothing.

Passing only the two `avocado.root_partuuid*` arguments and nothing else boots to a login prompt every time, on both a from-scratch flash and the published reference image. That is the form #391 ships, so this is not blocking anything today.

## Why it exists at all

The carry-forward was written to duplicate #347, which fixes a real problem on AGX Thor (tegra264): an empty `mkbootimg` cmdline sends L4TLauncher to the DTB bootargs. tegra264 and tegra234 do not have the same console story, and the remedy does not transfer. #391's own body flagged the hunk as one the forward-port should drop; it should have been dropped on this side too, and now is.

## What was never isolated

Which token does it. The whole set was carried or none of it was, and nobody bisected. For `avocado-jetson-orin-nano` the effective value is the `orin-nano.inc` base plus the layer's append:

```
mminit_loglevel=4 console=ttyTCU0,115200 firmware_class.path=/etc/firmware
fbcon=map:0 video=efifb:off console=tty0 efi=runtime
nvme.use_threaded_interrupts=1 pci=pcie_bus_perf
audit=0 nvme_core.default_ps_max_latency_us=0 pcie_aspm=off
```

## Leading hypothesis, unproven

`console=tty0` appears after `console=ttyTCU0,115200`, and the kernel gives `/dev/console` to the **last** `console=` argument while printk still goes to all of them. That would produce exactly this symptom: kernel messages keep arriving on the serial console, then everything stops the moment userspace opens `/dev/console`, because userspace output is now going to tty0. `video=efifb:off` is in the same list, so tty0 has nowhere to render either and the output is simply lost.

If that is right the board is not hung at all, it is running headless with its console pointed at a disabled framebuffer, and the fix for anyone who wants `KERNEL_ARGS` carried forward on tegra234 is to ensure the serial console is the last `console=` rather than to drop the whole set.

## How to test it

[...]

PR #403, opened 2026-09-18
Give the serial console /dev/console, and carry KERNEL_ARGS into the repacked boot.img

jetm:
Fixes #398.

## Problem

A provisioned Jetson runs with none of the arguments its machine config asks for. The repack passes mkbootimg a command line holding only the two rootfs PARTUUIDs, and the prebuilt image is the only other source, so `nvme_core.default_ps_max_latency_us=0` and `pcie_aspm=off` - both set for this hardware - reach the kernel by no path at all.

The obvious fix is to carry `KERNEL_ARGS` forward, and that hunk existed once. It was removed because the board hung: reached real systemd, then silence, no panic, reproduced on two flashes.

It was not hung. It was headless.

`printk` writes to every registered console, but `/dev/console` follows the **last** `console=` on the line. meta-tegra's tegra234 lists end with `console=tty0` and the same lists set `video=efifb:off`, so the moment userspace opened `/dev/console` its output went to a framebuffer that had been switched off. Kernel messages kept arriving because printk does not care; everything else vanished.

## Solution

Two parts, and the ordering fix has to come first for the carry-forward to be safe.

**Restate the serial console last in each machine's `KERNEL_ARGS`.** meta-avocado already owns that variable for all three tegra234 boards via `KERNEL_ARGS:append`, so this lands where the arguments are assembled. It also covers the prebuilt boot.img, which bakes them at build time and is what a board boots when provisioning does not repack - a fix confined to the repack would leave that path headless.

**Then carry the arguments forward in the repack**, with two guards.

## Verification

Hardware, Jetson Orin Nano. Produced the same ordering at provision time and booted it:

```
# cat /proc/consoles
ttyTCU0              -W- (EC p a)  243:0
netcon0              -W- (E     )
tty0                 -WU (E  p  )    4:1
```

`C` is the flag marking the console behind `/dev/console`, and it is on `ttyTCU0`. The board reaches a login prompt with the console live throughout, carrying the full argument set that previously "hung" it.

The arguments now actually apply:

```
# cat /sys/module/nvme_core/parameters/default_ps_max_latency_us
0
```

That is the first time this setting has been in effect on a provisioned board. The PARTUUID pin is unaffected by the longer line - root still resolves to the pinned partition on `/dev/nvme0n1p1`.

Both guards were checked against the real prebuilt image rather than assumed:

```
magic=[ANDROID!]  hdr_version=[0]     -> v0, so offset 64 is correct
length=379 (limit 511)                -> no split
```

**What is not yet verified:** the machine-config half has not been through a feed rebuild, so the hardware run above produced the correct ordering by patching the provisioning script rather than by rebuilding `KERNEL_ARGS`. The resulting command line is identical and the boot behaviour is what it demonstrates; what remains unexercised is that a bitbake build assembles the same string. That wants one build before merge.

## Reviewer notes

**Stacked on #391**, which is where the removed carry-forward hunk lives. GitHub will retarget on merge.

**Why the guards are not paranoia.** Both failure modes are silent. A v3+ header moved the command line, and reading offset 64 there returns a fragment of another field - non-empty, plausible-looking, and it would ship. mkbootimg does not reject an over-long v0-v2 command line either, it splits at 511 bytes into `extra_cmdline`; the PARTUUID arguments are appended last, so an overflow would move exactly them and silently un-name the rootfs.

**An earlier revision of this PR picked the serial console by rule** - last `console=` that is not `console=tty0` - inside the provisioning script. Review killed it, correctly. The rule reads "tty0 is the only non-serial console", so `console=tty1` or `console=ttynull` would be re-stated last and reproduce the exact bug being fixed; and it put machine-specific policy in a shared script with no way for a display-attached board to opt out. agx-orin is the case that makes any positional rule untenable, since it lists tty0 both first and last.

**Two machines deliberately untouched.** agx-thor needs nothing - tegra264 lists `console=ttyUTC0` and no tty0, so its last `console=` is already serial. icam-540 restates `KERNEL_ARGS` in full, ends with tty0, and does **not** set `video=efifb:off`, so whether it wants a framebuffer console is a product question rather than this defect. Worth a second opinion from someone who knows that board.

**What I did not do:** filter the carried arguments. A `root=` arriving from a future BSP's `KERNEL_ARGS` would fight the PARTUUID pin. None of the current tegra lists carry one, and an allowlist felt like the wrong instinct without a case to point at, but it is the thing a reviewer should push back on if they disagree.


jetm:
## The gap in the original verification is now closed

When I opened this I flagged one thing as unexercised: the machine-config half had never been through bitbake, so the hardware run produced the right command line by patching the provisioning script rather than by rebuilding `KERNEL_ARGS`. That is now done.

Built with `bakar build --target avocado-distro` for `jetson-orin-nano`: 13,260 tasks, no failures. The freshly built prebuilt boot.img carries the ordering assembled by bitbake from the machine config:

```
# dd if=.../tegraflash-bsp/boot.img bs=1 skip=64 count=512 | tr -d '\0'
mminit_loglevel=4 console=ttyTCU0,115200 firmware_class.path=/etc/firmware
fbcon=map:0 video=efifb:off console=tty0 efi=runtime
nvme.use_threaded_interrupts=1 pci=pcie_bus_perf audit=0
nvme_core.default_ps_max_latency_us=0 pcie_aspm=off console=ttyTCU0,115200
```

That matters on its own: this is the image a board boots when provisioning does not repack, and it now has a usable console.

## Then flashed, from the committed code path

The provisioning script that ran came from the build rather than from a hand patch - it carries the header guards this PR adds. It produced:

```
cmdline: ...pcie_aspm=off console=ttyTCU0,115200
         avocado.root_partuuid=1644ECC0-... avocado.root_partuuid_b=0C5CCF9B-...
```

and on the booted board:

```
# cat /proc/consoles
ttyTCU0              -W- (EC p a)  243:0
netcon0              -W- (E     )
tty0                 -WU (E  p  )    4:1

# cat /sys/module/nvme_core/parameters/default_ps_max_latency_us
0

# grep ' / ' /proc/mounts
/dev/nvme0n1p1 / erofs ro,relatime,...
```

`C` on `ttyTCU0` is `/dev/console`. The console stays live through userspace carrying the full argument set that previously read as a hang, the APST setting is applied, and the PARTUUID pin still resolves with the longer command line.

Both guards were also exercised against the real image rather than only reasoned about: the header reads `ANDROID!` with `header_version=0`, and the assembled line is 379 bytes against the 511 limit.

PR #404, opened 2026-09-18
Mount the ESP from the disk the system booted from

### LINKED ISSUE #401: ESP mount unit resolves /dev/disk/by-partlabel/esp, which is ambiguous on a dual-provisioned board

jetm:
Fixes #401.

## Problem

`setup-nv-boot-control` ships `boot-efi.mount` naming its device by PARTLABEL, and enables it from its `pkg_postinst`:

```ini
[Mount]
What=/dev/disk/by-partlabel/esp
Where=/boot/efi
```

A PARTLABEL names a role, not a disk. A board provisioned to more than one medium carries an `esp` on each, and udev gives the symlink to whichever partition it processed. Measured on the bench Orin Nano, booted from NVMe:

```
# ls -l /dev/disk/by-partlabel/esp
/dev/disk/by-partlabel/esp -> ../../mmcblk0p11
```

That is the SD card. Both disks carry `esp` at p11 and `esp_alt` at p14.

This is the same defect class as #390 (OTA target slot) and #391 (initrd rootfs discovery), in a third place.

## Why it is worth fixing now rather than later

Nothing in meta-avocado reads `/boot/efi`, and `setup-nv-boot-control.sh` writes EFI *variables* through efivarfs rather than files on the ESP. So today the cost is a wrong mount that misleads whoever is diagnosing something else.

It stops being harmless the moment a bootloader update path exists, because then something writes an ESP and it would be the disk the firmware did not boot. That is the same failure as the rootfs and OTA slot ones with firmware as the payload, and it is a bad one to discover on a customer board.

## Solution

Resolve it the way the rest of this family was resolved: the booted rootfs names its disk, and the ESP belonging to this install is the one on that disk. A generator reads the root device from `/proc/mounts`, maps it to its whole disk, finds the partition whose sysfs `PARTNAME` is `esp` there, and writes a drop-in overriding `What=`.

`LoaderDevicePartUUID` would be a better source and is what `avocado-slot-root-generator` reads. It is not available here - L4TLauncher does not set it, and this platform exposes no `Loader*` EFI variables at all:

```
# ls /sys/firmware/efi/efivars/ | grep -i loader || echo 'NO Loader* EFI vars'
NO Loader* EFI vars
# ls /sys/firmware/efi/efivars/ | wc -l
56
```

Reading `PARTNAME` from sysfs rather than `blkid` matters for the reason #400 documents: this rootfs ships busybox's blkid, which reports no `PARTLABEL` and silently ignores the flags that would ask for one.

## Verification

On the bench Orin Nano, booted from NVMe, with the SD still installed:

```
# sh avocado-tegra-esp-generator /tmp/gtest; echo "exit=$?"
avocado-tegra-esp-generator: boot-efi.mount -> /dev/nvme0n1p11 (booted disk /dev/nvme0n1)
exit=0

# cat /tmp/gtest/boot-efi.mount.d/10-avocado-booted-disk.conf
[Mount]
What=/dev/nvme0n1p11
```

`/dev/nvme0n1p11` is the booted disk's ESP, against the `mmcblk0p11` the shipped unit would have used. The script that ran matches the one in this PR by md5, so this exercised the committed code.

The no-match path was exercised separately on a host whose booted disk carries no `esp` partition: it logs, writes nothing, and exits 0.

**Not yet verified:** boot-time integration. The generator has not run as an actual systemd generator on a booted image, because that needs a feed rebuild to get the new package in. What is verified is the resolution and the drop-in it emits, run on the target with the real partition layout.

## Reviewer notes

**A generator must not break a boot**, so every branch that cannot answer the question leaves the unit exactly as shipped and exits 0. That is deliberately not fail-closed: refusing to mount would change behaviour on a single-disk board where the old ambiguous lookup happened to be right, and this change should not be able to make any currently-working board worse.

Two of those branches are worth a look. Finding no `esp` on the booted disk leaves the shipped unit alone. Finding more than one also does nothing and says so - a disk whose own layout is ambiguous is not something to resolve by picking, which is the lesson from the other three.

**Placement.** It is in `packagegroup-avocado-tegra-rootfs` rather than beside `setup-nv-boot-control`, because it has to be present wherever the mount can be enabled - including images that pull the redundant-boot chain in through the BSP extension rather than a packagegroup. It carries no RDEPENDS on `setup-nv-boot-control` for the same reason: it does nothing harmful when that unit is absent, and depending on it would drag the whole chain into any image wanting this fix.

**The alternative I did not take** is masking `boot-efi.mount`. Defensible today since nothing reads `/boot/efi`, and cheaper. I went the other way because the mount is a capability something will want once bootloader updates land, and a mask would have to be found and removed at exactly the moment someone is busy with something else.


jetm:
## Verified at boot, from a real build

Built through bitbake, synced to a feed, installed, flashed, and booted on the bench Orin Nano with SD and NVMe both provisioned and the system running from NVMe.

The package is installed and the generator ran during early boot:

```
# ls -l /usr/lib/systemd/system-generators/avocado-tegra-esp-generator
-rwxr-xr-x 1 root root 3062 /usr/lib/systemd/system-generators/avocado-tegra-esp-generator

# cat /run/systemd/generator/boot-efi.mount.d/*.conf
# Written by avocado-tegra-esp-generator. See that script for why the shipped
# What=/dev/disk/by-partlabel/esp cannot be trusted on a multi-disk board.
[Mount]
What=/dev/nvme0n1p11
```

The decisive check is systemd's own view of the unit, rather than the file the generator wrote:

```
# systemctl show boot-efi.mount -p What -p LoadState -p ActiveState -p Result
LoadState=loaded
ActiveState=inactive
What=/dev/nvme0n1p11
Result=success
```

systemd loaded the unit with the override applied. Against what the unit ships with:

```
# ls -l /dev/disk/by-partlabel/esp
/dev/disk/by-partlabel/esp -> ../../mmcblk0p11
```

`mmcblk0p11` is the SD card's ESP. The unit as shipped would have mounted it on a board running from NVMe; with this change systemd resolves the ESP on the booted disk instead.

## What is and is not covered

`ActiveState=inactive` and `is-enabled` reporting `static` are expected here, and worth being precise about: nothing pulls the unit in on this image, so the mount never runs. That is the same dormancy I measured earlier when the unit arrives through the BSP extension. So what this verifies is that the generator runs, resolves correctly, and that systemd picks up the override - not that the resulting mount succeeds, which needs an image where something wants the unit.

The no-match branch was exercised separately on a host whose booted disk carries no `esp` partition: it logs, writes nothing, exits 0, and leaves the unit exactly as shipped.

## One defect this found in itself

The first build of this recipe failed:

```
ERROR: Directory name ${@d.getVar('S') contains unexpanded bitbake variable
✗ FAILED: avocado-tegra-esp-generator-1.0-r0 do_unpack
```

`S = "${UNPACKDIR}"` - the same break 4d101bd3 fixed in `avocado-boot-device`. The recipe now follows that one's shape, with no `S` override and `${WORKDIR}` in `do_install`. It is worth noting that CI's parse check passed this recipe; only a real build caught it.

PR #408, opened 2026-09-19
qcom(rb3gen2): boot the UEFI FIT device-tree path to stop the intermittent brick

mobileoverlord:
## What

The RB3 Gen 2 intermittently bricks on boot (bench-measured ~65% of boots): the board drops into XBL emergency download (`05c6:900e`) and is unrecoverable without a physical EDL power cycle. The crash is in the vendor UEFI, before systemd-boot counts a boot attempt, so A/B rollback cannot recover it.

Root cause: UEFI's legacy `DtPlatformLoadDtbBlob` path loads `combined-dtb.dtb` and runs `BoardInfoDxe`'s `qcom,platform-parts-info` fixup, which intermittently reads an uninitialised pointer and takes a data abort (`panic (caller 0x9fb02c88): die`). It is a vendor-firmware bug, not our device tree; padding or shrinking the base dtb only shifts the odds.

## How

Boot the UEFI down its FIT device-tree path instead of the legacy fallback. UEFI tries `qclinux_fit.img` on `dtb_a` first; supplying a valid FIT makes that load succeed and do FIT-based DT selection, so the buggy legacy fixup is never entered (Qualcomm's documented forward path).

- **`avocado-rb3gen2.conf`**: enable meta-qcom's `dtb-fit-image` and point it at the linux-qcom overlay compatibles (combo `qcom_qcs6490-iot-el2kvm = "qcs6490-rb3gen2 kodiak-el2"`).
- **`image-avocado-qcom-deploy.bbclass`**: when a FIT is built, assemble `dtb.bin` as a FAT holding `qclinux_fit.img` plus `combined-dtb.dtb` as a fail-safe fallback (a bad FIT degrades to the old behaviour instead of a hard brick). Guard the `nostamp` `do_deploy_fixup` on `do_generate_qcom_fitimage` so a build can never stage the legacy dtb before the FIT is deployed. `QCOM_DTB_DEFAULT` stays the real board dtb because the UKI merge reads it, so meta-qcom's `QCOM_DTB_DEFAULT="multi-dtb"` switch is not usable here.
- **`avocado-img-ufs.bb`**: bump `PR` so the changed bootfiles archive is not served from a stale cache.

## Results

Measured on hardware with a reboot loop and serial capture:

| Config | Bricks |
| --- | --- |
| Legacy `combined-dtb.dtb` (baseline) | ~65% |
| FIT | 0/20 |
| Revert to legacy (control) | brick returns |

A FIT boot shows `ParseFitDt` -> `FitLoadDtbFromFdt: Config fdt-qcs6490-rb3gen2.dtb` with no `status=0xE`, no Legacy MultiDtb and no fault. Also validated end to end through `avocado install` -> `build` -> `provision --profile ufs`, booting clean on the FIT.

Note: `rubikpi3` shares this SoC and very likely the same firmware bug; the same enablement should apply there in a follow-up.


meta-avocado-qcom/classes/image-avocado-qcom-deploy.bbclass
```diff
@@ -128,13 +137,43 @@ do_deploy_fixup () {
...
+    # QCOM_DTB_DEFAULT must stay the real board dtb. dtb-fit-image still builds
+    # qclinuxfitImage on its own, so we fold it into dtb.bin here instead.
+    #
+    # combined-dtb.dtb is copied into the SAME FAT as a fail-safe: UEFI tries
+    # \qclinux_fit.img first and only falls back to \combined-dtb.dtb if the FIT
+    # is ever unreadable, so a bad FIT degrades to the old behaviour instead of
+    # a hard brick. mkfs geometry must match the board's UEFI FAT driver
+    # (QCOM_VFAT_SECTOR_SIZE, 4096 on UFS parts); 4 MiB matches the legacy
+    # dtb.bin and fits the 64 MiB dtb_a partition.
+    if [ -f ${DEPLOY_DIR_IMAGE}/qclinuxfitImage ]; then
```
nicksinas: **The FIT branch and its task dependency use two different conditions, so a missing FIT silently ships the bricking image.**

Line 48 gates the dependency on `dtb-fit-image` being in `KERNEL_CLASSES`. This line gates the actual work on `[ -f .../qclinuxfitImage ]`. When the two disagree, the `elif` below stages the legacy raw dtb and the build stays green.

How it happens:

- Someone enables the class from a kernel bbappend (`KERNEL_CLASSES:append` in a `linux-avocado-qcom_%.bbappend`) instead of machine conf. The inline python on line 48 reads the image recipe datastore and does not see a kernel-recipe-level append, so the dependency is absent and the FIT can be missing when this runs.
- meta-qcom changes `qclinuxfitImage` or `do_generate_qcom_fitimage` to something else on a layer bump.

Impact: `avocado provision` writes a dtb_a that boots the legacy path at the ~65% brick rate you measured, and recovery needs a physical EDL power cycle per board. `qcom_check_rawprogram_payloads` cannot catch it, because `dtb.bin` exists on both paths. The comment on lines 71-74 of this file warns about this exact silent-skip class.

Suggested fix: drive the branch from the same condition as the dependency, and stop the build when the FIT is enabled but absent. `linux-avocado-qcom-uki.bb:119-125` already uses that pattern for its base dtb.

```sh
if ${@bb.utils.contains('KERNEL_CLASSES', 'dtb-fit-image', 'true', 'false', d)}; then
    if [ ! -f ${DEPLOY_DIR_IMAGE}/qclinuxfitImage ]; then
        bbfatal "dtb-fit-image is enabled but ${DEPLOY_DIR_IMAGE}/qclinuxfitImage is missing." \
                "Staging the legacy dtb here would ship the UEFI path that bricks this board."
    fi
    ...
```

nicksinas: **A stale `qclinuxfitImage` keeps this branch active after the machine turns the FIT off.**

`do_generate_qcom_fitimage` publishes into `DEPLOY_DIR_IMAGE`, which BitBake never prunes. `do_deploy_fixup[cleandirs]` cleans the staging directory, not the deploy directory this test reads.

How it happens: a developer builds with `KERNEL_CLASSES += "dtb-fit-image"`, removes that line to bisect a problem, then rebuilds in the same TOPDIR. The file test still passes, so `dtb.bin` gets a FIT built from an older kernel and older device trees.

Impact: dtb_a holds a device tree that does not match the deployed kernel. The board boots, so the symptom is a peripheral that stops probing rather than a clean build error. CI never reproduces it, because CI starts with an empty deploy directory. That combination makes this expensive to track down.

Suggested fix: the same one as the other comment on this line. Branch on `KERNEL_CLASSES`, not on the file. That leaves the machine conf as the single source of truth and closes the stale-artifact path in both directions.

meta-avocado-qcom/classes/image-avocado-qcom-deploy.bbclass
```diff
@@ -128,13 +137,43 @@ do_deploy_fixup () {
...
+        rm -f dtb.bin
+        mkfs.vfat -S ${QCOM_VFAT_SECTOR_SIZE} -C dtb.bin 4096
```
nicksinas: **The 4 MiB size is hardcoded here, but meta-qcom drives the same value from `${DTBBIN_SIZE}`.**

`linux-qcom-dtbbin.bbclass:10` sets `DTBBIN_SIZE ?= "4096"` and uses it at lines 32 and 39 for the two vfat images it builds. This line writes `4096` literally.

How it breaks: meta-qcom raises `DTBBIN_SIZE` on a layer bump, because upstream FITs grew. The legacy path below follows the new size. This path does not.

There is also no check that the FIT plus the fallback dtb fit in 4 MiB. `KERNEL_DEVICETREE` holds five entries today, and adding the camx overlays grows the FIT. `mcopy` then fails with a disk-full error that does not point at this line. The dtb_a partition is 64 MiB, so 4 MiB is a self-imposed cap.

Suggested fixes:

1. Use `${DTBBIN_SIZE}` instead of the literal, so this path tracks meta-qcom.
2. Add an explicit size check before `mcopy` that fails with a message saying which file overflowed and what the budget was.

nicksinas (commented):
Findings are on the lines they apply to.

meta-avocado-qcom/classes/image-avocado-qcom-deploy.bbclass
```diff
@@ -128,13 +137,43 @@ do_deploy_fixup () {
...
+        mcopy -i dtb.bin ${DEPLOY_DIR_IMAGE}/qclinuxfitImage ::/qclinux_fit.img
```
nicksinas: **`dtb.bin` is now regenerated on every build and is not byte-reproducible.**

`mkfs.vfat` without `-i` writes a volume ID taken from the clock, and `mcopy` without `-m` stamps the current time on each file.

Before this change `dtb.bin` came from `do_qcom_dtbbin_deploy`, an sstate task, so a rebuild restored identical bytes. This task is `nostamp` (line 59), so it runs every build and produces different bytes every time.

Impact: `avocado-img-ufs` publishes a different archive under the same NEVRA on each build. A device that already cached `${DISTRO_VERSION}-r1` keeps its old bootfiles. A device that fetches fresh gets different bytes. Nothing tells you which one is on the board. It also removes the ability to answer "did the device tree change between these two builds" by comparing artifacts.

Suggested fix: make the output deterministic.

```sh
mkfs.vfat -S ${QCOM_VFAT_SECTOR_SIZE} -i 00000000 -C dtb.bin ${DTBBIN_SIZE}
touch -d @${SOURCE_DATE_EPOCH} <staged inputs>
mcopy -m -i dtb.bin ...
```

`-m` preserves the mtime you set, which is what meta-qcom's own `mcopy -vsmpQ` does.

meta-avocado-qcom/recipes-avocado/images/avocado-img-ufs.bb
```diff
@@ -5,6 +5,11 @@ built rootfs/var images at provision time."
 LICENSE = "Apache-2.0"
 
 PV = "${DISTRO_VERSION}"
+# Bump when the bootfiles content changes but PV does not, so avocado build/
+# provision gets a new NEVRA instead of reusing the cached archive. r1: dtb.bin
+# is now a FAT holding the FIT (qclinux_fit.img) instead of a raw dtb -- the
+# provisioned dtb_a must carry the new FIT for the board to boot the FIT path.
+PR = "r1"
```
nicksinas: **The `PR` bump is manual, unenforced, and applies to every UFS machine.**

This is the only `PR =` line in the repo, so there is no established habit of bumping it, and nothing checks that it moved.

How it breaks: the next change to `do_deploy_fixup` or to any staged bootfile ships with `PR = "r1"` still in place, and every already-provisioned device reuses the cached archive. The failure is silent and shows up as "my fix did not reach the board".

Second problem: `avocado-img-ufs` is shared. `avocado-stone.bbappend:1` depends on it for all `stone-ufs` machines, so rubikpi3 also gets a new NEVRA and a full re-download although none of its bootfiles changed.

Suggested fixes:

1. Derive the release field from the staged content, for example a short checksum over the bootfiles tarball, so the NEVRA moves when and only when the content moves. That also closes the reproducibility gap flagged on the `mkfs.vfat` line in the bbclass.
2. If the manual bump stays, add a comment at the top of `do_deploy_fixup` in `image-avocado-qcom-deploy.bbclass` that points at this line, so the next person to edit the staging logic sees the rule.

meta-avocado-qcom/conf/machine/avocado-rb3gen2.conf
```diff
@@ -132,3 +132,27 @@ MACHINE_FEATURES = "usbhost usbgadget wifi bluetooth alsa pci efi"
...
+# path is never entered -- Qualcomm's documented forward path (meta-qcom #1124,
+# qcom-deb-images #249).
+#
+# meta-qcom already builds the FIT (dtb-fit-image.bbclass) and ships the
+# qcs6490-rb3gen2 compatible combos (fit-dtb-compatible-linux-qcom.inc:
+# qcom_qcs6490-iot-el2kvm = "qcs6490-rb3gen2 kodiak-el2"); this turns it on and
+# points it at the linux-qcom overlay compatibles. image-avocado-qcom-deploy
+# then stages qclinuxfitImage into the dtb_a filesystem as qclinux_fit.img.
+KERNEL_CLASSES += "dtb-fit-image"
+LINUX_QCOM_FIT_DTB_COMPATIBLE = "conf/machine/include/fit-dtb-compatible-linux-qcom.inc"
```
nicksinas: **This has no effect on the default kernel.**

`linux-qcom_6.18.bb:11` already sets `LINUX_QCOM_FIT_DTB_COMPATIBLE` with a hard `=`, and `linux-qcom-next_git.bb:11` does the same. A recipe-level `=` overrides a machine conf assignment, so this line is dead for both.

The comment above says the change "points it at the linux-qcom overlay compatibles", which reads as though this line does that work. The one kernel it does reach is `linux-qcom-rt`, which does not set the variable, and the comment does not say so.

Impact: maintainability. A future reader edits this line to change which combos the FIT holds, rebuilds, sees no change, and loses time before finding the recipe-level assignment.

Suggested fixes: delete the line, or keep it for `linux-qcom-rt` and say so in the comment, for example "only `linux-qcom-rt` needs this, because `linux-qcom` and `linux-qcom-next` set it in the recipe".

meta-avocado-qcom/classes/image-avocado-qcom-deploy.bbclass
```diff
@@ -128,13 +137,43 @@ do_deploy_fixup () {
...
+        if [ -n "${QCOM_DTB_DEFAULT}" ] && \
+           [ -f ${DEPLOY_DIR_IMAGE}/dtb-${QCOM_DTB_DEFAULT}-image.vfat ]; then
+            mcopy -i ${DEPLOY_DIR_IMAGE}/dtb-${QCOM_DTB_DEFAULT}-image.vfat ::/combined-dtb.dtb combined-dtb.dtb
+            mcopy -o -i dtb.bin combined-dtb.dtb ::/combined-dtb.dtb
+            rm -f combined-dtb.dtb
+        fi
```
nicksinas: **The stated fail-safe covers "FIT unreadable", but the likely failure is "FIT loads and no config matches".**

UEFI picks a FIT config by matching the board root compatible against the config nodes. `dtb-fit-image.bbclass` builds those nodes only from `FIT_DTB_COMPATIBLE` combos whose every part is present in `KERNEL_DEVICETREE`. This machine declares five device trees, so combos needing `kodiak-staging` or the camx overlays are dropped at build time with a `bb.note`, not an error.

How it happens: XBL reports a compatible with no matching config. The live example is `QCOM_XBL_CONFIG_VARIANT` in `avocado-qcs6490.inc:44`, which pins `kvm` today and is the reason `qcom,qcs6490-iot-el2kvm` matches at all. A change there, or a board variant reporting a `-staging` or subtype compatible that no built combo covers, lands in this case.

Impact: the behavior is not measured. UEFI can fail the load, or it can fall back to `\combined-dtb.dtb` and brick at the old rate. The results table in the PR body does not separate the two, and the comment above claims the fallback covers it.

Suggested fixes:

1. Do one boot with a deliberately mismatched compatible, record what UEFI does, and state that result in the comment.
2. Until that is measured, soften the comment to say the fallback is proven only for an absent FIT.
3. Add a build-time check that at least one `FIT_DTB_COMPATIBLE` combo survived the `KERNEL_DEVICETREE` filter, so a machine that builds a FIT with no usable config fails at build time rather than at boot.

meta-avocado-qcom/conf/machine/avocado-rb3gen2.conf
```diff
@@ -132,3 +132,27 @@ MACHINE_FEATURES = "usbhost usbgadget wifi bluetooth alsa pci efi"
...
+# it. Supplying a valid FIT makes the first load succeed so the buggy legacy
+# path is never entered -- Qualcomm's documented forward path (meta-qcom #1124,
+# qcom-deb-images #249).
+#
+# meta-qcom already builds the FIT (dtb-fit-image.bbclass) and ships the
+# qcs6490-rb3gen2 compatible combos (fit-dtb-compatible-linux-qcom.inc:
+# qcom_qcs6490-iot-el2kvm = "qcs6490-rb3gen2 kodiak-el2"); this turns it on and
+# points it at the linux-qcom overlay compatibles. image-avocado-qcom-deploy
+# then stages qclinuxfitImage into the dtb_a filesystem as qclinux_fit.img.
+KERNEL_CLASSES += "dtb-fit-image"
```
nicksinas, line 167: **rubikpi3 keeps the bricking path although it shares the SoC, the UEFI and this bbclass.**

The PR body says rubikpi3 very likely has the same firmware bug and defers it. This enablement is machine-scoped, so `avocado-rubikpi3.conf` still stages a legacy `combined-dtb.dtb`.

How it reaches a user: a rubikpi3 owner provisions from the same release and hits the same unrecoverable EDL state, at whatever rate that board has. Nobody has measured that rate.

Impact: a known-brick configuration stays in the tree, with no note in the file a rubikpi3 user reads and no tracking link.

This is real follow-up work rather than a copied line. Enabling it there needs a `FIT_DTB_COMPATIBLE` combo for `qcs6490-thundercomm-rubikpi3`, which `fit-dtb-compatible-linux-qcom.inc` does not ship.

Suggested fix: add a comment in `avocado-rubikpi3.conf` that points at this change and at the tracking issue, so the gap is visible from the file a rubikpi3 owner reads, and open that follow-up before this merges.

nicksinas (approved):

meta-avocado-qcom/classes/image-avocado-qcom-deploy.bbclass
mobileoverlord: Fixed in f6d17967 — the FIT branch now gates on `bb.utils.contains('KERNEL_CLASSES','dtb-fit-image',...)`, the same machine-conf condition as the task dependency, and `bbfatal`s if the class is enabled but `qclinuxfitImage` is absent (fail-closed, matching linux-avocado-qcom-uki.bb). No more silent fall-through to the legacy dtb.

mobileoverlord: Fixed in f6d17967 by the same change — branching on `KERNEL_CLASSES` instead of the file test makes machine conf the single source of truth, so a stale `qclinuxfitImage` in DEPLOY_DIR_IMAGE no longer keeps this branch active after the class is removed.

meta-avocado-qcom/classes/image-avocado-qcom-deploy.bbclass
mobileoverlord: Fixed in f6d17967 — the FAT is now sized from the legacy dtb vfat (whose size *is* meta-qcom's `DTBBIN_SIZE`), so it tracks a `DTBBIN_SIZE` bump. I derive it from the built vfat rather than `${DTBBIN_SIZE}` directly because that var is `?=` in linux-qcom-dtbbin (a KERNEL class) and isn't in the image recipe's datastore. Also added a pre-`mcopy` size check that `bbfatal`s naming the payload size vs the budget.

meta-avocado-qcom/classes/image-avocado-qcom-deploy.bbclass
mobileoverlord: Fixed in f6d17967 — `mkfs.vfat -i 00000000` (fixed volume id), input mtimes normalised to `SOURCE_DATE_EPOCH`, and `mcopy -m` to preserve them. Verified byte-identical dtb.bin across two independent builds.

meta-avocado-qcom/recipes-avocado/images/avocado-img-ufs.bb
mobileoverlord: Addressed in f6d17967 — scoped to `PR:rb3gen2` so the shared recipe doesn't force other stone-ufs machines (e.g. rubikpi3) into a needless re-download, and documented the rule. With the reproducibility fix the archive changes only when its inputs do. Left a note that a content-derived release field is the right long-term automation.

meta-avocado-qcom/conf/machine/avocado-rb3gen2.conf
mobileoverlord: Correct, fixed in f27f04c8 — removed the `LINUX_QCOM_FIT_DTB_COMPATIBLE` line (dead: virtual/kernel is linux-qcom, which sets it with a recipe-level `=` that overrides machine conf). Kept `KERNEL_CLASSES += dtb-fit-image` (still required — only auto-inherited when QCOM_DTB_DEFAULT == multi-dtb) and corrected the comment. Confirmed on a rebuild that dropping the line produces the same FIT dtb.bin.

meta-avocado-qcom/classes/image-avocado-qcom-deploy.bbclass
mobileoverlord: Good catch — softened in f27f04c8. The comment now says the fallback is proven only for an absent/unreadable FIT (what was measured); a FIT that loads but whose configs don't match the board compatible is UEFI's own behaviour and is not measured here. Noted that the build-time 'at least one FIT_DTB_COMPATIBLE combo survived' check belongs upstream in dtb-fit-image, where the combo filtering happens.

meta-avocado-qcom/conf/machine/avocado-rb3gen2.conf
mobileoverlord: Addressed in f27f04c8 — added a KNOWN GAP note in avocado-rubikpi3.conf pointing at this change, and opened tracking issue #409 (enabling the FIT there needs a FIT_DTB_COMPATIBLE combo for qcs6490-thundercomm-rubikpi3, which meta-qcom doesn't ship, plus on-rubikpi3 validation).

PR #411, opened 2026-09-19
qcom(qcs6490): provision-time hypervisor (Gunyah/KVM) selection + rubikpi3 reboot-edl

mobileoverlord:
## What

Add a provision-time knob to select the qcs6490 XBL hypervisor config, and wire `reboot edl` on rubikpi3.

The qcs6490 bootfiles ship two signed XBL configs — `xbl_config_gunyah.elf` and `xbl_config_kvm.elf` — and only one is flashed as `xbl_config.elf`. They are a hard tradeoff:

- **gunyah**: Gunyah owns EL2 (dormant — no `/dev/kvm`, no guest VMs); the DSP/GPU secure firmware authenticates and runs (cdsp/adsp remoteprocs up, Adreno a660 zap passes PAS auth).
- **kvm**: Linux owns EL2, `/dev/kvm` and guest VMs work, but PAS auth fails with `-22` (KVM does not provide the Qualcomm secure-memory mediation the TrustZone expects), so cdsp/adsp/gpu-zap will not load.

Same TZ, same firmware; the only difference is who owns EL2.

## How

- `stone-provision-ufs.sh` (rubikpi3 + rb3gen2) copies `xbl_config_<variant>.elf` over `xbl_config.elf` before `qdl`, selected by `AVOCADO_HYPERVISOR` (`gunyah` | `kvm` | `none`; default `gunyah`). No re-signing — both variants already ship signed.

  ```
  avocado provision --env AVOCADO_HYPERVISOR=kvm
  ```

- rubikpi3 DTS: add `&psci { reboot-mode { mode-edl = <0 0x1>; ... } }` (matches `qcs6490-rb3gen2.dts`) so `reboot edl` reaches EDL without the physical button. `qcom,dload-mode` comes from `sc7280.dtsi`.

## Results (verified on a RUBIK Pi 3)

- `--env AVOCADO_HYPERVISOR=gunyah` → `error -22` count 0, cdsp + adsp remoteprocs running, Adreno GPU up, no `/dev/kvm`.
- `--env AVOCADO_HYPERVISOR=kvm` → `/dev/kvm` present (VHE initialized), DSPs back to `-22` (expected).

Applies to both qcs6490 boards. rb3gen2 verified by inspection: both signed xbl_config variants ship, the GPU/CDSP firmware is present, and its DTB firmware-names match shipped files; its reboot-mode is already present in the mainline dts. Boot-partition choice written at flash time, orthogonal to A/B OS updates (both slots share `xbl_config`).


meta-avocado-qcom/recipes-kernel/linux/files/qcs6490-thundercomm-rubikpi3.dts
```diff
@@ -970,6 +970,21 @@
...
 
+/*
+ * Software reboot-into-EDL, so `reboot edl` reaches Qualcomm EDL (9008) without
+ * the physical EDL button -- same as qcs6490-rb3gen2.dts. The magic values are
+ * passed to PSCI SYSTEM_RESET2; the firmware (identical XBL/TZ to rb3gen2)
+ * interprets mode-edl to enter EDL on the next reset. qcom,dload-mode comes from
+ * sc7280.dtsi. Without this the rubikpi3 DTS had no reboot-mode node at all, so
+ * `reboot edl` fell through to a normal reboot (rb3gen2 worked, rubikpi3 did not).
+ */
+&psci {
```
lee-reinhardt, line 981: Error: This DTB change also needs a RUBIK Pi bootfiles package revision bump. The board gets its DTB through `dtb.bin` in `avocado-img-ufs`, but this PR leaves that package's version/release unchanged. When `DISTRO_VERSION` stays the same, existing projects can reuse the cached bootfiles archive and reflash the old DTB, so `reboot edl` still lacks the new configuration. Add `PR:rubikpi3 = "r1"` to `meta-avocado-qcom/recipes-avocado/images/avocado-img-ufs.bb`, as that recipe's revision policy requires for changed bootfiles.

lee-reinhardt (changes requested):

meta-avocado-qcom/recipes-kernel/linux/files/qcs6490-thundercomm-rubikpi3.dts
mobileoverlord: I checked this against the build and I don't think the bump is needed — the PR service already does it automatically.

`kas/vendor/oe.yml` sets `PRSERV_HOST = "localhost:0"`, so bitbake runs a PR service and appends its own auto-increment to whatever `PR` the recipe declares. That's where the `.N` in the package names comes from:

```
avocado-img-ufs-0.0.0-r1.4.avocado_rb3gen2.rpm
                      ^^ recipe PR      ^ prserv auto-increment
```

The service keys on the package's output checksum, so any change to the staged bootfiles — including a new board dtb — yields a new checksum and therefore a new NEVRA, with no recipe edit. From this build's `prserv.sqlite3`:

| recipe PR base | distinct content hashes | highest auto-increment |
| -------------- | ----------------------- | ---------------------- |
| `...-r0`       | 29                      | 28                     |
| `...-r1`       | 5                       | 4                      |

29 separate NEVRAs were issued under the bare `r0` base alone, one per distinct bootfiles content, before any manual `PR` existed. So a project cannot reuse a cached archive across a dtb change — the version it resolves has already moved.

Which also means the `PR:rb3gen2 = "r1"` that #408 added was unnecessary, and the comment block above it (including my note about a content-derived release field being "the right follow-up") is wrong — prserv was already doing exactly that. I'd rather delete that block in a follow-up than add a second copy of it here.

Happy to add `PR:rubikpi3` anyway if you'd prefer the explicit marker, or if there's a builder in the release path that runs without `PRSERV_HOST` — that's the one case where this argument breaks, and you'd know better than me whether the release builders are configured the same way as the kas file.

lee-reinhardt (approved):

